PR #25, opened 2021-09-27
Zalamar/inl 55 projectsprojectidmachine translations

Zalamar:


packages/dashboard/src/routes/projects/[projectId]/machine-translations.svelte
```diff
@@ -0,0 +1,48 @@
...
+                <Toggle labelText="Automatically translate new keys" disabled={!toggled}/>
+            </div>
+            <div class="toggle">
+                <Toggle labelText="Automatically translate new languages" disabled={!toggled}/>
+            </div>
+        </TabContent>
+    </div>
+</Tabs>
+
+<style>
```
Comment (samuelstroschein) [inline packages/dashboard/src/routes/projects/[projectId]/machine-translations.svelte:44]: Please do not use custom CSS styles instead use TailwindCSS. [Here](https://tailwindcss.com/docs/margin#class-reference) is the documentation for margins ([which should never by `px` btw ](https://chiamakaikeanyi.dev/sizing-in-css-px-vs-em-vs-rem/#:~:text=px%20is%20not%20scalable%2C%20it,px%20unit%20by%20the%20browser.)). 

So `<div class="toggle">` becomes `<div class="mb-2">`. Even better, most of the times Carbon components allow to pass classes down, so `<Toggle class="mb-2" ...>` likely works

samuelstroschein (changes requested):

samuelstroschein (approved):

PR #26, opened 2021-09-27
Finish initial card for translations, WIP machine translation

Zalamar:


packages/dashboard/src/lib/components/modals/TranslationModal.svelte
```diff
@@ -0,0 +1,44 @@
...
+            {#each Object.keys(originalTranslations) as lang}
+                <Tab>{lang}</Tab>
+            {/each}
+            <div slot="content">
+                {#each Object.keys(originalTranslations) as lang}
+                <TabContent>
+                    <TextArea rows={10} labelText="Translated text:" bind:value={translations[lang]}/>
+                    <ButtonSet>
+                        <Button kind="secondary" on:click={() => translations[lang] = originalTranslations[lang]}>Reset</Button>
+                        <Button disabled={!machineTranslation} on:click={() => translations[lang] += " but cooler"}>Machine Translation</Button>
```
samuelstroschein, line 37: I prefer to not use the negate operator `!` but instead `machineTranslation === false` because the `!` bang operator has no meanings in Typescript and is easy to overlook. The first meaning is negation when at the front e.g. `!something` and the second is `something!.value` where `something` can be null but `!` casts `something` to non-null. 

Furthermore, `boolean` variables should most of the time start with `is` or `has`:

`machineTranslation` -> `isMachineTranslation`

packages/dashboard/src/lib/components/modals/TranslationModal.svelte
```diff
@@ -0,0 +1,44 @@
+<script lang="ts">
+    // TODO: Implement actual machine translation
+
+    import {Button, ButtonSet, Modal, Tab, TabContent, Tabs, TextArea} from "carbon-components-svelte";
+
+    export let open = false;
```
samuelstroschein, line 6: `open` -> `isOpen`

packages/dashboard/src/lib/components/modals/TranslationModal.svelte
```diff
@@ -0,0 +1,44 @@
...
+
+<Modal
+    bind:open
+    modalHeading="Translation Card"
+    size="sm"
+    hasform
+    primaryButtonText="Confirm translations"
+    secondaryButtonText="Cancel"
+    on:click:button--secondary={() => (open = false)}
+    on:submit={() => {open = false; newTranslations = true}}
```
samuelstroschein, line 24: Why does the submission differ between new and "old" translation?

packages/dashboard/src/lib/components/modals/TranslationModal.svelte
```diff
@@ -0,0 +1,44 @@
...
+    // TODO: Implement actual machine translation
+
+    import {Button, ButtonSet, Modal, Tab, TabContent, Tabs, TextArea} from "carbon-components-svelte";
+
+    export let open = false;
+    export let machineTranslation = true; // This should be local settings
+    // Default values for understanding, will be removed once implemented
+    export let original = "Settings";
+    export let translations = {de: "Einstillinger", da: "Indstillinger"}; // Sounds more fun than einstellungen
+    export let newTranslations = false; // Will be true when user submits new translations
```
samuelstroschein, line 11: `newTranslations` -> `hasNewTranslations` or `is` but its plural?

samuelstroschein (approved):
Added comments. Not extremely important so you can merge it without modifying the code

packages/dashboard/src/lib/components/modals/TranslationModal.svelte
Zalamar: Just an explicit way of saying "please update" when calling the component after it closes.

PR #61, opened 2021-11-01
Sebastian/inl-183-docs-getting-started-sections-are-not

SebastianLJ:


Zalamar:
change svelte-kit-example to svelte-kit

PR #89, opened 2022-01-27
Migrate-to-supabase-cli-for-migrations

samuelstroschein:
## Context 

Inlang started when supabase was in early alpha. The latter entailed that supabase had no migration tooling. Prisma was adopted alongside a manually "managed" docker compose setup as migration tool. Unfortunately, prisma was not a good fit because it only manages migrations for the schema itself, not functions, triggers and policies which are a crucial part of Postgres. Furthermore, maintaining the docker compose setup was a hassle and lead to multiple bugs.


samuelstroschein:
Nice, the migration resulted in 1000 lines fewer code.

PR #127, opened 2022-08-02
rfc-001-core-architecture

samuelstroschein:
**[Rendered version](https://github.com/inlang/inlang/blob/rfc-001-architecture/rfcs/001-core-architecture.md)**


rfcs/001-foundational-architecture.md
```diff
@@ -0,0 +1,564 @@
...
+
+Using an SDK does not relieve developers from two extremely time-consuming and ever-repeating tasks:
+
+1. Extracting hard-coded strings.
+
+   Developers have to manually copy & paste hardcoded strings into resources. That process can take weeks.
+
+2. Validating translations.
+
+   No widely adopted tools exist to validate translations: Does the German translation exist? Is the French translation correct? Does the UI work as expected for Arabic?
```
hannes-oster, line 240: 'Is the french translation correct' seems like a poor example. Sounds like we are in the business of translating strings rather than providing tools for localization.

rfcs/001-foundational-architecture.md
```diff
@@ -0,0 +1,564 @@
...
+</Translated>
+```
+
+> At this point you might be wondering why the text "Click the button to continue." itself is not used as key/id. The reasons stems from changing the text and thereby losing the connection to the translations or invalidating them. The Fluent project explained their rationale of using ids [here](https://github.com/projectfluent/fluent/wiki/Fluent-vs-gettext#translation-invalidation).
+
+#### Observations
+
+1. Dev tools beyond SDKs are required to make localization effortless.
+
+2. A CLI to extract and validate messages enables CI/CD.
```
hannes-oster, line 267: Why? Not explained in the previous paragraphs (first time mentioning CLI)

rfcs/001-foundational-architecture.md
```diff
@@ -0,0 +1,564 @@
...
+        Providers-.-Host
+        git[git-js]
+        Host[git-provider-api]
+        validation
+        ast
+        config
+    end
+    subgraph Legend
+        direction LR
+        start1[ ] --->|depends on| stop1[ ]
```
hannes-oster, line 494: reverse direction of dependency

rfcs/001-foundational-architecture.md
samuelstroschein: Changed "translation" to "message".

rfcs/001-foundational-architecture.md
samuelstroschein: Good remark. Removed the legend. The arrows are clear, the legend makes the relations unclear.

rfcs/001-foundational-architecture.md
samuelstroschein: You're right. I have to extend that section. Will do that early next week.

ivanhofer:
Hi @samuelstroschein, very detailed write-down 👏.
Here are some comments on the points of th RFC:

## Human Language Syntax

I would choose **Proposal A**, if it supports already > 95% of all use-cases (parameters, plural-rules, formatting, gender-rules, combinations). I cannot imagine that it will have fewer features than the current version of fluent. Maybe they could get rid of the "Terms".

## SDK

I would advise against creating your own SDK 😉. Supporting mutiple features, multiple frameworks, CSR & SSR, CJS vs ESM and other things will take a lot of effort. Also convincing a team to use a different SDK is not that trivial. There will probably be not that many teams that will switch to a new SDK for a running project. Making it possible to add inlang on top of the current architecture would be a smart decision.

## Dev tools

A developer should be able to check if something is wrong with the translations. Missing keys, missing/wrong params in messages, invalid keys, invalid message syntax and other possible errors should be able to checked by running a single command in the project. 

## Editor

If no server is involved, each translator would need a GitHub (or other platform) account with access to the full repository. This could have some security implications if the source code of a private repo would be shared with external translators. If it is running on a server, the git acccess could be hidden to translators and an inlang-bot (with the TOKEN of a developer of that project) would then commit the changes to the repository.

## CI/CD

We use this process in a project and it feels great. Not having to manually import new translations, updating messages, looking for errors in those messages feels liberating. A daily process imports the translations automatically, CI checks if the format is correct and then creates a PR and merges the PR. Coming from a project where we had to do this boring and time-consuming process manually every few weeks I cannot imagine to do i18n without CI/CD anymore.

The Editor should already have some checks in place to instantly give feedback to a translator.

## Architecture

### 1. How are messages stored?

I think the development team should be able to decide what message format they want the files stored in the repo. If they don't want to use the Editor to change something, they should be able to do it in a normal text-editor. The read write operations will anyhow need to transform it to an AST, then it is not important to inlang in what format the messages are stored on disk.

### 2. How is inlang configured?

The JS config file with the read and write methods sound good. I think you'll also have to think about different actions a translator could do inside the Editor (e.g. add/remove/change a whole locale). It makes sense if the write function knows what type of action was performed in order to be able to pick the correct "write" action and maybe also to trigger a different workflow depending on the action (e.g. create the PR only when the translator clicks on "submit" at the end of his editing session).


That's it. I hope you'll find my feedback useful.

samuelstroschein:
@ivanhofer as always, thank you very much for your feedback. 

> I would advise against creating your own SDK 😉. 

Your SDK is among the reasons why we went back to the drawing board to support a variety of SDKs. 

> If no server is involved, each translator would need a GitHub (or other platform) account with access to the full repository. This could have some security implications if the source code of a private repo would be shared with external translators. If it is running on a server, the git acccess could be hidden to translators and an inlang-bot (with the TOKEN of a developer of that project) would then commit the changes to the repository. 

Indeed. I view that as a good thing for now. In the future, some auth abstractions could be built if desired. 

--- 

I take from your remarks that inlang is headed in the right direction?

ivanhofer:
> I take from your remarks that inlang is headed in the right direction?

Everything you have described above aligns with the [`long-term-goals`](https://github.com/ivanhofer/typesafe-i18n/discussions/324) of `typesafe-i18n`, so yes, I think you are heading in the right direction ;)

I'm still unsure about the client-only editor, but I don't know that much in that specific area ^^. I am curious to see the result.

PR #176, opened 2022-12-10
fix: border color of input

### LINKED ISSUE #167: align shoelace outline color with `outline` color system token
The outline colors are not matching

image

samuelstroschein:
By overwriting the shoelace color variables. This closes #167. But be aware, not all border colors have been adjusted.

### DIFF
--- a/source-code/website/src/renderer/_default.page.client.tsx
+++ b/source-code/website/src/renderer/_default.page.client.tsx
@@ -6,7 +6,6 @@ import { setCurrentPageContext } from "./state.js";
 import type { PageContextRenderer } from "./types.js";
 // only imported client side as web components are not supported server side
 // importing the shoelace components that are used.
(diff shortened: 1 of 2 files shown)


source-code/website/src/renderer/_default.page.client.tsx
```diff
@@ -6,7 +6,6 @@ import { setCurrentPageContext } from "./state.js";
 import type { PageContextRenderer } from "./types.js";
 // only imported client side as web components are not supported server side
 // importing the shoelace components that are used.
-import "@shoelace-style/shoelace/dist/themes/light.css";
```
samuelstroschein, line 9: Removed the CSS import here to import the CSS in the stylesheet. The latter allows variable overwrites. Otherwise, the server sends the stylesheet which is then overwritten by the client stylesheet that is imported here.

PR #187, opened 2022-12-14
Add the blog section and change the surface100-500 colors

jannesblobel:


samuelstroschein:
- [ ] I would place the "back to index" on top of a blog post. I assume that most readers open a post, decide whether they read it, and more often than not, don't scroll all the way down. 
- [ ] I would rename "Index" to "back to blog"



image

Furthermore, can you render the blog from /content/blog instead of /blog at the root level? The /content directory is supposed to contain "content" that is not related to the source code (in contrast to documentation). /content/blog already contains one blog post (with a directory-based schema). I would keep the directory-based schema as a directory-based schema makes it easier to embed assets like images (just put them in the directory) and later translations of an article.

samuelstroschein (changes requested):
- [ ] I would place the "back to index" on top of a blog post. I assume that most readers open a post, decide whether they read it, and more often than not, don't scroll all the way down. 
- [ ] I would rename "Index" to "back to blog"



image

Furthermore, can you render the blog from /content/blog instead of /blog at the root level? The /content directory is supposed to contain "content" that is not related to the source code (in contrast to documentation). /content/blog already contains one blog post (with a directory-based schema). I would keep the directory-based schema as a directory-based schema makes it easier to embed assets like images (just put them in the directory) and later translations of an article.

source-code/design-system/src/color-system/generateTokens.cts
```diff
@@ -2,7 +2,7 @@ import type { Config } from "./types/config.cjs";
 import type { ColorTokens } from "./types/colorTokens.cjs";
 import { TinyColor } from "@ctrl/tinycolor";
 import type { Color } from "./types/color.cjs";
-
+const surface100Color = 82;
```
samuelstroschein, line 5: You should add a comment what this is for. The name of the variable is also incorrect. This is not the surface100 color but the surface 100 "lighten base".

source-code/design-system/src/color-system/generateTokens.cts
```diff
@@ -22,19 +22,19 @@ export function generateTokens(config: Config): ColorTokens {
 		background: white,
 		"on-background": neutralColors.neutral[900],
 		"surface-100": new TinyColor(neutralColors.neutral[900])
-			.lighten(88)
+			.lighten(surface100Color)
 			.toHex8String(),
 		"surface-200": new TinyColor(neutralColors.neutral[900])
-			.lighten(85)
+			.lighten(surface100Color - 3)
```
samuelstroschein, line 28: The Material spec does not jump in conistent steps. Read https://m3.material.io/styles/color/the-color-system/color-roles. I am going ahead and fix that

source-code/design-system/src/color-system/generateTokens.cts
samuelstroschein: Btw, this seems to be an over-abstraction. The benefit of having a dedicated variable is close to null. Only when a value is changed but that happens maybe once (now)? On the flip side, it makes reading the code harder. 

```ts
"surface-200": new TinyColor(neutralColors.neutral[900])
	.lighten(85)
```

is easier to reason about than 

```ts
"surface-200": new TinyColor(neutralColors.neutral[900])
	.lighten(surface100Color - 3)
```

1. Why is something subtracted from a color?
2. What is the resulting value?  

I'll leave it for now. Your call whether you want to remove it or not

source-code/website/src/pages/blog/index.page.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+// export function Page() {
+// 	onMount(() => {
+// 		// redirect
+// 		navigate("/blog/intro");
+// 	});
+// }
+export function Page() {
+	return (
+		<div class=" flex-row min-h-full   items-center justify-center   mx-auto md:max-w-2xl divide-y divide-solid divide-outline  ">
```
samuelstroschein, line 14: please remove whitespace in css. you have whitespace everywhere

PR #214, opened 2022-12-25
remove bundle from ast before launch

samuelstroschein:
The bundle type was inspired by Fluent, see https://github.com/projectfluent/fluent.js/tree/master/fluent-bundle, but seemes like an uneccessary abstraction for inlang. 

- messages can be grouped in one resource already due to inlang's configurable nature. no bundle concept needs to be introduced. or rather, introduce a bundle _if really required_ for certain usecases. 

- using a bundle id as language identifier (like Fluent) will lead to numerous issues. a bundle id (like a resource id) does not necessarly need to be a language. 

<a href="https://gitpod.io/#https://github.com/inlang/inlang/pull/214"></a>


samuelstroschein:
A good example of using Fluent without modification, as once considered in #127, would have been a wrong decision.

Issue #186, opened 2022-12-14
hardcode inlang branch to enable pull requests


samuelstroschein:
## Problem

The editor uses the "default" branch to make commits. This leads to a convoluted commit history. 

## Proposal 

(Short-term solution) hardcode the editor to commit and push only on an "inlang" branch. 

### Pros
- maintainers can manually choose when and how to merge changes on the inlang branch (squash merge, etc.)
- changes on the inlang branch don't affect the rest of the repo (if not merged) 
- easy to implement

### Cons
- not a long-term solution

## Implementation 

- [x] Create 'inlang' branch if not exist
- [x] Switch to 'inlang' branch after cloning

samuelstroschein:
closed by #219

PR #219, opened 2022-12-27
add a "check branches and create branch" function 

jannesblobel:
This function will check the existing branches on a repo. This branch is created if there is no "inlang" translation branch. 
related to #186

<a href="https://gitpod.io/#https://github.com/inlang/inlang/pull/219"></a>

### DIFF
--- a/source-code/website/src/pages/editor/state.ts
+++ b/source-code/website/src/pages/editor/state.ts
@@ -180,10 +180,10 @@ async function cloneRepository(args: {
 
 	const date = new Date();
 	//checked the existing branches and added a "inlang" translation branch, if this branch does not exist
-	let branches = await raw.listBranches({ fs, dir: "/", remote: "origin" });
-	if (branches.includes("inlang") == false) {
+	const branches = await raw.listBranches({ fs, dir: "/", remote: "origin" });
+	if (branches.includes("inlang") === false) {
 		await raw.branch({ fs, dir: "/", ref: "inlang", checkout: true });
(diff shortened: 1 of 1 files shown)


source-code/website/src/pages/editor/state.ts
```diff
@@ -177,7 +177,20 @@ async function cloneRepository(args: {
 	});
 	// triggering a side effect here to trigger a re-render
 	// of components that depends on fs
+
 	const date = new Date();
+	//checked the existing branches and added a "inlang" translation branch, if this branch does not exist
+	let branches = await raw.listBranches({ fs, dir: "/", remote: "origin" });
```
samuelstroschein, line 183: why `let` instead of `const`? the variable does not seem to be reassigned

source-code/website/src/pages/editor/state.ts
```diff
@@ -177,7 +177,20 @@ async function cloneRepository(args: {
...
+	if (branches.includes("inlang") == false) {
```
samuelstroschein, line 184: always use triple equality `===`! using double equality is a footgun. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness#:~:text=Triple%20equals%20(%20%3D%3D%3D%20)%20will%20do,types%20differ%2C%20false%20is%20returned.

samuelstroschein (changes requested):

samuelstroschein (approved):

samuelstroschein:
Reverted due to #220

Issue #243, opened 2023-01-02
fix telefunc


samuelstroschein:
Telefunc functions are not being executed.

PR #244, opened 2023-01-02
243-fix-telefunc

samuelstroschein:
The middleware `express.text()` has previously been used in a global context with `app.use(express.text())`. Using the middleware in a global context is not desired as the middleware is telefunc specific. Upon moving the middleware in into the telefunc route "/_telefunc", `app.use()` has not been removed. The latter led to a silent drop in executing telefunc functions.

<a href="https://gitpod.io/#https://github.com/inlang/inlang/pull/244"></a>


samuelstroschein:
closes #243

PR #281, opened 2023-01-16
Improve messenger ux

jannesblobel:
Add a language Filter and Fix: forking-banner is not working   

<a href="https://gitpod.io/#https://github.com/inlang/inlang/pull/281"></a>

### DIFF
--- a/source-code/website/src/pages/editor/@host/@owner/@repository/Layout.tsx
+++ b/source-code/website/src/pages/editor/@host/@owner/@repository/Layout.tsx
@@ -219,7 +219,7 @@ function LanguageFilter() {
 				<sl-button slot="trigger" prop:caret={true}>
 					Languages
 				</sl-button>
-				<sl-menu class="min-w-full px-8 space-y-1">
+				<sl-menu class="min-w-full px-4 space-y-1">
 					<div class="flex space-x-6">
 						<a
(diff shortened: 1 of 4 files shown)


samuelstroschein:
## Problem 1: Inbalanced x and y axis.

The x and y axis are off.  The reference language has the language above and not to the left of the input field. That decision leads to an unbalanced UI. 

CleanShot 2023-01-16 at 08 59 53@2x

## Proposal:

1. Move the language back to the left for the reference language too.
2. Using paddings to achieve a balanced x and y-axis. 

One is quicker to implement.

## Problem 2: Imbalanced paddings, and typography

The paddings (1) and typography (2) in the filter are off. 

CleanShot 2023-01-16 at 09 04 26@2x

## Proposal 

Use `text-sm` and `font-medium` for the select all and reset all typography instead of a lighter font. That's how most other dropdowns do it and the light font looks disturbing. 

CleanShot 2023-01-16 at 09 07 10@2x

## Problem 3: The reset all button does not reset the selection

The default is to show all languages. The reset all button deselects all languages. Not the expected outcome. The inspiration has no languages selected by default. 
image


## Proposal

1. Reset selects all languages. 
2. Rename reset to "Deselect all"

I vote for proposal 2.

samuelstroschein (approved):
leeeets go. merge this baby

PR #293, opened 2023-01-20
208 add default error page

jannesblobel:
CleanShot 2023-01-20 at 13 55 43@2x

add Error Default page and fix paddings 

<a href="https://gitpod.io/#https://github.com/inlang/inlang/pull/293"></a>


samuelstroschein:
1. The icon of the button is wrong. That's an outbound arrow (leaving inlang.com). The arrow should be an left arrow (go back). 
2. Since this action is a go back, it's probably good to align the button on the left, not on the right. 

CleanShot 2023-01-20 at 14 47 09@2x

PR #299, opened 2023-01-22
fix messages alignment

ivanhofer:
before: 
image

after: 
image


samuelstroschein:
I believe the self-center property comes from centering the reference message if multiple other messages exist. Frankly, it looks questionable. The idea was to differentiate the reference message from other messages. @jannesblobel can provide more context. 

CleanShot 2023-01-22 at 12 07 10@2x

I am leaning towards just listing all messages beneath each other.

samuelstroschein:
Decided with @jannesblobel to switch back to the single-column view to avoid issues like this. A fix will come soon.

Issue #309, opened 2023-01-26
validate config 


samuelstroschein:
## Problem

1. Programatically testing the config is not possible e.g. "Does the `readResources` function work?", "Are my lints correct?". 
2. Plugin authors are starting to "hack" the inlang config. That's great. But will lead to unintended consequences and breaking changes in the future.

## Proposal

Provide a module (`inlang/core/test` ?) that can validate the config schema and functionality.  That module can be used by, for example, the CLI that provides an `inlang config validate` command and the editor that checks the config on the fly. 

- [ ] Validate the config schema with something like `zod`.
- [ ] "Test run" functions that are defined in the config
- [ ] Provide a CLI command like `inlang config validate` to test the config file programmatically
- [ ] Use regex to ban code in the config file that won't run in the browser like `import`, node globals, etc.

## Additional information

Using zod for the validation seems to make sense.

leolurch:
Would it make sense, to only allow importing plugins from one central authority like a inlang repo? This way you could make sure that plugin authors don't (intentionally or unintentionally) abuse their ability to run code on inlang editor clients.

samuelstroschein:
@leolurch

> Would it make sense, to only allow importing plugins from one central authority like a inlang repo?  

Would add too much overhead for us + the community. What we should do is to provide `@inlang/core/test` that plugin authors can use to test their plugins. 

> This way you could make sure that plugin authors don't (intentionally or unintentionally) abuse their ability to run code on inlang editor clients. 

Seems unlikely. We can't validate code for malicious intent. A plugin could use a malicious npm package which would even be out of the scope of the plugin itself. The right way to deal with security is sandboxing JS.

samuelstroschein:
@leolurch it could make sense to consolidate community plugins in a dedicated repo like solidjs does it see https://github.com/solidjs-community

PR #366, opened 2023-02-09
`@inlang/core/lint`

ivanhofer:
implements RFC https://github.com/inlang/inlang/pull/351
closes #351


samuelstroschein:
Let me know when I should review the PR

ivanhofer (commented):
@samuelstroschein please take another look. Meanwhile I'll work on the `createRuleCollection` functionality.

source-code/core/src/lint/reporter.ts
```diff
@@ -0,0 +1,50 @@
+import type { Resource, Message, Pattern } from '../ast/schema.js'
+import type { LintableNode, LintRuleId } from './rule.js'
+
+export type LintLevel = 'error' | 'warning'
+
+export type LintConfigSettings<Settings> = boolean | LintLevel | Settings
```
ivanhofer, line 6: currently has the wrong type

ivanhofer: fixed in latest commit

source-code/core/src/lint/rules/missingKey.ts
```diff
@@ -0,0 +1,24 @@
...
+	let referenceLanguage: string
+
+	return {
+		initialize: (config) => {
+			reporter = config.reporter
+			referenceLanguage = config.referenceLanguage
+		},
+		visitors: {
+			Resource: ({ target }) => {
+				if (target && target.languageTag.name === referenceLanguage) return 'skip'
```
ivanhofer, line 15: it could be a common use case to skip the processing of the reference resource. Should we add a flag to the configuration so rule authors don't need to add these 3 lines of code?

ivanhofer (commented):
@samuelstroschein I have created 3 rules. Do you think it is easy enough for authors to create their own rules? I have introduced a `createRule` utility function to make it even easier.

ivanhofer:
Other questions that came to my mind when looking at the global picture:
 - what should happen if someone adds the same rule twice? If we detect multiple rules with the same id, should we prevent the second rule to run and show a warning?
 - we initially said that each rule has an id that looks like this: `[prefix].[rule]` e.g. `'inlangStandardRules.missingKey'`. Should we allow rules to have just a regular string as an id e.g. `missingKey` and instead just add the prefix when using a collection of rules e.g. `inlangStandardRules`?
	But then the possibility of id collision is slightly bigger. See point above.
	`eslint` does not have this problem because all rules are added via a `plugin`. All included rules then get prefixed with the name of the plugin. Someone still could name a plugin identical to another and have the same conflict, but it is less likely.
	Or should we require rules to be included in a collection. That would be one additional function wrap:
	```ts
	export default createRuleCollection('inlangStandard', { missingKeyRule });
	```
 - should we use the same naming convention as [`eslint`](https://www.npmjs.com/package/eslint-plugin-jquery#usage) and [`stylelint`](https://github.com/stormwarning/stylelint-config-recess-order#advanced) have?
 `kebab-case` and `/` as a separator e.g. `inlang-standard-rules/missing-key`

Maybe a question for the future:
How can we ignore false positives or exceptions? We currently can only disable a whole rule but not for a single `Message`. A comment like `// eslint-ignore` will not work for translations. Maybe a solution like #355 is needed.

source-code/core/package.json
```diff
@@ -30,7 +30,11 @@
   "scripts": {
     "build": "tsc --build ./tsconfig.build.json",
     "dev": "tsc --watch",
-    "test": "vitest run --passWithNoTests --coverage",
+    "test": "tsc && vitest",
+    "tsc": "tsc --noEmit",
+    "tsc:watch": "tsc --noEmit",
```
samuelstroschein, line 35: Does build and dev not suffice?

source-code/core/src/lint/rules/additionalKey.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export const additionalKeyRule = createRule('inlang.additionalKey', () => {
+	let context: Context
+	let referenceLanguage: string
+
+	return {
+		initialize: (config) => {
+			context = config.context
+			referenceLanguage = config.referenceLanguage
+		},
```
samuelstroschein, line 12: I wonder if we can abstract more with the `createRule` utility (which seems like a great idea!). 

Initialize seems boilerplaty. For example, the `createRule` utility could pass the config and env (?) as arguments to the function. 

```ts
export const additionalKeyRule = createRule('inlang.additionalKey', (config) => {

	return {
        // 
    }
```

source-code/core/src/lint/rules/additionalKey.ts
```diff
@@ -0,0 +1,24 @@
+import type { Context } from '../context.js'
+import { createRule } from '../rule.js'
+
+export const additionalKeyRule = createRule('inlang.additionalKey', () => {
+	let context: Context
+	let referenceLanguage: string
+
+	return {
+		initialize: (config) => {
```
samuelstroschein, line 9: I am confused what `config` refers to.

I thought the inlang config is referred to but the config object seems to be a custom lint "config" object. Calling this `args` but be better whereas args is `{ env, context }`

source-code/core/src/lint/rules/additionalKey.ts
```diff
@@ -0,0 +1,24 @@
...
+		},
+		visitors: {
+			Resource: ({ target }) => {
+				if (target && target.languageTag.name === referenceLanguage) return 'skip'
+			},
+			Message: ({ target, reference }) => {
+				if (!reference && target) {
+					context.report(target, `Message with id '${target.id.name}' is specified, but missing in the reference`)
+				}
+			},
```
samuelstroschein, line 21: This seems like a great API. 

Super simple if reference does not contain target report something on `target`. I would have expected to report the level here too though: `context.report("warn", "Message with ...")`. The user could "overwrite" the level in the lint config.

source-code/core/src/lint/rules/additionalKey.ts
```diff
@@ -0,0 +1,24 @@
...
+			context = config.context
+			referenceLanguage = config.referenceLanguage
+		},
+		visitors: {
+			Resource: ({ target }) => {
+				if (target && target.languageTag.name === referenceLanguage) return 'skip'
+			},
+			Message: ({ target, reference }) => {
+				if (!reference && target) {
+					context.report(target, `Message with id '${target.id.name}' is specified, but missing in the reference`)
```
samuelstroschein, line 19: We should probably use objects instead of positional arguments to increase the flexibility to extend APIs like `context.report` in the future. 

```ts
// ❌ can only be extended with more positional arguments
//    every positonal argument must be provided
context.report(target, message, id)
context.report(target, message, undefined, level)

// ✅ mix and match
context.report({ target, message, id })
context.report({ target, message, level })
context.report({ level, target, id })

// ✅ easier deprecation
//    better target is just a dummy name for a property that replaces the fictional old `target`
//    not using positional arguments makes such breaking but not breaking change possible.
context.report({ betterTarget, message })
```

source-code/core/src/lint/rules/brandingRule.ts
```diff
@@ -0,0 +1,24 @@
...
+export const createBrandingRule = (brandName: string, incorrectBrandingNames: string[]) => createRule('inlang.brandingRule', () => {
+	let context: Context
+
+	return {
+		initialize: (config) => {
+			context = config.context
+		},
+		visitors: {
+			Pattern: ({ target }) => {
+				if (!target) return
```
samuelstroschein, line 13: Why is "skip" not returned as in the other rules?

source-code/core/src/lint/rules/additionalKey.ts
samuelstroschein: It leads to a tiny bit more boilerplate but the tradeoff of 

1. future-proofing the API 
2. (easier comprehension e.g. "ah blabla error is the message" `message: "Blabla error"`)

seems like a good tradeoff to do.

source-code/core/src/lint/_dummy.test.ts
```diff
@@ -106,8 +61,10 @@ const dummyConfig = {
 	writeResources: async () => undefined,
 	lint: {
 		rules: [
-			missingKeyRule('warning'),
-			additionalKeyRule(),
+			standardRules(),
+			// missingKeyRule('warning'),
+			// additionalKeyRule(true),
+			createBrandingRule('Red Bull', ['redbull', 'RedBull', 'Redbull'])(),
```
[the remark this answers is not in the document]
samuelstroschein: > This looks weird
> `createBrandingRule('Red Bull', ['redbull', 'RedBull', 'Redbull'])()`

Indeed. It is not clear what happens here. It also deviates from the schema we discussed where each rule needs to define the level e.g. 

```ts 
createBrandingRule("warn", 'Red Bull', ['redbull', 'RedBull', 'Redbull'])()
```

> don't have a concept of required params. The rule iself does not make sense if you don't configure it.
> Question: should we add the concept of required rules?

> we always need to pass the level before the rule specific settings. If you don't want to change the default level, you would need to pass true as the first parameter.
> Question: should we also accept the settings as the first and only parameter?

Did you forget the API we speced during the call? :D Seems to solve this discussion. The first parameter is the log level and must always be provided. The second argument is an object that every rule can define as they need. 


```ts
rules: [
  superGermanRules({
    siezen: ["warn", { minimumSentiment: 0.5 }],
    duzen: ["off"],
  }),
  myRule("warn", { someArgument: "foo" }),
],
```

source-code/core/src/lint/rules/missingKey.ts
samuelstroschein: I vote to refrain from adding sugar for now. Let's try to keep the surface area of the API as small as possible.

source-code/core/src/lint/linter.ts
```diff
@@ -0,0 +1,193 @@
...
+import { createContext } from './context.js';
+import { getLintRulesFromConfig, LintableNode, LintableNodeByType, ConfiguredLintRule, NodeVisitor, NodeVisitors, TargetReferenceParameterTuple } from './rule.js';
+
+const getResourceForLanguage = (resources: Resource[], language: string) =>
+	resources.find(({ languageTag }) => languageTag.name === language);
+
+export const lint = async (config: Config, env: EnvironmentFunctions) => {
+	const { referenceLanguage, languages, readResources } = config
+
+	const lintRules = getLintRulesFromConfig(config)
```
samuelstroschein, line 12: Using one liner functions like `getLintRulesFromConfig` makes the code harder to read and bloats the implementation IMHO. 

The following is much easier to reason about:

```ts
const lintRules = config.lint?.rules
// vs 
const lintRules = getLintRulesFromConfig(config)
```

Even if the array is flattened:

```ts
const lintRules = config.lint?.rules ? config.lint.rules.flat() : []
// or
const lintRules = (config.lint?.rules ?? []).flat()

// vs 
const lintRules = getLintRulesFromConfig(config)
```

The conditional logic of whether or not rules are undefined and the `lintRules` variable are not even required if the if statement is moved to the top of the function: 

```ts
export const lint = async (config: Config, env: EnvironmentFunction) => {
    if (config.lint?.rules === undefined) {
		console.warn('No lint rules specified. Aborting ...')
		return
	}
    // ...
    await Promise.all(
        // use config rules directly
		config.lint.rules.flat().map(lintRule =>
			processLintRule({ env, lintRule, referenceLanguage, languages, reference, resources })
		)
	)
}

**Review by samuelstroschein — COMMENTED** [review]
> what should happen if someone adds the same rule twice? If we detect multiple rules with the same id, should we prevent the second rule to run and show a warning?

We should probably implement https://github.com/inlang/inlang/issues/309, assume in the lint module that the config is correct and "do nothing". Otherwise, we end up with spagehetti code that tries to fix incorrect config properties. The spagehetti code problem applies to all modules that depent on config. Thus, fix the root problem: An incorrect inlang.config.js that contains two rules with the same identifier.

> we initially said that each rule has an id that looks like this: [prefix].[rule] e.g. 'inlangStandardRules.missingKey'. Should we allow rules to have just a regular string as an id e.g. missingKey and instead just add the prefix when using a collection of rules e.g. inlangStandardRules?

Reducing the API surface area. Thus, let's have one convention. The inconvenience of having to define a rule with `x.myRule` is relatively low. 

> should we use the same naming convention as eslint and stylelint have?

Given that we use camelCase everywhere, stick to camelCase? Seems like a minor issue. Whatever we define will be OK.

### Summary 

- defining a rule seems "good enough". The introduction of utilities like `createRule` could improve the DX in the future.

- the implementation seems complex. That might just be because of missing comments.

File: source-code/core/src/ast/schema.ts [hunk 11]
```diff
@@ -1,59 +1,77 @@
...
  * A resource is a collection of messages.
  */
-export type Resource = Node<"Resource"> & {
-  languageTag: LanguageTag;
-  body: Array<Message>;
-};
+export type Resource<Extension extends ExtensionInformation = ExtensionInformation> =
+  Node<"Resource", Extension> & {
+    languageTag: LanguageTag<Extension>;
+    body: Array<Message<Extension>>;
```
samuelstroschein, line 47: I saw a utility type on Twitter that could improve the readability of the Nodes with extensions on hovering.

source-code/core/src/lint/_dummy.test.ts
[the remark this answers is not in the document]
ivanhofer: > Did you forget the API we speced during the call? :D Seems to solve this discussion. The first parameter is the log level and must always be provided. The second argument is an object that every rule can define as they need.

As far as I remember we said that everything **can** be configured, but can also be left as it is (default). You don't have to do anything unless you want to change the defaults, or else you would have to configure a lot of things before you could use a collection of rules.

source-code/core/src/lint/rules/additionalKey.ts
ivanhofer: No it can't. You have to think what will be called by the regular `JavaScript` code (when you import the `inlang.config.js` file) and what get's called by the `lint` function. When configuring a rule, we can't inject any `inlang` concepts.
Unless we would do it like this:
```diff
config = {
-	lint: [
+ 	lint: () => [
 		additionalKeyRule('warning')
 	]
}
```
But I think this will cause more irritation as it is unusual for config files.

source-code/core/src/lint/rules/additionalKey.ts
ivanhofer: Forgotto rename this. Thanks!

source-code/core/src/lint/rules/additionalKey.ts
ivanhofer: > I would have expected to report the level here too though: `context.report("warn", "Message with ...")`. The user could "overwrite" the level in the lint config.

This already happens. If the `rule` author would specify the `level` on each message, then how would we be able to tell which one the user wants to override? A `rule` can only report a single `level`. The `rule` author can set a default level, but a user can override it.

source-code/core/src/lint/rules/brandingRule.ts
ivanhofer: Because here we also want to check the reference resource for errors. For the other rules, checking the reference does not make sense.

source-code/core/src/ast/schema.ts
ivanhofer: I was thinking about the same and tried it already. Didn't work and made the type more complex.

samuelstroschein: Sad to hear :(

source-code/core/src/lint/_dummy.test.ts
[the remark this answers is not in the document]
samuelstroschein: Correct, my bad! Forcing the first argument to be the _log level_ (or really, just an argument reserved for inlang) seems to be the way to go!

source-code/core/src/lint/rules/additionalKey.ts
samuelstroschein: > what get's called by the lint function.  

The lint function could inject the config. But as stated in the summary, I think this API "is good enough".

source-code/core/src/lint/rules/brandingRule.ts
samuelstroschein: Ah, is returning `undefined` like "continue with linting the other nodes" but don't "skip linting other rules"?  

```ts
if (!target) return 

// is "equivalent" to

if (!target) return "continue"
```

source-code/core/src/lint/rules/additionalKey.ts
ivanhofer: > The lint function could inject the config.

That's exactly what happens in the `initialize` function

source-code/core/src/lint/rules/brandingRule.ts
ivanhofer: yes, if
 - `'skip'` get's returned, then children will not be processed
 - `void` or `undefined` get's returned, then chidren will be preocessed with the payload that get's passed by the parent
 - somevalue get's returned, children will be process with that payload

ivanhofer: This pattern is called "early return" and prevents a nesting level in the implementation.
```ts
function doSomething() {
	if (condition) {
		console.log('Hello')
	}
}
```

vs.

```ts
function doSomething() {
	if (!condition) return

	console.log('Hello')
}
```

When you have a few lines of code inside the `if` statement, then the second variant is easier to read. And once you have multiple `if`s then you would end up with multiple levels of nesting. Therefore I prefer early returns.
_Except for single if statements where the condition is not trivial_

source-code/core/src/lint/linter.ts
ivanhofer: If someone would have this config:
```ts
{
	lint: []
}
```
then your code will not print a warning.

I could inline the `getLintRulesFromConfig` function, but if a line of code does multiple things (check if `lint` property is present, return the rules, flatten them), it should be it's own function with a meaningful name.

source-code/core/package.json
ivanhofer: `dev` outputs the code.
`tsc` will just check for errors.
You can see `tsc` as the `vitest` command for `TypeScript`. It will make more sense if we implement `type-definition` tests (see [example here](https://github.com/ivanhofer/exceptionally/blob/main/src/core/utils.test.types.ts)). There the `tsconfig` for `build` and `test` will be different.

source-code/core/src/lint/rules/additionalKey.ts
ivanhofer: Ok, I'll change that

ivanhofer:
> We should probably implement #309

Yes, I was thinking too about implementing this soon.

> Otherwise, we end up with spagehetti code that tries to fix incorrect config properties

I don't think this would be spaghetti code. It is just moving the checks to another place. Doing this at a central place makes sense.

> The inconvenience of having to define a rule with `x.myRule` is relatively low.

👍

> Given that we use camelCase everywhere, stick to camelCase?

👍

> * the implementation seems complex. That might just be because of missing comments.

Sure, once we decided on the final API, I can add the comments

source-code/core/src/lint/linter.ts
samuelstroschein: Adding a conditional check to the top level `if` seems like a potential fix?  

```ts
export const lint = async (config: Config, env: EnvironmentFunction) => {
    if (config.lint?.rules === undefined || config.lint.rules.length === 0) {
		console.warn('No lint rules specified. Aborting ...')
		return
	}
    const rules = config.rules.flat()
    // ...
}
```

samuelstroschein: Regardless of this specific example, I found the code harder to read than it should have but that might just be me. 

Hiding one line of code behind a function that is only used once leads to: 

- harder to reason about.  the reader needs to jump into the function instead of just "seeing" what the line of code does.
- more code. the function needs to be defined, imported, called. 
- (the calling function will not be pure, but this is minor)

ivanhofer: > ```ts
> if (config.lint?.rules === undefined || config.lint.rules.length === 0) {
> ```

And I find this harder to read and reason about than a function with the name `getLintRulesFromConfig` and a `length` check afterwards ^^.

> * harder to reason about.  the reader needs to jump into the function instead of just "seeing" what the line of code does.

The question is: does someone really have to know how lint rules are specified in the config? If I see some code that says `readInlangConfig`, I trust that function to do what it tells and I don't care about the implementation. With the beauty of `TypeScript` I can immediately see if the function returns what I would expect without ever seeng the code of that function.

> * more code. the function needs to be defined, imported, called.

The code will probably be minimized, and will end up with less characters than your suggested alternative above. Even without minification, the function approach takes an equal amount of characters. And a single function call will not make the code noticeable slower ;)

The benefit of doing this in it's own function is, that we can test it in isolation and if the config changes, we only need to update a single function.

If the name of the function is unclear, feel free to suggest an alternative.

ivanhofer:
Just something that came to my mind right now and I wanted to mention here.
Once we have something like a `context` object on a `Message` block (see https://github.com/inlang/inlang/discussions/341#discussioncomment-4873241), lint rules can be configured on a message per message base. Someone could mark a message with a "disable lint" option or he can say `lint.inlangStandardRules.messageLength.ratio: 1.5` to tell the `messageLength` rule to only show a warning if a translation for this specific message exceeds 1.5 times the reference. The possibilities are endless :)
This use case is already possible with the current implementation of the `lint` framework 🚀

source-code/core/src/lint/linter.ts
samuelstroschein: You are implementing this, thus go ahead and do what you find better.

This seems like a preference. I find a one-line function more difficult to read for something so trivial as a conditional. I'd rather see the conditional as a developer and know exactly what is going on.  

> does someone really have to know how lint rules are specified in the config? 

Yes, please. Makes it obvious where the rules are coming from. One core concept of UX is discoverability. Does a user understand how something works? 

Imagine that I am looking at this code on GitHub or similar and wonder where the rules are coming from. I am out of luck (if GitHubs new reference feature is not working) and have to do extra work to understand where rules are coming from. 

>> more code. the function needs to be defined, imported, called.
> The code will probably be minimized, and will end up with less characters than your suggested alternative above.

I was referring to code to maintain and comprehend, not the production build.

ivanhofer: As discussed privately, I moved the function to the same file to improve the point of "discoverability".

source-code/core/src/lint/README.md
```diff
@@ -0,0 +1,289 @@
...
+This package provides the functionality to automatically check `Resources` for potential errors and other insights.
+
+Use cases could be:
+
+- check if all `Messages` were translated to another language
+- check if the name of a brand always uses the same naming schema e.g. `MyApp` instead of `my-app` or `myApp`
+- check if the tone of a message does not sound aggressive
+- check if the translated message does not exceed a certain length
+- ... and probably a ton of other use cases. Please tell us if we should add one to this list.
+<!-- TODO: link to discussion -->
```
ivanhofer, line 12: do we want to open a discussion for lint rules?

ivanhofer (commented):
@samuelstroschein this PR is ready. Please check and let me know what you think

source-code/core/src/lint/index.ts
```diff
@@ -0,0 +1,24 @@
...
+  createLintRule,
+  type LintRule,
+  type LintRuleInitializer,
+  type LintRuleId,
+} from "./rule.js";
+export {
+  createLintRuleCollection,
+  type RuleCollectionInitializer,
+} from "./ruleCollection.js";
+
```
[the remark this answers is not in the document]
samuelstroschein: Do you mean lowering the surface area and waiting for users to demand this feature? 

I would vote yes. Instead of an issue, I would open a discussion though. Issues are tasks that should be worked on soonish.

source-code/core/src/lint/README.md
samuelstroschein: Do you refer to `standard-lint-rules` or opening a general discussion for more lint rules in https://github.com/inlang/inlang/discussions? 

If you refer to the latter, sounds reasonable. Opening discussions in `standard-lint-rules` would fragment the discussions too much. I am going to ask GitHub whether they can introduce a feature to let Discussions on repositories all refer to the main discussions.

source-code/core/src/lint/_utilities.ts
```diff
@@ -0,0 +1,12 @@
+import { inspect } from "util";
```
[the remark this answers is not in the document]
samuelstroschein: Up to you. I am not deep in the implementation to understand the benefits and downsides.

samuelstroschein (commented):
LGTM but I was not able to look at an example rule. 

Do you have a full example rule like missing translations to look at the syntax of defining a rule that visits nodes? 

FYI the tests are failing

CleanShot 2023-02-28 at 14 23 00@2x

samuelstroschein:
> Does the lint functionality also need to be added to the inlang documentation website?
https://inlang.com/documentation 

You could add the README to the tableofcontents https://github.com/inlang/inlang/blob/main/documentation/tableOfContents.ts

source-code/core/src/lint/_utilities.ts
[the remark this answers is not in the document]
ivanhofer: I have moved them to core/utilities

source-code/core/src/lint/index.ts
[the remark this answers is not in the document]
ivanhofer: Added the discussions #404, #405 and removed those comments.

source-code/core/src/lint/README.md
ivanhofer: created discussion #406

ivanhofer:
> Do you have a full example rule like missing translations to look at the syntax of defining a rule that visits nodes?

You find 3 rules in the [standard-lint-rules](https://github.com/inlang/standard-lint-rules) repo.
This PR needs to be merged first so I can finish the implenmentation and tests over there.

ivanhofer:
> You could add the README to the tableofcontents [`main`/documentation/tableOfContents.ts](https://github.com/inlang/inlang/blob/main/documentation/tableOfContents.ts?rgh-link-date=2023-02-28T13%3A52%3A24Z)

done.

I have resolved all points above.

Issue #133, opened 2022-08-27
preview message inline


samuelstroschein:
This feature exists in the current vs code extension already but will be re-implemented in anticipation of the architectural changes laid out in https://github.com/inlang/inlang/pull/127.

![image](https://user-images.githubusercontent.com/35429197/187033391-d92a077d-52e5-4f84-a5c8-68cd20b8ea26.png)

PR #372, opened 2023-02-14
feat(ide-extension): message extraction

openscript:
Open tasks:
- [x] Better error handling when query fails (not just `unwrap()`)

### DIFF
--- a/source-code/ide-extension/src/main.ts
+++ b/source-code/ide-extension/src/main.ts
@@ -1,12 +1,9 @@
 import * as vscode from "vscode";
-import { ExtractMessage } from "./actions/extractMessage.js";
-import { setState, state } from "./state.js";
+import { setState } from "./state.js";
 import { extractMessageCommand } from "./commands/extractMessage.js";
 import { inlinePattern } from "./decorations/inlinePattern.js";
 import { determineClosestPath } from "./utils/determineClosestPath.js";
 import type { Config as InlangConfig } from "@inlang/core/config";
-import { $import } from "@inlang/core/config";
-import fs from "node:fs/promises";
 
 export async function activate(
   context: vscode.ExtensionContext
@@ -58,11 +55,9 @@ async function main(args: { context: vscode.ExtensionContext }): Promise<void> {
   const configModule: InlangConfig = (await import(closestConfigPath)).config;
   setState({
     config: configModule,
-    configPath: closestConfigPath,
-    bundles: await configModule.readBundles({ $import, fs: fs as any }),
+    configPath: closestConfigPath
   });
-  const b = state().bundles;
-  console.log(b);
+
   // register the commands
   args.context.subscriptions.push(
     vscode.commands.registerCommand(
@@ -70,18 +65,7 @@ async function main(args: { context: vscode.ExtensionContext }): Promise<void> {
       extractMessageCommand.callback
     )
   );
-  // register the code actions
-  for (const language of state().config.ideExtension.documentSelectors) {
-    args.context.subscriptions.push(
-      vscode.languages.registerCodeActionsProvider(
-        language,
-        new ExtractMessage(),
-        {
-          providedCodeActionKinds: ExtractMessage.providedCodeActionKinds,
-        }
-      )
-    );
-  }
+
   // register decorations
   inlinePattern({ activeTextEditor });
 }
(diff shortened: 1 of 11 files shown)


openscript:
in  #133 the following will be done:
- manage resources inside the extension state

source-code/ide-extension/src/main.ts
```diff
@@ -55,33 +57,39 @@ async function main(args: { context: vscode.ExtensionContext }): Promise<void> {
     options: potentialConfigFileUris.map((uri) => uri.path),
     to: activeTextEditor.document.uri.path,
   });
-  const configModule: InlangConfig = (await import(closestConfigPath)).config;
+
+  // Change current working directory to configuration file directory, so relative paths to the languages files work.
+  // Otherwise the current working directory path is the directory where the node binary resides.
+  process.chdir(dirname(closestConfigPath));
```
openscript, line 63: Deprecates `configPath`.

samuelstroschein: Have you verified whether VSCode isolates extensions? Otherwise, we might be crashing other extensions with this line of code. 

Furthermore, I remember running into this problem. We need to develop our own filesystem implementation that does not rely on node internals with pluggable backends e.g. virtual filesystem, local storage filesystem, etc. That's part of https://github.com/inlang/inlang/issues/278

source-code/ide-extension/src/main.ts
```diff
@@ -55,33 +57,39 @@ async function main(args: { context: vscode.ExtensionContext }): Promise<void> {
...
+
+  // TODO: find better fs (vscode.workspace.fs)
+  const $import = initialize$import({ fs: fs.promises as any, fetch });
+  const configModule: InlangConfig = await (await import(closestConfigPath)).defineConfig({ $fs: fs.promises, $import });
```
samuelstroschein, line 67: I would split this line of code into two lines: 

```ts
const module = await import(closestConfigPath)
const config = await module.defineConfig({ $fs: fs.promises, $import })
```

For minor change suggestions like this one, should I go ahead and commit an improvement directly?

samuelstroschein (commented):
Aweeeesome to see the ide extension working again! 

## Open questions

### Should the callback be used as "name" and "id" be removed from `replacementOptions?

The developer experience could be improved by showing the result of the action instead of an "id". 

I extracted a message and was taken back for a few seconds when I got prompted whether I want to "extract a message as JSX or JS string". Answering the question leads to multiple seconds of thinking. The question collides with the result I want to achieve e.g. "reference the message in a t() function that is wrapped in curly brackets" and not "extract as JSX string". 

#### Proposal 

1. Show the result of the `callback` as a replacement option instead of the id/name in the extract message prompt. 
2. Remove `id` from the replacement option type but keep the replacement option as an object to allow the later introduction of `id`, `name`, or other properties. 

https://user-images.githubusercontent.com/35429197/219095619-a1f647aa-c6bb-4a74-a07a-6df10bf5dccd.mp4

source-code/ide-extension/src/main.ts
openscript: I'll change this one, but I'm totally fine if you commit into my branches :)

source-code/ide-extension/src/main.ts
openscript: After some thinking, I think it is still the best solution, if it works in all environments of the ide-extension. My reasons are:
-  The core shouldn't need to handle this, as the to be imported URI is provided from the outside and the outside should make sure that the URI work in its environment.
- When we start handling paths inside the core, we open a little Pandoras box. We would need to load `node:path` and also start to use one of the path polyfills for enabling the browser. 

If the currently implementation does it's job, we should continue work with it, otherwise I suggest a callback function on the `environment` passed to `$import`, which resolves given URIs.

Furthermore I've opened a discussion about this here: https://github.com/microsoft/vscode-discussions/discussions/466

I suggest to move onward with this solution for now.

samuelstroschein: Yes, let's move on. 

If we are not breaking other extensions, we should be fine. 

> The core shouldn't need to handle this, as the to be imported URI is provided from the outside and the outside should make sure that the URI work in its environment.

Agree. The filesystem implementation we are using should handle the working directory.

openscript:
Great proposal. I've implemented this:

[Kooha-2023-02-16-15-09-33.webm](https://user-images.githubusercontent.com/1105080/219389261-aa38446d-9b2a-4ff0-b37f-4e402d24a660.webm)

samuelstroschein:
LGTM. 

If ready for merge, go ahead and merge this PR :)

PR #376, opened 2023-02-19
feat(ide-extension): message preview

openscript:
To do:

- [x] Redecorate after message extraction (will be done later)
- [x] Truncate message preview
- [x] Strange behavior with the first character after the preview

### DIFF
--- a/source-code/core/src/config/schema.ts
+++ b/source-code/core/src/config/schema.ts
@@ -58,12 +58,12 @@ export type Config = {
 
   ideExtension?: {
     /**
-     * Defines finders for inlined translations.
+     * Defines finders for translations inside the code.
      *
      * @param args represents the data to conduct the search on
      * @returns a promise with matched inlined translations
      */
-    inlinedTranslationFinders?: (args: {
+    translationFinders: ((args: {
       documentText: string;
     }) => Promise<
       Array<{
@@ -79,7 +79,7 @@ export type Config = {
           end: { line: number; character: number };
         };
       }>
-    >[];
+    >)[];
 
     /**
      * Defines the options to extract messages.
--- a/source-code/core/src/config/schema.ts
+++ b/source-code/core/src/config/schema.ts
@@ -58,18 +58,6 @@ export type Config = {
 
   ideExtension?: {
     /**
-     * Defines how message previews look like.
-     */
-    preview?: {
-      /**
-       * Defines the maximum length of a preview before it's truncated.
-       *
-       * If not defined, the message previews are not truncated.
-       * If set to lower than 1, message previews are deactivated.
-       */
-      maximumLength?: number;
-    }
-    /**
      * Defines matchers for message references inside the code.
      *
      * @param args represents the data to conduct the search on
--- a/source-code/core/src/config/schema.ts
+++ b/source-code/core/src/config/schema.ts
@@ -58,6 +58,18 @@ export type Config = {
 
   ideExtension?: {
     /**
+     * Defines how message previews look like.
+     */
+    preview?: {
+      /**
+       * Defines the maximum length of a preview before it's truncated.
+       *
+       * If not defined, the message previews are not truncated.
+       * If set to lower than 1, message previews are deactivated.
+       */
+      maximumLength?: number;
+    }
+    /**
      * Defines matchers for message references inside the code.
      *
      * @param args represents the data to conduct the search on
--- a/source-code/core/src/config/schema.ts
+++ b/source-code/core/src/config/schema.ts
@@ -58,12 +58,12 @@ export type Config = {
 
   ideExtension?: {
     /**
-     * Defines finders for translations inside the code.
+     * Defines matchers for message references inside the code.
      *
      * @param args represents the data to conduct the search on
-     * @returns a promise with matched inlined translations
+     * @returns a promise with matched message references
      */
-    messageFinders: ((args: {
+    messageReferenceMatchers: ((args: {
       documentText: string;
     }) => Promise<
       Array<{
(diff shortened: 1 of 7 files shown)


samuelstroschein:
I will review the PR tomorrow/today (20 Feb).

Please let me know when slow PR reviews are blocking you. I am packed with tasks and prioritize on the go. If I don't receive a signal that someone is blocked by me, I will not immediately do a task.

samuelstroschein (commented):
Nice, the feature works! 

A few hiccups occurred while testing the feature:


- [x] the finder is broken. shows multiple annotations that should not be matched.

CleanShot 2023-02-20 at 12 27 12@2x

- [ ] the inlang.config.js should probably not be matched against

The extraction patterns are (of course :D) matched.

CleanShot 2023-02-20 at 12 32 38@2x


- [x] is "matchers" a better word than "finders"? 

I found myself thinking and describing "finding" messages as "matching" patterns in source code. The mental disconnect between the word and what happens here could be removed by using the word `match[ing]` instead of `finder`. I suspect that the change slightly increases comprehension and expandability of what the property does.

`config.ideExtension.messageFinders` would become `config.ideExtension.messageMatchers` 

- [x] "messageFinders|messageMatchers" is slightly ambiguous

I found myself to be unsure whether "messageFinders" refers to "[find and] extract a message" or "find [references of] messages [in source code]" when I wrote https://github.com/inlang/inlang/discussions/377#discussioncomment-5052432. 

The problem could be mitigated by naming the property `messageReferenceMatchers`, `messageUsageMatchers`, or similar.MessageReferenceMatchers might be ambiguous because of the "referenceLanguage" concept.

openscript:
Thank you for the feedback.

In my opinion it should be possible to configure a list of files which are included or excluded. So in the `inlang.config.js` there is something like:

```js
{
   // ....
   ideExtension: {
     excludesFromMatchers: ['inlang.config.js', 'node_modules/**/*']
   }
   // ...
}
```

samuelstroschein:
> In my opinion it should be possible to configure a list of files which are included or excluded. So in the `inlang.config.js` there is something like

Agree! 

The `inlang.config.js` file will likely not be the only file people want to have ignored.

openscript:
Do other Inlang components need the `exclude` function?

samuelstroschein:
> Do other Inlang components need the exclude function? Linter? Editor? 

Likely.  

For now, I would put everything under the `ideExtension.x` property. Once the need arises, we can move properties that are used across inlang out of `ideExtension.x` with a deprecation warning for the old property and removal upon v1.0

openscript:
> > Do other Inlang components need the exclude function? Linter? Editor?
> 
> Likely.
> 
> For now, I would put everything under the `ideExtension.x` property. Once the need arises, we can move properties that are used across inlang out of `ideExtension.x` with a deprecation warning for the old property and removal upon v1.0

I'll handle this in #382

openscript:
When the user extract a message, the decorations are not updated. Right now this is hard to fix. After calling `readResources` we should create file watches to all resources. When these files change, we should redecorate.
`readResources` should tell which files it has loaded. This involves quite a few changes, which isn't really linked to the message preview feature.

source-code/core/src/config/schema.ts
```diff
@@ -57,6 +57,18 @@ export type Config = {
   experimental?: object;
 
   ideExtension?: {
+    /**
+     * Defines how message previews look like.
+     */
+    preview?: {
```
samuelstroschein, line 63: Do you expect that people will need to customize the message preview?

samuelstroschein (commented):
As mentioned in https://github.com/inlang/example/pull/15#issuecomment-1437362704, I am hesitant of merging this PR because of the additional config property `ideExtension.preview`. Let's resolve the discussion in the other issue.

samuelstroschein:
> readResources should tell which files it has loaded. This involves quite a few changes, which isn't really linked to the message preview feature. 

That's the custom `fs` implementation I referred to several times. The custom `fs`  implementation could report what files have been loaded which is important for sparse checkout too. In the meantime, the effect can be achieved with a proxy.

source-code/core/src/config/schema.ts
openscript: Will be removed and hard coded for now.

samuelstroschein (commented):
The current working directory change which might break other extensions is still included. Can we merge this PR until this is resolved? 

```ts
process.chdir(dirname(closestConfigPath));
```
_https://github.com/inlang/inlang/pull/376/files#diff-c93698b8be10319741e6970431849fba35d2f1dffd63d26ea6ed9dd12fab8413R65_

EDIT: I assume merge this PR, tick the task as complete, and then move on to the next task?

openscript:
> The current working directory change which might break other extensions is still included. Can we merge this PR until this is resolved?
> 
> ```ts
> process.chdir(dirname(closestConfigPath));
> ```
> 
> _https://github.com/inlang/inlang/pull/376/files#diff-c93698b8be10319741e6970431849fba35d2f1dffd63d26ea6ed9dd12fab8413R65_
> 
> EDIT: I assume merge this PR, tick the task as complete, and then move on to the next task?

This depends on https://github.com/orgs/inlang/discussions/373

samuelstroschein:
Merging into `main` indicates to me "production ready" but we are moving step-by-step for the ide extension since it's not released right?

openscript:
I agree that it feels odd. As long as the version is below 1.0.0 and we can keep track of this things, it's kind of ok in my opinion.

Another option is to have a `develop` branch and push it there, but I'm not sure if the new fs is ready until beginning of March, where VS Code 1.76 is released with - hopefully - the new Electron version. Furthermore you don't want to release something from the `develop` branch as well.

samuelstroschein:
Got it. Let's merge to main!

PR #411, opened 2023-03-02
Imrpove getting started docs

jannesblobel:
CleanShot 2023-03-02 at 12 10 37@2x

### DIFF
--- a/source-code/website/src/services/markdown/src/tags/Figure.tsx
+++ b/source-code/website/src/services/markdown/src/tags/Figure.tsx
@@ -1,7 +1,18 @@
-export function Figure(props: { src: string; alt: string; caption: string }) {
+export function Figure(props: {
+  src: string;
+  alt: string;
+  caption: string;
+  height: string;
+  width: string;
+}) {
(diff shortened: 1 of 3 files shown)


source-code/website/src/services/markdown/src/tags/Figure.tsx
```diff
@@ -1,7 +1,18 @@
-export function Figure(props: { src: string; alt: string; caption: string }) {
+export function Figure(props: {
+  src: string;
+  alt: string;
+  caption: string;
+  height: string;
```
samuelstroschein, line 5: Good but hardcoded width height values break responsiveness. 

Can you please open a follow up PR that removes `width` and `height` again/replaces it with responsive values?

samuelstroschein: @jannesblobel

jannesblobel: https://github.com/inlang/inlang/pull/414

PR #412, opened 2023-03-02
Revert "Imrpove getting started docs"

samuelstroschein:
Reverts inlang/inlang#411


source-code/website/src/services/markdown/src/tags/Figure.tsx
```diff
@@ -1,18 +1,7 @@
-export function Figure(props: {
-  src: string;
-  alt: string;
-  caption: string;
-  height: string;
```
samuelstroschein, line 5: this breaks responsiveness.

samuelstroschein: no hardcoded width height values

PR #418, opened 2023-03-02
improves and adds some missing lint utilities

ivanhofer:


source-code/core/src/lint/README.md
```diff
@@ -16,7 +16,6 @@ Use cases could be:
 - check if the translated message does not exceed a certain length
```
ivanhofer, line 16: this file only contains formatting related changes

PR #423, opened 2023-03-04
improve linting

ivanhofer:


source-code/core/src/utilities/debug.ts
```diff
@@ -1,12 +1,4 @@
-import { inspect } from "util";
```
ivanhofer, line 1: this single line is actually why I reated this PR ^^

source-code/core/src/utilities/polyfills/util.ts
```diff
@@ -0,0 +1,468 @@
+// @ts-nocheck
```
ivanhofer, line 1: this file creates is the polyfill to be able to log an object in a pretty way instead of `'[object Object]'`

package.json
```diff
@@ -8,8 +8,19 @@
   "scripts": {
     "build": "turbo run build",
     "dev": "turbo run dev",
-    "clean": "npm run clean -ws && rm -f -r ./node_modules",
+    "---- TEST ----------------------------------------------------------": "",
     "test": "turbo run test",
+    "---- LINT ----------------------------------------------------------": "",
+    "lint:base": "eslint --ignore-path .gitignore --cache",
```
ivanhofer, line 14: looks a bit weird first, but it enables us to have the same config for all packages. Or else we would need to check if each individual script uses the correct ignore path, cache option etc.

ivanhofer:
CI is currently failing because:
 - not all code is formatted correctly
   I would wait until this PR get's approved to run `format:fix` on all files. Or else we will have a lot of files that only contain style changes
- `website` and `ide-extension` use node apis. Is this intended?
	If yes, should we add a `eslint-ignore` comment on top of the imports or disable the lint rule comletely for those packages?

ivanhofer:
I would suggest to also add [`lint-staged`](https://github.com/okonet/lint-staged) in a next PR to automatically fix lint and formatting related errors on each commit. It will take a few seconds until a commit is completed, but it increases the feedback loop a lot.

PR #443, opened 2023-03-09
validate-inlang-schema

### LINKED ISSUE #309: validate config

samuelstroschein:
closes #309

This PR is not perfect. 

- the test coverage is not 100%
- the code is a big ugly 

But, I want to get #436 out this week.


samuelstroschein:
Will merge this PR tomorrow around 11 am CEST if nobody objects.

PR #446, opened 2023-03-14
Set up continuous deployment for ide-extension

openscript:
### No scoped extension names
VSCodes Marketplace doesn't support scoped names like `@inlang/ide-extension`. `vsce` infers the name from `package.json`. This leads to: `ERROR  Invalid extension name '@inlang/ide-extension'`.

https://github.com/microsoft/vscode-vsce/issues/186

### VSCE adds .js to entry points

Because inlang are ESMs, but ide-extension cannot be ESM, we use CommonJS for it with `.cjs` extension to declare this. Prior `2.18.0` of `vsce` it added `.js` to the entrypoint and eventually searching for a `main.cjs.js` file. Luckily this is fixed in the most recent version.

https://github.com/microsoft/vscode-vsce/pull/825

### Monorepo 

`vsce` tries to package parent directories as well. This isn't necessary and leads to errors (naming collisions in node_modules), so they should be ignored inside `.vscodeignore`:

```
../../**
../**
```

### DIFF
--- a/source-code/ide-extension/package.json
+++ b/source-code/ide-extension/package.json
@@ -1,5 +1,5 @@
 {
-	"name": "inlang",
+	"name": "@inlang/ide-extension",
 	"private": true,
 	"displayName": "inlang",
 	"type": "module",
--- a/source-code/ide-extension/package.json
+++ b/source-code/ide-extension/package.json
@@ -1,10 +1,10 @@
 {
-	"name": "@inlang/ide-extension",
+	"name": "inlang",
 	"private": true,
 	"displayName": "inlang",
 	"type": "module",
 	"description": "i18n extension that makes you more productive.",
-	"publisher": "Inlang",
+	"publisher": "inlang",
 	"repository": {
 		"type": "git",
 		"url": "https://github.com/inlang/inlang.git"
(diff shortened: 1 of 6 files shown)


source-code/ide-extension/package.json
```diff
@@ -1,18 +1,18 @@
 {
-	"name": "@inlang/ide-extension",
+	"name": "inlang",
```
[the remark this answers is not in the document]
openscript: It was a mistake that I pushed this. There is a `script` which renames the name upon build.

source-code/ide-extension/scripts/preparePackage.js
```diff
@@ -0,0 +1,7 @@
+import("node:fs").then((fs) => {
```
samuelstroschein, line 1: why are you not using a regular import?

source-code/ide-extension/package.json
samuelstroschein: the name is still inlang. it should be "vs-code-extension"

openscript: That is weird:
image

https://github.com/inlang/inlang/pull/446/commits/742ad605e4f42b102e18f540af0d494437afe7a5

samuelstroschein: Okay let's merge. Either it deplys or it doesn't

samuelstroschein:
@openscript  Feel free to merge

PR #450, opened 2023-03-16
Add RFC for choosing between isomorphic-git and libgit2  

nikitavoloboev:


samuelstroschein (changes requested):
The RFC lacks architectural evaluations and implications. 

The missing features of isomorphic git are not a concern. The question is whether or not we build out git-sdk on pure JS with a headstart because we can fork isomorphic git, or switch to ligbit2 which requires WebAssembly, different toolchains, but is battle-tested and "fast". 

However, if the JS implementation is fast enough because network calls are the bottleneck, the speed advantage of libgit2 is neglectable. The iteration speed and reduced operational complexity of a JS implementation are likely favorable.

samuelstroschein:
@nikitavoloboev close this one in favor of #455 ?

PR #455, opened 2023-03-20
RFC git-sdk requirements

samuelstroschein:
[Rendered](https://github.com/inlang/inlang/blob/rfc-git-sdk-requirements/rfcs/git-sdk-requirements/RFC.md)

Give feedback by reviewing the RFC as file [link](https://github.com/inlang/inlang/pull/455/files)


samuelstroschein:
@nikitavoloboev This RFC is work in progress. I intend that we complete the RFC together with your finding and evaluation of isomorphic git vs libgit2. 

We are facing high ambiguity. Thus, I put the most emphasis on iteration speed and left some notes on what answers I would try to seek from a comparison of a JS vs WebAssembly implementation.

rfcs/git-sdk-requirements/RFC.md
```diff
@@ -0,0 +1,122 @@
...
+// does not clone the entire repository.
+// only metadata that enable other git commands to run
+await clone("https://github.com/inlang/inlang", fs)
+
+// lazy fetches the file and git commit history of README.md
+const commitHistory = await history("/README.md", fs)
+
+// lazy fetches xyz file
+const readme = await fs.readFile("/xyz.md")
+```
```
nikitavoloboev, line 54: This is a nice API and idea. 

As I read this it becomes more clear to me that the answer to the RFC should be whatever codebase we can modify easier. Isomorphic Git is clear winner here compared to libgit2 written in C. 

As another option I thought of compiling [go-git](https://github.com/go-git/go-git) to WASM via [TinyGo](https://tinygo.org). Go code is much easier to extend than C.  

However I am confident now that Isomoprhic Git should allow us to build this kind of lazy loading feature and more without disregarding web performance and velocity of adding new features.

samuelstroschein: > answer to the RFC should be whatever codebase we can modify easier. 

I also realized that. Iteration speed should be the focus. Having a heavy toolchain that includes compiling C or GO to WASM will slow down iteration speeds. 

> However I am confident now that Isomoprhic Git should allow us to build this kind of lazy loading feature and more without disregarding web performance and velocity of adding new features. 

I am looking forward to your completion of the RFC! :)

rfcs/git-sdk-requirements/RFC.md
```diff
@@ -0,0 +1,697 @@
...
+
+Goal 5: (Future) Support for large files out of the box [Medium Confidence] 🚧
+
+> note: need to research more to give good estimate how doable this is to add
+
+There is [open issue on this](https://github.com/isomorphic-git/isomorphic-git/issues/1375).
+
+Goal 6: (Future) Plugin system to support different files [Medium Confidence] 🚧
+
+Would be doable to add irrespective of whether libgit2 or Isomorphic Git is chosen.
```
samuelstroschein, line 210: Please elaborate. 

I assume that plugins will be written in JS (that can call WASM under the hood). A JS implementation of git seems to serve us better because we can natively import JS plugins. Choosing libgit2 would likely complicate things.

rfcs/git-sdk-requirements/RFC.md
```diff
@@ -0,0 +1,697 @@
...
+```
+
+And you get back a fully fledged file system fetched from the repo. As the SDK is focused on running on context of browser.
+
+Above function call can do a fetch of just enough resources to do further actions.
+
+`git clone --depth 1 --sparse --no-checkout --filter=blob:none https://github.com/inlang/inlang`
+
+After this an app would most likely want to fetch a file to edit. It can do:
+
```
samuelstroschein, line 239: Abstracting the filesystem away makes the git sdk lose one of its appeals. 

Files (and thereby filesystems) are amazing and enable inlang to exist. The VSCode extension uses a local filesystem provided by VSCode. The browser uses a filesystem that fetches files remotely. The CLI uses a filesystem provided by node.    

If the git-sdk would abstract away the filesystem, using the git-sdk outside of the browser becomes a burden. If users have their files locally on disk, there is no appealing reason to force a remote filesystem.

rfcs/git-sdk-requirements/RFC.md
```diff
@@ -0,0 +1,697 @@
...
+The API for above is quite readable and nice. And powerful too in some ways. As you don't have to think about how fetch happens. It's done for you automatically.
+
+```js
+const inlang = await gitSDK.clone("https://github.com/inlang/inlang.git", { depth: 1 })
+```
+
+You can also as second arg to `clone` send some options to already fetch up to certain depth of commits or any other option that would make sense to run for Git to be run in the browser.
+
+In many ways this opens up the API surface too and lets us explore what to expose to users and what not. A lot of complexity can be hidden away. Namely all the memfs api learning you will need to do, together with more lines of code and potential bugs.
```
samuelstroschein, line 297: I don't see a learning requirement with filesystems.

Yes, we use `memfs` _in the browser_ right now. But there is no learning. Just call `readFile()` as one would do in a local environment like Node.js.  

Related to a remark earlier, abstracting the filesystem away makes the git-sdk less versatile for no immediate upside other than saving one from passing the filesystem as arguments to git commands like `clone(fs)`.

rfcs/git-sdk-requirements/RFC.md
```diff
@@ -0,0 +1,697 @@
...
+signIn()
+signOut()
+createFork()
+syncFork()
+openPullRequest()
+```
+
+#### Q: Is 833 KB WASM file an issue to load?
+
+Should be non issue as its under 1 MB.
```
samuelstroschein: What makes you believe that? 

Assuming that SSR is important, what implications will an 800KB WASM file have on SEO? Will a JS implementation avoid large bundle sizes?

rfcs/git-sdk-requirements/RFC.md
```diff
@@ -0,0 +1,697 @@
...
+
+#### Q: Can you keep the file system inside WASM too?
+
+No, you can't. There is no file system in the WebAssembly runtime.
```
samuelstroschein, line 381: There is. I am aware that EMSCripten is working on a WASM file system.

rfcs/git-sdk-requirements/RFC.md
```diff
@@ -0,0 +1,697 @@
...
+FS.chdir("inlang")
+```
+
+You can expose different C functions that run the commands. For example, we have built [open pr for shallow clones](https://github.com/libgit2/libgit2/pull/6396) and exposed it via a new method for `callMain`:
+
+```js
+const result = libgit.callMain(["shallowClone", "https://github.com/inlang/inlang.git", "inlang"])
+```
+
+That function was same as clone, just also set `clone_opts.fetch_opts.depth` option before doing the clone. Extending libgit2 with new commands would work in the same way.
```
samuelstroschein: This is a usage explanation. Not an architectural exploration. 

What happens under the hood? What approach will make the usage of the git-sdk easier? The surface APIs like `libgit.callMain` can always be changed.

rfcs/git-sdk-requirements/RFC.md
```diff
@@ -0,0 +1,697 @@
...
+
+Goal 3: Must be git compatible but not necessarily up to spec [Medium Confidence] ✅
+
+No issue here for same reason as Isomorphic Git. We use commands from libgit2 we need.
+
+However the barrier to contributing to libgit2 itself is going to be that command is fully spec compliant so we need to be aware of that.
+
+Goal 4: (Future) File-based auth [High Confidence | Server-related ] 🚧
+
+Is part of server so outside of scope.
```
samuelstroschein, line 483: FYI the server side implementation will likely use this git-sdk too.

rfcs/git-sdk-requirements/RFC.md
```diff
@@ -0,0 +1,697 @@
...
+
+Both features need some git know how and time to implement.
+
+With libgit2 we get those features out of the box.
+
+Performance is not a concern in either case so the other focus is on DX.
+
+I think you can provide the same Git SDK API regardless whether Isomorphic Git is chosen or libgit2 to do git operations.
+
+As for regretting choose either technology in the future. We would potentially need to think through different kinds of applications that will be built with Git SDK. Some are done as part of this RFC below.
```
samuelstroschein, line 507: > Having written above, I believe going with libgit2 compiled to WASM option will allow us to potentially move faster in near future. **As we won't need to implement sparse-checkout and rebase.** 

This conclusion is missing the goal of this RFC. This RFC is not about libgit2 has more features. This RFC is about choosing the right architecture for the next years. Implementing sparse-checkout and rebase in JS shouldn't be an issue.

samuelstroschein (commented):
I am missing deeper technical comparisons between both approaches.  The following sentence from the RFC reflects my criticism:

> The reason you'd want to move to libgit2 is that libgit2 supports more options out of the box. And potentially performance.

*I* don't want to move to libgit2. *We* need to elaborate on the architectural implications of choosing one approach over the other. What does libgit2 supporting more features imply? For example: 

> Libgit2 supporting more features likely allows for faster iteration cycles. The git-sdk can combine existing libgit2 commands to provide new functionality.

Let's make the counterargument for isomorphic git: 

> While libgit2 may enable us to glue together existing commands, there is no need to glue together existing commands. Git stores information as text files. The git-sdk does not need up-to-spec commands. Thus, the git-sdk could directly modify the text files in the `.git` directory and avoid the gluing together of commands. A JS implementation might lead to faster iteration cycles because we will need to modify git text files regardless of libgit2. If that is the case, a JS implementation is anticipated to yield faster results. Debugging is easier and testing is easier.

samuelstroschein:
From https://news.ycombinator.com/item?id=35288059. 

Apparently, EMScripten is unmaintained. If true, this should play a role in our decision process. 

CleanShot 2023-03-24 at 18 30 44@2x

samuelstroschein:
Directly committed https://github.com/inlang/inlang/blob/main/rfcs/git-sdk-requirements/RFC.md

PR #459, opened 2023-03-21
refactor: simplify core/lint

samuelstroschein:
## This PR introduces the following changes

_All changes intend to reduce complexity, mostly by exposing less features, and wait for user feedback on edge cases before we expose features that might not align with user needs but need to be maintained by us. As a side effect, the DX is expected to increase due to simpler docs._

### remove setup ☑️ 

The `createLintRule` function is the setup function, see https://github.com/inlang/inlang/issues/457#issue-1631994310.  

### remove teardown ☑️ 

See https://github.com/inlang/inlang/issues/457#issuecomment-1476138324

### remove the concept of `payload` ☑️ 

The concept of payload is redundant due to the `createLintRule` function. 

State exists within the scope of the `createLintRule` function. There is no need for a dedicated payload concept that needs to be explained and maintained. 

```diff
const rule = createLintRule("error.rule", "error", () => {

+ const patternIsInvalid = false;

 return {
  visitors: {
   Pattern: ({ target, report }) => {
    report({ node: target, message: "Test" });
+    patternIsInvalid = true;
   },
   Resource: ({ target, report }) => {
+    if (patternIsInvalid) {
     report({ message: "Error in Pattern" });
    }
   },
  },
 };
});

```

### remove the concept of booleans for a rule level ☑️ 

The concept of passing booleans to a rule level unjustifiably increases comprehension and code-base complexity.  There were two reasons for passing booleans to a lint rule: 

1. True indicates that the default level of the rule should be used.  

This concept has been refactored to use "default" instead. 

```ts
// uses the standard lint level and passes custom settings to the rule
missingKeyRule(true, { threshold: 4 }),
```

```ts
// uses the standard lint level and passes custom settings to the rule
missingKeyRule("default", { threshold: 4 }),
```

4. To disable rules. 

Disabling rules with a boolean flag is unnecessary. Outcommenting a rule is more intuitive and decreases the complexity of lint.

```ts 
rules: {
  // disabling a rule
  missingKeyRule(false)
}
```
```ts
rules: {
  // Disabling a rule by commenting it out 😅

  // missingKeyRule()
}
```

For conditionals, inline conditionals can be used. Given, the syntax is inferior. But, will users use conditionals for rules? If so, they can achieve the desired effect in plain JS. If users strongly demand disabling rules via the lint level, we can re-introduce it.

```ts
rules: {
  // disables the rule if it runs on Mondays
  missingKeyRule(!isTodayMonday),
}
```

```ts
rules: {
  // disables the rule if it runs on Mondays
  ...!isTodayMonday ? [missingKeyrule()] : [],
}
```

### replace the concept of `context` with top-level arguments ☑️

The create lint rule exposes a report function and partial inlang config in the callback with the argument object pattern. 

- removes the concept of context which has only been used for `report` -> less things to explain
- the (partial) config is now exposed as object, allowing us to expose more config options if user request them
- the callback now takes an object of arguments instead of a positional `settings` argument, allowing us to extend the callback arguments based on user requests. 

```diff

export const additionalKeyRule: LintRuleInitializer = createLintRule(
  "inlang.additionalKey",
  "warn",
+  ({ report, config }) => {
      visitors: {
        Resource: ({ target }) => {
          if (target && target.languageTag.name === config.referenceLanguage)
            return "skip";
        },
        },
      },
    };
  }
);

```

## Additional information 

It seems like complexity originated because the `createLintRule` was an afterthought [that made original concepts redundant]. 

### [Problem] Not a single end-to-end test existed; the majority of tests had little value 

- Over 90% of all tests passed even though the implementation was completely broken at one point. 
- No end-to-end test existed. 

Many tests used mock rules even though the difference between setting up a mock rule and a real rule is neglectable:

#### Example 1

used in the tests

```ts
const errorRule = {
	id: "error.rule",
	level: "error",
	visitors: {
		Pattern: vi.fn(),
	},
} satisfies LintRule
```  

alternative

```ts
const errorRule = createLintRule("error.rule", "error", () => {
 return {
  visitors: {
    Pattern: ({target}) => report({node: target, message: "Error"})
  },
 };
});
```

#### Example 2

The following test passed even though the implementation was broken. Visitors have been replaced with an empty object at one point. The test passed regardless due to the artificial testing conditions. A real-world usage would pass object with visitors, subsequently failing this test.

```ts
const visitors = {};

const rule1 = createLintRule("my.id", "error", () => {
 return {
  visitors,
 };
});

test("calling the function should return an object with the passed visitors", async () => {
 const configuredRule = rule1();

 expect(configuredRule.visitors).toBe(visitors);
});
```

#### Learning

- Prefer end-to-end tests over unit tests. 
- Focus on code coverage instead of writing (duplicate) unit tests for every function like "for 'Message' test if 'leave' is not present" and "for 'Pattern' test if 'leave' is not present". The test is identical. Writing and maintaining both tests seems unnecessary. The underlying logic is identical.


samuelstroschein:
## Open questions 

- [ ] should lint return errors? 

I encountered numerous problems because `lint` implicitly catches errors. That behavior makes sense. Just because one rule is throwing, the linting process shouldn't abort. However, debugging became harder and consuming apps like the editor or CLI will never now if a linting process failed. 

An easy fix could be to return a tuple: 

```ts
const [ resources, errors ] = await lint()
``` 

But the following pattern wouldn't work anymore (likely doesn't anyways because resources would have type `Resource[]` not `ResourceWithLints[]` 

```ts
let resources = readResources()
resources = await lint(resources)
```

If we decide for a `Result` pattern, I propose overhauling the current `Result` type to be a simple tuple return as well.

samuelstroschein:
@ivanhofer Props to your type writing skills! Took me hours to understand some but impressive what degree of type safety you get source code. Sometimes going a bit over the edge though ;) 

I think the lint design is more future-proof and easier to explain now. Docs are not ready yet. I am awaiting your input on the open questions before I continue.

ivanhofer (changes requested):
I'm sorry that you had to make that much changes. When you told me to create a sophisticated lint process, I included everything that came to my mind to be sure to cover 99% of all needs and hopefully "never" touch it again or introduce breaking changes.

And that's the reason why some things are more complicated then they could be.

In reality you did want a simple linting process. 

Some notes on your points:

### remove setup

`setup` is meant for initializing something asynchronously. If you need to make fetch calls to an API, you may need to authorize first. Not every API will have a dedicated npm package. Without the setup function, you would need to write special logic to do the authorization before the first API call. It was also meant to pass options that are meant for the whole linting process.

We could make the `setup` function optional to still support that use case. Passing those options mentioned above can also happen at another place.

### remove teardown

Was meant to free up resources. It is optional an can be used if someone needs it. Nobody is required to use it, if he doesn't need it.

### remove the concept of payload

Again, something optional that can be used if someone needs it. It is not required to use it.

Your code will not work, because `Pattern` is called after `Resource`. If you did want to support this, we would need the `leave` functionality again. And for exact such use cases the `leave` functionality is meant for. Most people won't use it, but if they need it it is there.

Another thing regarding your code example. This will work fine if you just have a single `Pattern` and `Resource`, but you will probably have hundreds of `Nodes`. How do you do that? Write hundreds of variables? Create an array? How do you know what index is the correct one?

With the `payload` concept you could pass the `Resource` to the `Pattern` visitor and directly call the `report` function there.

Or maybe just throw an error if > 5 `Patterns` contain bad language. Maybe you are fine with a single one or you are in a migration process and want to be sure that no additional error get's added, but you are temporary fine with the existing ones. But you don't want to fail your CI process.

### remove the concept of booleans for a rule level

I would say that most beginner don't know this syntax:

```ts
[
	...!isTodayMonday ? [missingKeyrule()] : [],
]
```

They will probably don't be able to disable rules conditionally. Or at least struggle with it.

I agree `true` is weird, but we could leave the existing behavior and additionally add the `'default'` keyword.

### replace the concept of context with top-level arguments

`context` was meant to be extendable and will have a `fix` option in the future.

Passing it like you did is a great option. The `createLintRule` was one of the last additions. Therefore this option was not considered, because everything else was already ready and working.

One "downside" is, that a creator of lint rules is required to use the `createLintRule` function. But the function is a really good wrapper that handles some stuff for a developer and is is certainly recommended to use it.

### No end-to-end test existed

Where would you expect such a test?

`linter.ts` is the only place where it makes sense and those are e2e tests.

### Example 1

`createLintRule` is (was) not required and calling the function resulted in the same object.

Test was probably written before the `createLintRule` function was introduced.

### Example 2

Why is this test wrong? It just makes sure that the returned object of the callback you add to the `createLintRule` function get's passed to the resulting lint rule.


### Focus on code coverage instead of writing (duplicate) unit tests for every function like "for 'Message' test if 'leave' is not present" and "for 'Pattern' test if 'leave' is not present". The test is identical. Writing and maintaining both tests seems unnecessary. The underlying logic is identical.

If you would refactor the `leave` function inside `Pattern`, how would you know that you didn't break the existing behavior? If you don't test it, you won't know.

### should lint return errors

It makes sense to return errors instead of just logging them. If it is something exposed externally, a tuple makes sense.

### how should the config be passed to a lint rule?

So we now need again a second "setup" function?

Certainly option A.

### out comment lint collections for now

It is there and it already works. Sure we can just don't expose it for now and wait until someone demands it.

### don't allow rules to define the lint level?

Rules should be configurable! An error for Team A may be just a warning for Team B. Or while migrating something you may wish to just have a warning for a while. And once migration is done, you can change it to an error.

#### Users face a hidden lint level that might make them wonder why certain things are errors or warnings.

We could say that rules can't specify a default level and everything is a warning, but I would not do that. The `additionalKey` rule is a good hint that you should take a deeper look, but it should not fail CI, because it does no real harm if it is there.

#### explicit behavior for close to no additional "cost"

The less an end user needs to do the easier the adoption will be. 

# Conclusion

I would not remove any functionality. I would just make `setup` optional, pass those arguments somewhere else and add `"default"` to the options that a lint rule accepts.

source-code/core/src/lint/linter.test.ts
```diff
@@ -618,273 +561,4 @@ describe("lint", async () => {
 			})
 		})
 	})
```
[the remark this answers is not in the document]
samuelstroschein: leave and enter remains! :)

source-code/core/src/lint/rule.ts
```diff
@@ -101,27 +95,35 @@ export type LintRule = {
...
- * 			return { token: '456' }
- * 		},
- * 		visitors: {}
+ * 			visitors: {}
  * 	}
  * })
  * ```
  */
-export const createLintRule = <Settings = never, RequireSettings extends boolean = false>(
+export const createLintRule = <Settings = never>(
```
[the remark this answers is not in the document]
samuelstroschein: I'll fix this. This PR was not meant to be reviewed yet :D Your type skills are god-like.

source-code/core/src/lint/README.md
```diff
@@ -253,11 +222,11 @@ export async function defineConfig(env) {
 				}),
 				// uses the standard lint level and passes custom settings to the `missingKey` rule
 				inlangStandardRules({
-					missingKey: [true, { threshold: 4 }],
+					missingKey: ["default", { threshold: 4 }],
 				}),
 				// disables the `missingKey` rule if it runs on mondays
 				inlangStandardRules({
-					missingKey: !isTodayMonday,
+					missingKey: false,
```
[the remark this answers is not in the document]
ivanhofer: But you said, that you want to get rid of `true` and `false` for enabling/disabling rules

samuelstroschein:
Oh wow, I just stumbled on a great syntax for entering and leaving nodes that is easier to explain and future-proofs the api.

```ts
const myRule = createLintRule("foo.x", ({ settings, report }) => {
 return {
  visitors: {
   Resource: ({ onLeave }) => {
    // the state is now local to a visitor node _wihtout the concept of payloads_
    const isError = true;
    // on leave (and many other features) could be exposed via a callback instead of 
    // changing the signature of 
    onLeave(() => report({ node: this, message: "foo" }));
   },
  },
 };
});

```

ivanhofer:
> * Initializing an API is always sync

Not really. Take a look at this: https://auth0.com/docs/secure/tokens/access-tokens/get-management-api-access-tokens-for-production.
You need to retrieve an access token before you can make an authenticated API call.
Making the function (optional) async is not really a big deal, right?

> Oh wow, I just stumbled on a great syntax for entering and leaving nodes that is easier to explain and future-proofs the api.

This is a good idea if you want to share state on the same level. I would remove the `onLeave` parameter and define if a function call get's returned, it is the function that get's called when leaving the node.

Once minor thing that got lost by removing `payload` is the ability to pass information down to the children. A `Pattern` does not really know in which `Resource` it is present.

---

A lot of the things mentioned above were not clear. I'll wait until you are ready with the refactoring.

samuelstroschein:
> Making the function (optional) async is not really a big deal, right?

With the current way I approach this, no. I can make it async. Will do. 

> I would remove the onLeave parameter and define if a function call get's returned, it is the function that get's called when leaving the node. 

This would collide with the concept of returning "skip" and lead to a concept that needs to be explained in general. The callback pattern `onLeave` is explicit and can be extended with other callbacks over time without introducing new concepts.

> A lot of the things mentioned above were not clear. I'll wait until you are ready with the refactoring. 

Roger 👍

samuelstroschein:
NOTE TO MYSELF 

I think I found the design improvement that makes lint much simpler. The setup function should return the visitors, not be a separate property. This resembles ESLint and has one major advantage: 

The double setup function described in https://github.com/inlang/inlang/issues/457 is eliminated without forcing the usage of `createLintRule` and a double setup function pattern like `missingMessageRule("error")(setup)`. The double setup function pattern would require the config to execute all rules to retrieve static information like the `id` or `level`.  

```ts
// current design
export type LintRule = {
        id: LintRuleId,
        level: LintRuleLevel,
	setup: (args) => MaybePromise<unknown>
	visitors: NodeVisitors
}
```

```ts

export type LintRule = {
        id: LintRuleId,
        level: LintRuleLevel,
	setup: (args) => Promise<Visitors>
}
```

source-code/core/src/lint/README.md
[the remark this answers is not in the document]
samuelstroschein: Yes. For a rule(). How rule collections would enable or disable rules is up to them. But, I'll remove rule collections for now. We can wait for user demand.

samuelstroschein:
@ivanhofer Done ✅

- [x] expose fewer features to not run into the risk of needing to maintain low usage features for years. let's wait for user feedback before we expose more
- [x] heavily lean into object argument pattern to ease future API changes
- [x] return errors during a lint process for apps to display the errors in the UI
- [x] no more 2 setup functions. changing the lint rule as described in https://github.com/inlang/inlang/pull/459#issuecomment-1482681050 was the breakthrough
- [x] don't expose lint test utilities. wait for the merger of  https://github.com/inlang/inlang/pull/436 which introduces `@inlang/core/test`

source-code/core/src/lint/README.md
```diff
@@ -100,93 +86,52 @@ Example:
...
-       Those properties can be either a single function or an object with a `enter` and/or `leave` function:
 
      ```ts
      visitors: {
-       Message: () => { /* implementation */ },
-       Pattern: {
-         enter: () => { /* implementation */ },
-         leave: () => { /* implementation */ },
-       },
+       Resource: ({ referece, target }) => { /* implementation */ },
```
ivanhofer, line 125:
```suggestion
       Resource: ({ reference, target }) => { /* implementation */ },
```

source-code/core/src/lint/README.md
```diff
@@ -229,55 +174,100 @@ lint: {
...
 ```
 
-## Lint Collections
+## Design decisions
 
-Lint collections are a way to group rules together. You can pass a collection to the config like you would do with regular rules. You can also customize and disable specific rules of a collection by passing an argument to the function call. A collection accepts an object where the keys are the name of the provided rule and the values are the configuration for that rule.
+#### Iterative improvements.
 
-_inlang.config.js_
+The lint module is small on purpose to not expose features that might not be used by users but need to be maintained for backwards compatiblity. We will implement features as requested by users.
```
ivanhofer, line 181:
```suggesion
The lint module is small on purpose to not expose features that might not be used by users but need to be maintained for backwards compatibility. We will implement features as requested by users.
```

source-code/core/src/lint/README.md
```diff
@@ -229,55 +174,100 @@ lint: {
...
+
+// 2. configure rule function [Public API: Using rules]
+rules: [
+  missingMessageRule("error", {
+    strict: true,
+  })
+]
+
+// 3. setup rule function [Internal]
+const rules = rules.map((rule) => await rule.setup({ config, report }))
```
ivanhofer, line 269: there is no `setup` function anymore

source-code/core/src/lint/README.md
```diff
@@ -229,55 +174,100 @@ lint: {
...
+	grammarRule("error"),
+]
 ```
 
-The `createLintRuleCollection` expects an object where the key is the name of the provided rule and the value is the rule itself. You can add as many rules as you want.
+#### Flow of the lint module
+
+```ts
+// 1. create lint rule [Public API: Authoring rules]
+const missingMessageRule = createLintRule({ id: "inlang.missingMessage" }, ({ config, report }) => {
```
ivanhofer, line 253: We should make an example how to create a rule with typed settings

source-code/core/src/lint/report.ts
```diff
@@ -0,0 +1,19 @@
...
+export const createReportFunction = (_args: { id: LintRule["id"]; level: LintRule["level"] }) => {
+	return (args: { node: LintableNode; message: string }) => {
+		if (!args.node) return
+
+		args.node.lint = [
+			...((args.node as LintedNode).lint ?? []),
+			{
+				id: _args.id,
+				level: _args.level,
+				...args,
```
ivanhofer, line 15: I would not add the `node` to this object. Just the message.

source-code/core/src/lint/createLintRule.ts
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * // you can also pass settings to the rule
+ * const myRule2 = createLintRule<{ strict: boolean }>({ id: "example.rule" }, (args) => {
+ *   // Implement your rule logic here
+ * });
+ *
+ * // settings must be provided when using the rule
+ * myRule2("error", { strict: true });
+ */
+export const createLintRule = <Settings extends object | undefined = undefined>(
```
ivanhofer, line 22: no more required and optional Settings?

source-code/core/src/lint/createLintRule.ts
```diff
@@ -0,0 +1,56 @@
...
+	args: { id: LintRule["id"] },
+	setup: (
+		args: Parameters<LintRule["setup"]>[0] & {
```
ivanhofer, line 25: why isn't `Settings` part of `LintRule["setup"]`?

source-code/core/src/lint/linter.ts
```diff
@@ -1,235 +1,166 @@
...
+	target?: Node
+}) => Promise<void>
 
 // --------------------------------------------------------------------------------------------------------------------
 
-const shouldProcessResourceChildren = (visitors: NodeVisitors) =>
+const shouldProcessResourceChildren = (visitors: Visitors) =>
 	!!visitors.Message || shouldProcessMessageChildren(visitors)
 
 // TODO: test passing `undefined` for reference
```
ivanhofer, line 87: should we remove this TODO? ^^

source-code/core/src/lint/createLintRule.ts
```diff
@@ -0,0 +1,56 @@
...
+ * const myRule2 = createLintRule<{ strict: boolean }>({ id: "example.rule" }, (args) => {
+ *   // Implement your rule logic here
+ * });
+ *
+ * // settings must be provided when using the rule
+ * myRule2("error", { strict: true });
+ */
+export const createLintRule = <Settings extends object | undefined = undefined>(
+	args: { id: LintRule["id"] },
+	setup: (
```
ivanhofer, line 24: should `setup` be integrated into `args`?

source-code/core/src/lint/createLintRule.ts
```diff
@@ -0,0 +1,56 @@
...
+		}
+	}
+
+	return configureLintRule
+}
+
+/**
+ * Type for the function that configures a lint rule.
+ */
+type ConfigureLintRuleFunction<Settings extends object | undefined = undefined> =
```
ivanhofer, line 51: `object` should be `Record<string, unknown>`

source-code/core/tsconfig.json
```diff
@@ -5,7 +5,8 @@
 		// DOM because of fetch. TODO: extract fetch from this package and
 		// let fetch be provided by the environment.
 		"lib": ["ESNext", "DOM"],
-		"types": [],
+		// need to specify node because `$fs` has a node type dependency
+		"types": ["node"],
```
ivanhofer, line 9: what was the issue that you needed to add this?

source-code/core/src/lint/linter.ts
```diff
@@ -1,235 +1,166 @@
...
 }) => {
-	const { level, id, setup, visitors, teardown } = lintRule
-	if (!level) return
-
-	const context = createContext(id, level)
+	const { referenceLanguage, languages, reference, resources } = args
 
-	const payload = await setup({ referenceLanguage, languages, context })
+	const report = createReportFunction(args.rule)
+	const { visitors } = await args.rule.setup({ config: { referenceLanguage, languages }, report })
```
ivanhofer, line 63: I would just pass the props needed for the `createReportFunction`
```suggestion
	const  { id, level, setup } = args.rule
	const report = createReportFunction({ id, level })
	const { visitors } = await setup({ config: { referenceLanguage, languages }, report })
```

source-code/core/src/lint/README.md
samuelstroschein: There is. I moved `visitors` into the setup function. https://github.com/inlang/inlang/pull/459#issuecomment-1482681050

source-code/core/src/lint/README.md
samuelstroschein: Done.

source-code/core/src/lint/linter.ts
samuelstroschein: I stopped destructering props because SolidJS looses reactivety on desctructering (proxy are not desctructed :/)

source-code/core/src/lint/report.ts
samuelstroschein: good catch. i haven't changed this code though ;)

source-code/core/tsconfig.json
samuelstroschein: $fs uses buffer to use node's filesystem. this "bug" has likely been introduced by https://github.com/inlang/inlang/issues/391. 

We need to build our own filesystem implementation with node independent types. I didn't want to change fs related things in this PR though.

FYI the build step fails otherwise

source-code/core/src/lint/createLintRule.ts
samuelstroschein: done

source-code/core/src/lint/createLintRule.ts
samuelstroschein: Still exists with easier types 😎

Checkout the type tests https://github.com/inlang/inlang/pull/459/files#diff-d3442843c87c4325671801e6262b8ce1e656aef697b1b87f6238968d1f026144. 

The type that provides optional types is straightforward

```
/**
 * Type for the function that configures a lint rule.
 */
type ConfigureLintRuleFunction<Settings extends Record<string, unknown> | undefined = undefined> =
	Settings extends undefined
		? // If settings are not defined, the function only takes a level
		  (level: LintRule["level"]) => LintRule
		: // Else, the function takes a level and settings
		  (level: LintRule["level"], settings: Settings) => LintRule
```

source-code/core/src/lint/createLintRule.ts
samuelstroschein: 1. The linter can't pass settings to rules.
2. Even if, the implementation would become more complex with no benefit. 

The job of `createLintRule`, which wraps the `setup` function of a rule, is to pass custom arguments like `settings` to a rule.   


```ts

rules: [
  // a "manually" written lint rule
  {
    id: "my.lintRule",
    level: "warn", 
    setup: ({ report }) => {
      // developers can directly access settings as desired + the linting implementation can't
      // pass down user-configured settings to a lint rule.   
      if (strict) {
        report()
      }
    }
  }
  // a lint rule that has been created with createLintRule
  missingMessge("error", { strict: true })
]
```

source-code/core/src/lint/createLintRule.ts
samuelstroschein: Yes. I'll change it!

----

While writing down the arguments I came to believe that you are right.

`args` = static "config". i didn't want to name it config though because setup takes the inlang config as argument.
`setup` = dynamic things that actually make up a rule. 

```ts
export type LintRule = {
	id: `${string}.${string}` // <-- static
	level: "error" | "warn"   // <-- static
	setup: Function           // <-- dynamic
}
```

### Problems with the current solution

1. Supossing that a lint rule will be extended with more dynamic "arguments" in the future, using `setup` directly will require us to introduce a third positional argument (we are not using the object argument pattern). 

2. `args` is not clear.

### Proposal 

YEAH NO. Abort the mission. 

```ts
export const createLintRule = <Settings extends Record<string, unknown> | undefined = undefined>(
	static: { id: LintRule["id"] },
	args: {
           setup: (
		args: Parameters<LintRule["setup"]>[0] & {
			settings: Settings
		},
	)} => ReturnType<LintRule["setup"]>,
): ConfigureLintRuleFunction<Settings>
```

samuelstroschein: Wuuups I know why I did this. We should probably go with `static` and the second argument being `callback`. 

If we put `setup` into `args`, a createLintRule can't have shared state anymore. We would end up with this issue again https://github.com/inlang/inlang/issues/457#issue-1631994310. 

```ts
const myRule = createLintRule({
id: "example.rule",
setup: () => {
	return {
		visitors: {
			Resource: () => {},
		},
	}
},
})
```

The current design is read as "the left argument is static config, the right one is a callback for dynamic properties"

```ts
const myRule = createLintRule({ id: "example.rule" }, async () => {

// shared global state
const x = "hello world"

// or initialize an API .
const API = await Grammarly()

return {
	visitors: {
		Resource: () => {},
	},
}
})
```

I tried to remove the `setup` entirely from `LintRule` but a lint rule must have an async callback property that the linter executes. Otherwise, execution must happen in `defineConfig()` which we want to avoid. 

Given that all dynamic "properties" of a lint rule must be returned by setup, I reverse my previous comment. Args and setup should be de-coupled. All dynamic properties will be behind `setup`, thus we are not running into the issue of requiring additional arguments for the `createLintRule` function in the future. 

BUT `args` is not optimal. `static` or `config` would be suitable. Unfortunately, `static` is a reserved keyword and `config` might be confused with the inlang config. 

@ivanhofer ideas for a better name?

samuelstroschein: I just realized, while laying in bed, that ofc we can have shared state if everything dynamic happens in the setup function. 

```diff
const myRule = createLintRule({
id: "example.rule",
setup: async () => {
+       const x = "here is shared state"
	return {
		visitors: {
			Resource: () => {},
		},
	}
},
})
```

For the sake of keeping it simple and future-proof, I now (again :D) vote for merging `args` and `config`.

source-code/core/src/lint/createLintRule.ts
ivanhofer: Not exactly true..
```ts
const rule = createLintRule<{ strict?: boolean }>()

// this shows an error
rule('error')
// you need to pass an empty object
rule('error', { })
```

source-code/core/src/lint/createLintRule.ts
ivanhofer: ok, got it

source-code/core/src/lint/createLintRule.ts
ivanhofer: I would merge them!

source-code/core/tsconfig.json
ivanhofer: You could also import it from "node:fs" and add the `eslint-ignore` on top. Then it is more clear that something needs to change. With global types it is hard to see where we use node types.

source-code/core/src/lint/createLintRule.ts
samuelstroschein: Done

source-code/core/tsconfig.json
samuelstroschein: Good hint. node does not export buffer though. I hope I fixed the issue with:

```ts
// eslint-disable-next-line no-restricted-imports
import type fs from "node:fs/promises"

/**
 * Minimal filesystem required by inlang to work
 *
 * TODO: node independent fs
 */
export type FS = {
	readFile: (
		...args: [id: string] | Parameters<fs.FileHandle["readFile"]>
	) => Promise<string> | ReturnType<fs.FileHandle["readFile"]>
	writeFile: (file: string, data: string) => Promise<void>
	readdir: (path: string) => Promise<Array<string | object>>
}
```

I expect type errors down the road but we need to build our own fs anyways.

source-code/core/src/lint/createLintRule.ts
samuelstroschein: I'll merge this for now. To unblock the editor team and @jannesblobel. We can add a boolean flag later. I vote against it though. Makes the lint usage more complicated and passing an empty object for defined settings with all optional properties is minor IMHO.

PR #460, opened 2023-03-21
add sdk-js RFC

ivanhofer:
[rendered](https://github.com/inlang/inlang/blob/sdk-js-structure/rfcs/sdk-js/RFC.md)

comments go [here](https://github.com/inlang/inlang/pull/460/files#diff-e75763f05e805b52ef115994027187bf14fcc31bd6e5dfc473edd9ccd34eb34c)

### DIFF
--- a/rfcs/sdk-js/RFC.md
+++ b/rfcs/sdk-js/RFC.md
@@ -39,7 +39,7 @@ The RFC is the first draft on how this could look like.
     - routes could collide; Some words exist in different languages and can have different meanings
       - e.g. the link `/bank` could lead to the english and german version. If the language is not encoded in the url, how do we render the correct language?
     - some routes may not exist in certain languages
-    - layouts could differ between languages; a simple if/else should be good enough to support that use case
+    - layouts could differ between languages (a simple if/else should be good enough to support that use case)
 
 ## How this SDK should function
 
@@ -47,7 +47,7 @@ The RFC is the first draft on how this could look like.
 
 The SDK should include some core functionality (called `runtime`) that can be shared across all frameworks.
 
-> By exposing the `runtime` functionality, other developers could use it to build their own i18n solution for any given framework that is compatible with `inlang`.
+> By exposing the `runtime` functionality, other developers could use it to build their own i18n solution that is compatible with `inlang` for any given framework.
 
 Runtime functionality will be:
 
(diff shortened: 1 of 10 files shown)


source-code/sdk-js/adapter-sveltekit/package.json
```diff
@@ -0,0 +1,4 @@
+{
+	"name": "@inlang/sdk-js-adapter-sveltekit",
```
ivanhofer, line 2: Is this the package-name we want developers to install the adapter?

ivanhofer: Each adapter should be it's own published npm package. Or else a Next.js project would need to install SvelteKit as a peer dependency.

`@inlang/sdk-js/sveltekit` is not a valid package name. A name can only contain a single `/`.

I just saw that we have a `git-sdk`. Should we then rename this package to `js-sdk` to be consitent?

samuelstroschein: > Or else a Next.js project would need to install SvelteKit as a peer dependency. 

We can unbundle the package in the future if developers start to complain about the size of the package.

Having everything in one package eases deployment, "SEO" because we only need to optimize one package on NPM/Readme, bundles all downloads together (higher chance of adoption / higher ranking on libhunt etc.).  

-------

Which of those SDKs would you be more inclined to adopt? 

```
@inlang/sdk-js             4500 weekly downloads
```

```
@inlang/sdk-js-sveltekit   1100 weekly downloads
@inlang/sdk-js-nextjs       900 weekly downloads
@inlang/sdk-js-astro        500 weekly downloads
@inlang/sdk-js-solid-start  600 weekly downloads
@inlang/sdk-js-react       1200 weekly downloads
```

> I just saw that we have a git-sdk. Should we then rename this package to js-sdk to be consistent? 

I think `sdk-{platform}` is better because the source code would be sorted. The git-sdk is "unrelated" to inlang. For ease of development, it lives in the monorepo for now.

```
website
sdk-js
sdk-flutter
sdk-ios
sdk-android
```

ivanhofer: Then let's do this for now. We can also change the structure later if we see that it makes sense

ivanhofer: One big downside that just came to my mind is:
Once we have multiple adapters, we will also have **functions with the same name multiple times**, because all adapters will share the same `API`. If you want to call `changeLanguage`, then your IDE will probably auto import it from `@inlang/sdk-js-svelte` instead of `@inlang/sdk-js-sveltekit`. If you don't pay attention then you will be wondering why your code does not work.

samuelstroschein: >One big downside that just came to my mind is:
Once we have multiple adapters, we will also have functions with the same name multiple times, because all adapters will share the same API. If you want to call changeLanguage, then your IDE will probably auto import it from @inlang/sdk-js-svelte instead of @inlang/sdk-js-sveltekit. If you don't pay attention then you will be wondering why your code does not work. 

That is a biggy. Can we simulate the exports and see how the IDE behaves?

ivanhofer: image

This is how it wlll look like. Sometimes the IDE shows the recently imported version at the top of the list and sometimes not..  There will be a lot of issues if we go with this approach.

Adding a custom path alias to `tsconfig.json` is also not a reliable option.
image
The "Quick fix..." panel will show it at the bottom, while "Code completion" shows it at the top
image

rfcs/sdk-js/RFC.md
```diff
@@ -0,0 +1,544 @@
...
+```ts
+const lookupFunction = (key: string) => ""
+
+export default lookupFunction
+
+export const i18n = lookupFunction
+export const t = lookupFunction
+export const l = lookupFunction
+export const $ = lookupFunction // will not work well together with Svelte
+```
```
ivanhofer, line 91: I would vote for `i18n` as the name for the lookup function. Everyone (that has heard of internationalization) knows the term and should immediately know what this line does compared to `t` or `l`. It is also relatively fast to write and takes not that much space in the markup compared to `translate`. Inlang in the future (once the tooling is built) could also support inline strings instead of Id based translations. 

```ts
const name = 'Inlang'
i18n('welcome', name) // Id based
i18n`Welcome to this application, ${name}!` // inline
``` 

Further additions could include a way to add a description
```ts
i18n('welcome', name)
	.describe('Welcome message that shows up at the top of the page after completing the registration.')
```

Since `t` is often used in Id based libraries, it may not be that obvious that Inlangs solution offers more capabilities.

rfcs/sdk-js/RFC.md
```diff
@@ -0,0 +1,544 @@
...
+
+There is the need to support multiple strategies. If the first one does not match, the next strategies will be tried until a match is found. If nothing matches, the `referenceLanguage` will be used.
+
+Setting up auto detection is optional. Developers can also just call the `changeLanguage` function directly.
+
+Each strategy will need a special framework agnostic implementation, to pass the necessary information to those detection functions.
+
+For the first version we probably won't need all of the strategies.
+
+> What detection strategies do we want to support?
```
ivanhofer, line 146: I would vote for `rootSlug`, `acceptLanguageHeader` and `navigator` since it is possible to cover SPAs, MPAs and content websites.

source-code/sdk-js/adapter-sveltekit/package.json
samuelstroschein: Wait. Do we even need to expose common functions like `changeLanguage` on a per-adapter level? 

If the adapters are only for the bundler and the SDK is "auto generated" based on the adapter, regardless of the framework everyone can import from `@inlang/sdk-js` or am i mistaken?

rfcs/sdk-js/RFC.md
samuelstroschein: I strongly vote for the lookup function.

1. Most teams will not use inlang to translate their product but enable non-technical users to make copy changes. Communicating "use the translate/i18n function" will confuse people "but, I don't want to translate anything". 
2. Lookup function is technically correct.

source-code/sdk-js/adapter-sveltekit/package.json
ivanhofer: Right, we could just not expose them publically and the `plugin` could replace the function calls with the correct one. Since the functionality of `changeLanguage` will differ depending on the configuration we will need to do that anyways.

rfcs/sdk-js/RFC.md
ivanhofer: so you would like to have `lookupFunction('welcome', name)` spread over the whole code base?

samuelstroschein: Ah, no ofc not. I thought the conversation is about the debating whether the "lookup function" should be named "i18n". My mistake!

I had a shorthand `t` as lookup function in mind. `t` is shorter than `i18n` and we should try to avoid i18n to ease our communication of inlang [as in, can be used without translating apps].

ivanhofer: `t` for `translate` is also not a good fir then ^^
Maybe `i` or `in` for `inlang`?

benjaminpreiss: Hmm, why not `l` for lookup? I prefer the term `lookup`over `translate`, since I would expect a function named "translate" to do a translation of some string I have to provide... `i` would also be a good fit IMO, though a little undescriptive? Also, how about the classic `get`? ^^

ivanhofer: the small `l` looks like a big `I`. That could be confusing. I would not use `get`. In my opinion this looks a bit weird in your template ^^

samuelstroschein:
@ivanhofer and @benjaminpreiss FYI feel free to merge if the review is complete

PR #463, opened 2023-03-24
lookup function basics

### LINKED ISSUE #467: runtime - lookup function
See https://github.com/inlang/inlang/blob/sdk-js-structure/rfcs/sdk-js/RFC.md#runtime

A function that holds a refernce to a `Resource` `AST`. The function call expects a `Key` and additional an optional object with `Placeholders`.

The `AST` needs to be enhanced with the `Placeholder` concept

ivanhofer:
Closes #467

Implements the core functionality from https://github.com/inlang/inlang/blob/sdk-js-structure/rfcs/sdk-js/RFC.md#runtime

 - adds the `Variable` node to the `AST`
 - allows to create a function to serialize a `Message` from the `AST` to string
    - already comes with full `TypeScript` support so we can easily enhance it in the future
    - allows to pass arguments which get's replaced
- provides utility function `initRuntime` to be able to get the base functionality for a runtime
   - loading resources into memory (actual implementation follows in another PR)
   - switch language functionality
   - creating a lookup function which uses the `Resource` loaded before
      - if `Resource` is not present a fallback lookup function get's provided
    - `initRuntime` uses the internal `initBaseRuntime` function
       - so it is easier to test
       - because we will need an "advanced" runtime with more capabilities in the future


source-code/sdk-js/src/runtime/runtime.ts
```diff
@@ -0,0 +1,53 @@
...
+	},
+) => {
+	const loadResource = async (language: Language) => {
+		const resource = dummyResource // TODO: actually load resource
+		state.resources.set(language, resource)
+	}
+
+	const switchLanguage = (language: Language) => (state.language = language)
+
+	const getLanguage = () => state.language
```
ivanhofer, line 38: what should we do if the language was never set? Return `undefined`? Throw an error?

source-code/core/src/ast/schema.ts
```diff
@@ -116,42 +138,5 @@ export type LanguageTag<Extension extends ExtensionInformation = ExtensionInform
 	 * @see https://www.ietf.org/rfc/bcp/bcp47.txt
 	 * @see https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry
 	 */
-	// language: string;
+	// language: string
 }
-
-// export type MessageComment = Node<"MessageComment"> & {
```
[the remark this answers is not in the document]
samuelstroschein: Those types are from the Fluent and upcoming Message Format 2.0 AST

source-code/core/src/ast/schema.ts
```diff
@@ -93,6 +102,19 @@ export type Text<Extension extends ExtensionInformation = ExtensionInformation>
...
+export type Expression<Extension extends ExtensionInformation = ExtensionInformation> =
+	Variable<Extension>
+
+export type Variable<Extension extends ExtensionInformation = ExtensionInformation> = Node<
+	"Variable",
+	Extension
+> & {
+	id: Identifier
+}
+
```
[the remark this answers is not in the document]
samuelstroschein: Term-wise, we should stick closely to the new MessageFormat 2.0 grammar [1] which is on track to become the standard for JS.

source-code/sdk-js/src/runtime/lookup-function.test-d.ts
```diff
@@ -0,0 +1,32 @@
...
+const resource = {} as Resource
+
+// ------------------------------------------------------------------------------------------------
+
+{
+	const i = createLookupFunction(resource)
+
+	i("hello")
+	i("welcome")
+	i("welcome", { name: "Inlang" })
```
samuelstroschein, line 13: Love it. Why didn't we think about `i` for inlang earlier? :D

samuelstroschein:
FYI would also be great if you keep the object argument pattern in mind

> To ease the iterative improvement approach, most public APIs use the object argument pattern.

```ts
// Example: Deprecating the "level" argument.

// before refactoring
function withObjectPattern(args: { id: string; level: string; type: string })
function withPositionalArguments(id: string, level: string, type: string)

withObjectPattern({ id: "hello", level: "world", type: "JavaScript" })
withPositionalArguments("hello", "world", "JavaScript")

// after refactoring
function withObjectPattern(args: { id: string; type: string })
function withPositionalArguments(id: string, level?: string, type: string)

withObjectPattern({ id: "hello", type: "JavaScript" })
withPositionalArguments("hello", undefined, "JavaScript")
```

source-code/core/src/ast/schema.ts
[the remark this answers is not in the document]
ivanhofer: > those types don't use the extension

Yes, but also some of them don't match the spec and for now we don't need them. We can introduce the correct version once we need them.

source-code/core/src/ast/schema.ts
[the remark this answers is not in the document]
ivanhofer: Ok, this is something that was not 100% clear to me. So everthing needs to be it's own node. Roger

ivanhofer:
> FYI would also be great if you keep the object argument pattern in mind

Where do you see the need for this?

samuelstroschein:
>Where do you see the need for this? 

I don't. I was posting this for you to keep it in mind :)

ivanhofer (commented):
@samuelstroschein I have changed the AST. Please take another look.

samuelstroschein (approved):

source-code/sdk-js/src/runtime/runtime.ts
```diff
@@ -0,0 +1,53 @@
...
+	LookupFunctionArgs extends BaseLookupFunctionArgs = BaseLookupFunctionArgs,
+>() => initBaseRuntime<Language, LookupFunctionArgs>()
+
+export const initBaseRuntime = <
+	Language extends string,
+	LookupFunctionArgs extends BaseLookupFunctionArgs = BaseLookupFunctionArgs,
+>(
+	state: RuntimeState<Language> = {
+		resources: new Map(),
+		language: undefined as unknown as Language,
```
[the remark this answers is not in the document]
ivanhofer: @samuelstroschein @benjaminpreiss what do you thing here?

source-code/sdk-js/src/runtime/runtime.ts
ivanhofer: @samuelstroschein @benjaminpreiss what do you thing here?

source-code/sdk-js/src/runtime/runtime.ts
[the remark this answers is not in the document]
samuelstroschein: If the default value is `undefined` now, we can always change the value later on.

I assume that real-world usage will show what a good default value is. If we try to set a value now and realize that `undefined` is better, changing this would likely be a breaking change.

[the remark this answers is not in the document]
ivanhofer: But what should the type definition of `getLanguage` be? `string | undefined`?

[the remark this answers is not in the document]
samuelstroschein: `string` there will always be a fallback, correct?

source-code/sdk-js/src/runtime/runtime.ts
samuelstroschein: i assume a language always exist, worst case the reference language, right?

source-code/sdk-js/src/runtime/runtime.ts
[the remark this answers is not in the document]
ivanhofer: Not in the core library. If the language was never set, then we don't have a value. We should probably use `steing | undefined`

source-code/sdk-js/src/runtime/runtime.ts
ivanhofer: I think it is best to return `string | undefined` here.

ivanhofer:
Waiting for the review of @benjaminpreiss. We can merg it afterwards.

source-code/sdk-js/src/runtime/runtime.ts
samuelstroschein: Yes. We can always narrow the type in a later iteration.

source-code/sdk-js/src/runtime/runtime.ts
benjaminpreiss, line 1: I can see that we are not using a reactive system here, but just the reference to a state... How will we make sure that the state will not be garbage collected at some point?

source-code/sdk-js/src/runtime/runtime.test.ts
```diff
@@ -0,0 +1,129 @@
...
+		})
+	})
+
+	describe("getLookupFunction", () => {
+		test("it should not throw if language was never set", () => {
+			const runtime = initBaseRuntime()
+
+			const i = runtime.getLookupFunction()
+
+			expect(i("test")).toBe("")
```
benjaminpreiss, line 74: I am not sure if I understand why this shouldn't throw (or at least how this should be an empty string)

source-code/sdk-js/src/runtime/runtime.ts
ivanhofer: When we hold a reference to this object, nothing get's garbage collected.

source-code/sdk-js/src/runtime/runtime.test.ts
ivanhofer: You probably won't crash your whole application if you reference a message that is not present. Rendering an empty string is the safest way to not affect anything else on screen.

benjaminpreiss (approved):
Great job!

PR #473, opened 2023-03-28
add loadResource functionality to runtime

ivanhofer:


source-code/sdk-js/src/config/index.ts
```diff
@@ -0,0 +1,47 @@
+import {
+	Config as InlangConfig,
+	EnvironmentFunctions,
+	initialize$import,
+} from "@inlang/core/config"
+
+// this is a copy from `source-code/website/src/pages/editor/@host/@owner/@repository/State.tsx`
+// TODO: move functionality into `core/config`
```
ivanhofer, line 8: We will need this functionality a few times. @samuelstroschein should we create a reusable function?

source-code/sdk-js/src/config/index.ts
```diff
@@ -0,0 +1,47 @@
...
+			"data:application/javascript;base64," + Buffer.from(file).toString("base64")
+
+		const module = await import(/* @vite-ignore */ withMimeType)
+		const config: InlangConfig = await module.defineConfig({
+			...environmentFunctions,
+		})
+
+		return config
+	} catch (error) {
+		// TODO: don't throw errors
```
ivanhofer, line 33: how do we wan't to deal with exceptions? Throwing errors is bad! The developer who uses this function needs to know implementation details or else he will not know that something can go wrong here.
We should return exceptions instead. 
https://github.com/ivanhofer/exceptionally or `[data, exception]` tuple?

source-code/sdk-js/src/runtime/runtime.test.ts
```diff
@@ -1,9 +1,45 @@
+import type { Message, Resource } from "@inlang/core/ast"
 import { describe, expect, test } from "vitest"
-import { initBaseRuntime, initRuntime, RuntimeState } from "./runtime.js"
+import { initBaseRuntime, initRuntime, RuntimeContext, RuntimeState } from "./runtime.js"
+
+// this is a copy from `source-code/core/src/lint/linter.test.ts`
+// TODO: expose utility functions somewhere
```
ivanhofer, line 6: I already have exposed those functions somehwere, but @samuelstroschein reverted those changes. I would love to have them defined somewhere so we can use them to write tests faster.

source-code/website/src/pages/editor/@host/@owner/@repository/State.tsx
```diff
@@ -561,21 +561,13 @@ async function readInlangConfig(args: {
 			$fs: args.fs.promises,
 		}
 		const file = await args.fs.promises.readFile("./inlang.config.js", "utf-8")
-		const withMimeType = "data:application/javascript;base64," + btoa(file.toString())
+		const withMimeType =
+			"data:application/javascript;base64," + Buffer.from(file).toString("base64")
 
 		const module = await import(/* @vite-ignore */ withMimeType)
-		// account for breaking change from renaming the config
-		// https://github.com/inlang/inlang/issues/291
```
ivanhofer, line 568: issue was closed. So we can remove this legacy code

ivanhofer (commented):
This PR fixes also some typos and removes legacy code

source-code/sdk-js/src/runtime/runtime.test.ts
samuelstroschein: Yes, coming in `@inlang/core/tests`. That is part of https://github.com/inlang/inlang/pull/472.

source-code/sdk-js/src/config/index.ts
samuelstroschein: Likely not. If you are referring to `readInlangConfig`, then you shouldn't use the function from the browser. 

The browser implementation uses a virtual filesystem which complicates things. You want to use `node:fs/promises` for the SDK. After all, the resources should be read from the local repository. Furthermore, you can directly import the inlang.config.js file in a node runtime. There is no need to import the module with mimeType. 

Taken from the [ide extension](https://github.com/inlang/inlang/blob/main/source-code/ide-extension/src/main.ts#L63-L67):

```ts

const $import = initialize$import({ fs, fetch })
// you just need to know the path of the config. the config must be at the root of a repo for now.
const module = await import("./inlang.config.js")
const config = await module.defineConfig({ $fs: fs, $import })

```

source-code/sdk-js/src/config/index.ts
samuelstroschein: SolidJS requires thrown Errors for ErrorBoundaries. 

As stated in https://github.com/inlang/inlang/pull/473/files/f1068c12d11fc3175a5e171fce9af702c598564e#r1150146838, you don't need this function.

source-code/sdk-js/src/runtime/runtime.test.ts
ivanhofer: When is this ready?

source-code/sdk-js/src/config/index.ts
ivanhofer: but I still need an error handling. Reading from fs can also go wrong

source-code/sdk-js/src/runtime/runtime.test.ts
samuelstroschein: End of this week. Just copy & paste the function for now

source-code/website/src/pages/editor/@host/@owner/@repository/State.tsx
samuelstroschein: Yes

source-code/sdk-js/src/config/index.ts
samuelstroschein: Not if you don't have a `readInlangConfig` function^^. If you decide to use one, just return whatever you'd like. The decision to use a result here or not is not exposed to the outside world.

ivanhofer: I know, but internally we should also use the same thing everywhere and not have 5 different solutions

samuelstroschein: > I know, but internally we should also use the same thing everywhere and not have 5 different solutions 

I agree. If we want the same type everywhere, I would either use the existing `Result` type (but it's too complicated) or use a simple union type with no extra bells and whistles

ivanhofer: I will just return `undefined` for now unti lwe have an official solution

source-code/sdk-js/src/config/index.ts
```diff
@@ -0,0 +1,42 @@
+import { DefineConfig, initialize$import } from "@inlang/core/config"
+import fs from "node:fs/promises"
+import { resolve } from "node:path"
+
+// eslint-disable-next-line @typescript-eslint/no-empty-function
+const noop = () => undefined
```
samuelstroschein, line 6: what is noop for?

source-code/website/src/pages/editor/@host/@owner/@repository/State.tsx
```diff
@@ -561,21 +561,13 @@ async function readInlangConfig(args: {
 			$fs: args.fs.promises,
 		}
 		const file = await args.fs.promises.readFile("./inlang.config.js", "utf-8")
-		const withMimeType = "data:application/javascript;base64," + btoa(file.toString())
+		const withMimeType =
+			"data:application/javascript;base64," + Buffer.from(file).toString("base64")
```
samuelstroschein, line 565: `btoa` is available in every JS runtime. `Buffer` is a node dependency. The preview link works though https://inlang-website-pr-473.onrender.com/editor/github.com/inlang/example. 

Thus, can stay as is. Implementing your ESLint no-node dependencies would likely be a win.

samuelstroschein (approved):
LGTM.

source-code/sdk-js/src/config/index.ts
ivanhofer: I will inline it. Previously I have used it in more places.

source-code/website/src/pages/editor/@host/@owner/@repository/State.tsx
ivanhofer: Ah right, this issue again ^^
I was confused by the deprecated flag and did want to fix it.

The problem here is that if any of our dependencies imports node types globally, we will have them in our project.
https://stackoverflow.com/a/67431122/9832456

We can't really get rid of the types unless we remove that dependency or that dependency releases an update which fixes the issue..

I have reverted those changes

PR #474, opened 2023-03-28
feat: add a feedback link to nav

### LINKED ISSUE #280: add feedback button to site
## Problem 

No explicit "Give feedback" button exists anywhere in the UI. 

## Proposal 

Add a button to the nav bar / potentially placed that opens https://github.com/inlang/inlang/discussions/categories/feedback.

boian-ivanov:
I'm picking some of the low hanging fruit. Feel free to share if you want me to modify this in any capacity.

New nav with discussions link added

Closes #280


source-code/website/src/pages/Layout.tsx
```diff
@@ -80,6 +80,12 @@ function Header(props: { landingpage?: boolean }) {
 	const links = [
 		{ name: "Blog", href: "/blog", type: "text" as buttonType },
 		{ name: "Docs", href: "/documentation", type: "text" as buttonType },
+		{
+			name: "Share Feedback",
+			external: true,
+			href: "https://github.com/inlang/inlang/discussions/categories/feedback",
+			type: "textPrimary" as buttonType,
```
samuelstroschein, line 87: Wow, thank you!

why did you choose "primary" over "text" like the other buttons?

NilsJacobsen: Hi @boian-ivanov, thank you, that's great. 
I would go for the `text`. So it's consistent with the other links.
I think it's really important to have this button in the Header, but I think on the marketing page it looks a bit cluttered. 
Maybe we could fix it by adding this condition:
`external: currentPageContext.urlParsed.pathname.includes("editor"),` in the object. 
This way we have the nice affordance in the editor but no cluttered header on the marketing page.

NilsJacobsen (changes requested):
Thank you @boian-ivanov the button makes absolutely sense. I left a comment. Would be great to hear your opinion or commit the small changes directly.

boian-ivanov:
@NilsJacobsen I tend to agree that it's cluttered with the text being too large, so I think shortening it to just "Feedback" should be fine. My only issue with removing the chevron (even for some of the pages) is that it wouldn't show any indication that it's an external link. Ideally instead of a chevron, a symbol for an external link would be used (like the one bellow). 



But still, I'm quite open to any ideas and if you guys feel that it won't need one, I wouldn't mind making the change.

source-code/website/src/pages/Layout.tsx
boian-ivanov: @NilsJacobsen I tend to agree that it's cluttered with the text being too large, so I think shortening it to just "Feedback" should be fine. My only issue with removing the chevron (even for some of the pages) is that it wouldn't show any indication that it's an external link. Ideally instead of a chevron, a symbol for an external link would be used (like the one bellow). 



But still, I'm quite open to any ideas and if you guys feel that it won't need one, I wouldn't mind making the change.

samuelstroschein:
@boian-ivanov thanks for the PR!

jannesblobel:
It's confusing with the feedback button and the arrow pointing to my profile. Can we add the feedback button in the Footer instead of the Header? 
CleanShot 2023-04-02 at 13 49 00@2x

PR #478, opened 2023-03-29
build config utility function for plugins 

samuelstroschein:
Closes https://github.com/inlang/inlang/issues/430. 

I needed this for the standard lint rules repository. 

- [x] validation of the config with extensive error messages

#### 1. Only export a build config instead of an entire build pipeline

- Our goal of defining a build step for the user that is compatible with inlang and controlled by us is achieved.
- Plugin authors have different requirements/preferences. Defining the entire build chain will likely lead to frustrations "why doesn't this work" and problems on our side because we need to support edge cases. 
- (Defining an entire build pipeline would take much longer).

#### 2. Use esbuild instead of tsup

- tsup can't bundle/has a "wrong" definition of bundling https://github.com/egoist/tsup/issues/619
- developers can use tsup regardless because tsup expects an esbuild config (what we are exporting with this utility function)
- tsup is another abstraction layer that we would eventually need to fight (see the first point)
- code-splitting (a benefit of rollup) is not useful for plugins because they are bundled into one file anyways (for now)

#### 3. Don't care about typescript declaration FOR NOW

- We need to implement TypeScript HTTP type imports (or find another solution; I strongly believe TS type imports over HTTP is the correct one) as described https://github.com/inlang/inlang/discussions/431#discussioncomment-5221303 first. 
- Since we have control over the build config now, we should be able to add type declarations later.

### DIFF
--- a/documentation/plugins.md
+++ b/documentation/plugins.md
@@ -22,10 +22,9 @@ Plugins can be imported via the `$import` [environment function](/documentation/
 // inlang.config.js
 
 export async function defineConfig(env) {
-  const plugin = await env.$import
-
-    "https://cdn.jsdelivr.net/gh/samuelstroschein/inlang-plugin-json/dist/index.js"
-  );
+	const plugin = await env.$import(
+		"https://cdn.jsdelivr.net/gh/samuelstroschein/inlang-plugin-json/dist/index.js",
+	)
 }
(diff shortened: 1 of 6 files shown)


ivanhofer (approved):
good solution!

PR #481, opened 2023-03-29
fix ci-cd-tests

samuelstroschein:
We are running into https://github.com/vitest-dev/vitest/issues/2008.


samuelstroschein:
Updated vitest, hoping that it fixes this issue

samuelstroschein:
Works!

PR #487, opened 2023-03-30
fix-require-not-defined

### LINKED ISSUE #486: ReferenceError: require is not defined
Sentry Issue: [INLANG-WEBSITE-CLIENT-21](https://inlang-com.sentry.io/issues/4046434389/?referrer=github_integration)

```
ReferenceError: require is not defined
  at None (/assets/src/pages/editor/@host/@owner/@repository/index.page.9e43f869.js:1:3952)
```

samuelstroschein:
Closes #486


samuelstroschein:
Man this bug only gets wilder. 

This fix turned `pluginBuildConfig` into an async function, subsequently making eslint complain about non-externalized modules that are imported in the scope of the function.

PR #494, opened 2023-04-02
replace `@rollup/plugin-typescript` with `@rollup/plugin-sucrase`

ivanhofer:
replace `@rollup/plugin-typescript` with `@rollup/plugin-sucrase` to improve build time and remove warnings

fyi @openscript

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2530,28 +2530,23 @@
 				}
 			}
 		},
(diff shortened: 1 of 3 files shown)


ivanhofer:
I got a lot of warnings when building:

image

Build time was reduced from ~1.2s to ~0.6s

PR #506, opened 2023-04-03
add-language functionality

ivanhofer:


NilsJacobsen:
Thank you @ivanhofer, gonna pick it up and add the add string modal.

source-code/website/src/pages/editor/@host/@owner/@repository/State.tsx
```diff
@@ -455,8 +478,7 @@ export function EditorStateProvider(props: { children: JSXElement }) {
 		if (!inlangConfig.error && inlangConfig() && fsChange()) {
 			// setting the origin store because this should not trigger
```
ivanhofer, line 479: We have a blocker.
We set the resources with `setOriginResources` in line 455, then write everything to disk, with `writeResources` which at the end updates `fsChange`. This triggers this code and calls `readResources`. But read resources will only read the old files from the file system. Then newly written resource does not get loaded.

ivanhofer: fixed it!

ivanhofer:
This PR is ready for review!

NilsJacobsen:
@ivanhofer i merged and tested it locally.
But it doesn't pass the tests in github actions.

NilsJacobsen:
@ivanhofer did you already hear about that octokit/core error?

ivanhofer:
@NilsJacobsen I have pushed some commits. Please check if it works

NilsJacobsen:
There is only a lint error left. If you don't fix it, i'm gonna do it tomorrow:)

PR #540, opened 2023-04-13
memoryFs implementation

araknast:
Here is a draft of a basic memoryFs implementation, nothing is set in stone, so would love to year your feedback on this.

At the current moment the filesystem is implemented using `Map` as opposed to a standard object, which should allow for faster file access operations at the expense of performance when serializing to and de-serializing from JSON.

Each directory including the root exists as a `Map()` which maps a string filepath to a string containing file data or another `Map()` representing a directory. Each directory is created with entries for '.', '..', and '' which are circular references to itself and its parent. This allows for straightforward handling of relative paths and multiple slashes, but may make it difficult to directly serialize the filesystem object if necessary (the current `toJson` method filters circular references).

Finally I've made some changes to the schema to better reflect the actual implementation. Namely all public  methods are async and  any `read*` method will return `null` if the path cannot be found. Also I've removed the definition of `fromJson` from the schema since it is currently implemented as a static method. Would like to hear how you all think we should handle this.


source-code-git/fs/src/schema.ts
```diff
@@ -1,7 +1,7 @@
 export type Filesystem = {
 	writeFile: (path: string, content: string) => Promise<void>
-	readFile: (path: string) => Promise<string>
-	readdir: (path: string) => Promise<string[] | undefined>
+	readFile: (path: string) => Promise<string | null>
+	readdir: (path: string) => Promise<string[] | null>
```
samuelstroschein, line 4: Is there a reason to return `null` instead of `undefined`? 

TypeScript recommends to never use `null` https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined. Douglas Crockford also recommends to never use `null` https://www.youtube.com/watch?v=PSGEjv3Tqo0&t=561s.

source-code-git/fs/src/schema.ts
```diff
@@ -25,6 +25,5 @@ export type Filesystem = {
 		 *   fs.toJson({ exclude: ["**\/node_modules"] })
 		 */
 		exclude?: string[]
-	}) => Record<string, any>
-	fromJson: (json: string) => Filesystem
+	}) => Promise<Record<string, any>>
```
samuelstroschein, line 28: `fromJson` should return an instance of the filesystem, not a Record. 

Here is a real world example https://github.com/inlang/inlang/blob/main/source-code/rpc/src/functions/generateConfigFile.ts#L60

source-code-git/fs/test/__snapshots__/suite.test.ts.snap
```diff
@@ -0,0 +1,16 @@
...
+  "home": {
+    "user1": {
+      "documents": {
+        "file1": "text in the first file",
+      },
+      "downloads": {},
+      "file3": "text in the third file",
+    },
+  },
+}
```
samuelstroschein, line 15: Can you add vitest snapshots to .gitignore?

source-code-git/fs/src/memory/implementation.ts
```diff
@@ -0,0 +1,121 @@
+import type { Filesystem } from "../schema.js"
+
+type FileData = string
+type Directory = Map<string, Inode>
+type Inode = FileData | Directory
+
+export class memoryFs implements Filesystem {
```
samuelstroschein, line 7: Classes should be CamelCase. Please rename `memoryFs` to `MemoryFs` :)

samuelstroschein (changes requested):

source-code-git/fs/src/schema.ts
araknast: I find null` and `undefined` to be useful to distinguish between things that have been intentionally defined as having no value, vs things that have not been assigned a value but may be assigned one in the future. Similar to the difference between `404 Not Found` and `410 Gone`. 

That way when I see a `null` value somewhere I can tell that its something that's not supposed to be defined vs something that may have been left undefined due to a runtime error.

Ultimately its more of stylistic choice, and matters less when using typescript, but that's my reasoning at least.

source-code-git/fs/src/schema.ts
araknast: Woops, that was a mistake, probably made when I was shuffling type declarations around :P

source-code-git/fs/src/memory/implementation.ts
araknast: Gotcha, I was unsure since `index.ts` had `memoryFs` in lower camel case, that makes sense though.

source-code-git/fs/src/schema.ts
samuelstroschein: Gotcha. I vote to stick with `undefined` for the codebase and avoid JavaScripts two billion dollar mistake of having two nullable values.

samuelstroschein:
Just crossed my mind. The MemoryFS for the browser should be reactive and must support all isomorphic git functions that are called. 

We want to get rid of [`setFsChange` in the editor](https://github.com/inlang/inlang/blob/main/source-code/website/src/pages/editor/%40host/%40owner/%40repository/State.tsx#L86-L93). `setFsChange` is manually triggered on each `writeFile` call. Basically, a manual reactive system :D 

CleanShot 2023-04-13 at 21 47 02@2x

Two questions: 

1. Is implementing required filesystem APIs that isomorphic git requires trivial?
2. Can the reactive system of SolidJS be used in the memory fs that makes memory fs usable in the browser out of the box and node without any problems? Otherwise, the memory fs could be wrapped in a Proxy and made reactive.

source-code-git/fs/src/schema.ts
araknast: The diff is confusing, but that is the return value for `toJson`. I removed `fromJson` from the type declaration as I had implemented it as a static method on the class.

source-code-git/fs/test/__snapshots__/suite.test.ts.snap
araknast: done

araknast:
> Just crossed my mind. The MemoryFS for the browser should be reactive and must support all isomorphic git functions that are called.
> 
> We want to get rid of [`setFsChange` in the editor](https://github.com/inlang/inlang/blob/main/source-code/website/src/pages/editor/%40host/%40owner/%40repository/State.tsx#L86-L93). `setFsChange` is manually triggered on each `writeFile` call. Basically, a manual reactive system :D
> 
> CleanShot 2023-04-13 at 21 47 02@2x
> 
> Two questions:
> 
>     1. Is implementing required filesystem APIs that isomorphic git requires trivial?
> 
>     2. Can the reactive system of SolidJS be used in the memory fs that makes memory fs usable in the browser out of the box and node without any problems? Otherwise, the memory fs could be wrapped in a Proxy and made reactive.

Should be fairly trivial to implement the subset of the api required for isomorphic-git. As for making the `fs` reactive I'm not sure since I'm not too familiar with SolidJS, but if it's possible it should be fairly simple to implement due to the size of the codebase.

samuelstroschein:
> As for making the fs reactive I'm not sure since I'm not too familiar with SolidJS, but if it's possible it should be fairly simple to implement due to the size of the codebase. 

Should be simple. Just wrap any fs implementation with a simple wrapper that tracks calls to writeFile, readFile, etc.

samuelstroschein:
I can't test the implementation because this PR originates from a fork. 

I can't switch to this branch to test whether the filesystem implementation is working/commit to this PR directly. I decided to create a new PR with a branch that originates from the inlang repo.

PR #550, opened 2023-04-13
chore: enable more typescript options

ivanhofer:
I was wondering why TypeScript didn't tell me that something could be undefined and I realized that we don't have enabled some TypeScript options. The most important one is probably `noUncheckedIndexedAccess`. When enabled it changes
```ts
names[0] // string
```
to
```ts
names[0] // string | undefined
```

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -23,7 +23,7 @@ jobs:
       - run: npm install --no-save
 
       - name: run format
-        run: npm run format
+        run: npm run format:fix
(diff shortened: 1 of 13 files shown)


samuelstroschein:
@ivanhofer can we turn `format:fix` into `format` and neglect auto-fixable formatter issues? 

Happens quite often that CI/CD test fail due to formatting which has no implication on a correctly functioning program. Building a github action that auto formats and makes a commit would take too much time

PR #555, opened 2023-04-16
Fix-filter-low-performance

samuelstroschein:
Investigation pending

### DIFF
--- a/source-code/website/src/pages/editor/@host/@owner/@repository/Layout.tsx
+++ b/source-code/website/src/pages/editor/@host/@owner/@repository/Layout.tsx
@@ -257,7 +257,7 @@ export const LanguageIcon = () => {
 }
 
 function StatusFilter() {
-	const { inlangConfig, filteredStatus, setFilteredStatus } = useEditorState()
+	const { inlangConfig, filteredLintRules, setFilteredLintRules } = useEditorState()
 
 	const lintRuleIds = () =>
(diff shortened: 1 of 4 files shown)


samuelstroschein:
The bug is triggered when the lint rules filters are set to an empty array (remove all filters). Subsequently, shoelace textareas are re-rendered with an expensive height calculation function. 

The lint rule filter is only called at once place, in `matchedLints`. Matched lints contains 6 nested loops :O 

CleanShot 2023-04-16 at 22 01 53@2x

samuelstroschein:
Hah. The `matchedLints` function gets called when `filteredLintRules().length === 0` which happens when all lint rules are removed. BINGO!

CleanShot 2023-04-16 at 22 07 03@2x

This `<Show>` triggers a re-render of every Pattern in the worst case. The patterns trigger the expensive shoelace function calls.  

CleanShot 2023-04-16 at 22 08 36@2x

samuelstroschein:
Fix landed https://github.com/inlang/inlang/pull/555/commits/6bd75557e3c2946e1770854efef62048a54a7588. 

Using <Show> removed DOM nodes which include Shoelace's textarea element. Usually, that's no issue but web components can't rely on Solids "fine grained reactivity". The workaround ist to use `class="hidden"` to leave the dom nodes rendered (once already rendered) when a node is filtered.

PR #559, opened 2023-04-17
inlang-git-fs

samuelstroschein:
Follow up PR from https://github.com/inlang/inlang/pull/540#issuecomment-1512070577

### DIFF
--- a/source-code-git/fs/src/schema.ts
+++ b/source-code-git/fs/src/schema.ts
@@ -1,0 +1,31 @@
+export interface Filesystem {
+	writeFile: (path: string, content: string) => Promise<void>
+	readFile: (path: string) => Promise<string | undefined>
+	readdir: (path: string) => Promise<string[] | undefined>
+	mkdir: (path: string) => Promise<void>
+	rm: (path: string) => Promise<void>
+	/**
+	 * Serializes the filesystem to a JSON string.
+	 *
+	 * @example
+	 *   const result = fs.toJson()
+	 *   >> { "file1.txt": "content", "file2.txt": "content" }
+	 */
+	toJson: (args: {
+		/**
+		 * Takes a list of glob patterns to include.
+		 *
+		 * @example
+		 *   fs.toJson({ include: ["**\/*.json", "**\/*.md"] })
+		 */
+		include?: string[]
+		/**
+		 * Takes a list of glob patterns to exclude.
+		 *
+		 * @example
+		 *   fs.toJson({ exclude: ["**\/node_modules"] })
+		 */
+		exclude?: string[]
+	}) => Promise<Record<string, any>>
+	fromJson: (json: Record<string, string>) => Promise<Filesystem>
+}
--- a/source-code-git/README.md
+++ b/source-code-git/README.md
@@ -1,0 +1,3 @@
+# @inlang-git/\*
+
+This directory contains the source code for the git modules that make up the "the next git" that power inlang.
--- a/source-code-git/fs/.gitignore
+++ b/source-code-git/fs/.gitignore
@@ -1,0 +1,2 @@
+# vitest snapshots
+test/__snapshots__/
--- a/source-code-git/fs/package.json
+++ b/source-code-git/fs/package.json
@@ -1,0 +1,23 @@
+{
+	"name": "@inlang-git/fs",
+	"type": "module",
+	"private": true,
+	"exports": {
+		".": "./dist/index.js"
+	},
+	"scripts": {
+		"---- DEV AND BUILD": "Are not required. Direct TypeScript import is possible because this module is only used internally.",
+		"---- TEST ----------------------------------------------------------": "",
+		"test": "tsc --noEmit && vitest run --passWithNoTests --coverage",
+		"test:watch": "vitest watch --passWithNoTests --coverage",
+		"---- LINT ----------------------------------------------------------": "",
+		"lint": "cd ../.. && npm run lint:base ./source-code/core",
+		"lint:fix": "cd ../.. && npm run lint:fix:base ./source-code/core",
+		"format": "cd ../.. && npm run format:base ./source-code/core",
+		"format:fix": "cd ../.. && npm run format:fix:base ./source-code/core",
+		"---- OTHER ---------------------------------------------------------": "",
+		"clean": "rm -r -f ./dist ./.turbo ./coverage ./node_modules"
+	},
+	"dependencies": {},
+	"peerDependencies": {}
+}
--- a/source-code-git/fs/src/index.ts
+++ b/source-code-git/fs/src/index.ts
@@ -1,0 +1,2 @@
+export type { Filesystem } from "./schema.js"
+export { createMemoryFs } from "./memory/index.js"
--- a/source-code-git/fs/src/memory/README.md
+++ b/source-code-git/fs/src/memory/README.md
@@ -1,0 +1,1 @@
+# @inlang-git/fs
--- a/source-code-git/fs/src/memory/implementation.ts
+++ b/source-code-git/fs/src/memory/implementation.ts
@@ -1,0 +1,121 @@
+import type { Filesystem } from "../schema.js"
+
+type FileData = string
+type Directory = Map<string, Inode>
+type Inode = FileData | Directory
+
+function dirToArray(dir: Directory, base: string): Array<Array<string>> {
+	const specialPaths = ["", ".", ".."]
+	let pathArray: Array<Array<string>> = []
+	for (let kv of dir) {
+		if (specialPaths.includes(kv[0])) continue
+		if (kv[1] instanceof Map) {
+			pathArray = pathArray.concat(dirToArray(kv[1], base + "/" + kv[0]))
+		} else if (typeof kv[1] === "string") {
+			pathArray.push([base + "/" + kv[0], kv[1]])
+		}
+	}
+	return pathArray
+}
+
+function initDir(parentDir: Directory) {
+	// A circular reference allows for simple handling of leading and
+	// trailing slashes, as well as "." paths
+	//	
+	// Since these only exist internaly, there shouldn't be issues
+	// with serialization
+	const dir = new Map()
+	dir.set("", dir)
+	dir.set(".", dir)
+	dir.set("..", parentDir)
+	return dir
+}
+
+function followPath(
+	target: Inode | undefined,
+	path: string, 
+	makeParent: boolean = false
+): Inode | undefined {
+	const pathList: string[] = path.split("/")
+	let parentDir: Directory
+
+
+	if (makeParent) {
+		for (let path of pathList) {
+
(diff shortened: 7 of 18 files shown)


samuelstroschein:
@araknast i went ahead and refactored a few things to speed up the deployment of fs. some commits have messages that explain the rationale.   

Still missing: 

- [ ] node fs promises wrapper to use node's filesystem. @araknast a node fs has no state. hence, only a wrapper/proxy called `nodeFs(fs)` instead of `createNodeFs`? There is nothing to create :D  for example:
https://github.com/inlang/inlang/blob/a8b42a01317a7c0c73e05d24ac7dbe220d74f551/source-code/cli/src/commands/config/init.ts#L36-L40

- [ ] replace memfs in the editor. requires drop in possibility for isomorphic git. for example:
https://github.com/inlang/inlang/blob/a8b42a01317a7c0c73e05d24ac7dbe220d74f551/source-code/website/src/pages/editor/@host/@owner/@repository/State.tsx#L604-L613

- [ ] nice to have: if `toJson` supports the filters. a large chunk of [generateConfigFile.config.ts](https://github.com/inlang/inlang/blob/f16c4d314bafef3685e54fedb4ef86c134a3642b/source-code/rpc/src/functions/generateConfigFile.client.ts) could be removed.
https://github.com/inlang/inlang/blob/a8b42a01317a7c0c73e05d24ac7dbe220d74f551/source-code-git/fs/src/schema.ts#L14-L30

araknast:
Thanks for the refactor, I had left the branch a bit of a mess with plans to clean it up later, in the future I'll try to push cleaner code directly instead.

Node wrapper and `toJson`filters should be fairly simple, replacing memfs in the editor is more difficult as it will most likely require replicating most or all of the isomorphic-git test suite. Will work on them in that order.

araknast:
Updated the filesystem api to match with node. There is a slight deviation from the node api since we cannot return buffer objects from memory fs, there is a note about it in the code.

I've also unified the tests since the node api and memory fs are nearly identical, this should also be a good test of their compatibility. Note the test is in `src` now, since both filesystems use the same test script I figured that would make the most sense.

I've also moved the git sdk to `souce-code-git`, and spit it into separate modules. 

I will work on migrating the isomorphic-git tests next,  then hopefully we will be able to fully remove memfs from the editor.

samuelstroschein:
## Minor things

- [ ] Rename `FsError` to `FilesystemError`

Given that the interface is called `Filesystem`, `FsError` should likely be named `FilesystemError` to increase consistency and thereby DX. https://github.com/inlang/inlang/blob/8a635a9f40c5b749d69352b7cbc033fb3a2e13d6/source-code-git/fs/src/FsError.ts#L1-L10

- i recommend getting accustomed to VSCode :P

There are several formatting issues. If you open the codebase with vscode, vscode will auto format everything and we have an aligned development environment.  

- tests across the monorepo where failing 

Not a big issue but annoying when I have to review the work and stumble on a clear "this doesn't work" moment. I recommend to run all commands on the root of the monorepo. Hence, `npm run dev` instead of `npm run dev -w @inlang-git/fs` etc. to immediately surface such issues. 

CleanShot 2023-04-20 at 21 40 46@2x


## Remove node:fs/promises compatibility [important]

Given https://github.com/inlang/inlang/pull/559#issuecomment-1514118280 and the incompatibility of node:fs/promises with the requirement to run in the browser, a node fs API wrapper is required. Given that a node fs API wrapper is required, defining the filesystem interface to contain nodefs specific arguments increases the surface API area manyfold, leading to degraded DX,  and doesn't work. 

Take `readFile` for example. `readFile` takes an encoding argument for the node API to distinguish between returning a Buffer or a string.

https://github.com/inlang/inlang/blob/8a635a9f40c5b749d69352b7cbc033fb3a2e13d6/source-code-git/fs/src/schema.ts#L5-L6

 Inlang operates on "utf-8" encoding only (for now). Hence, Buffers are not required/supported but even worse the following test will fail if the encoding is not passed as an argument (which is not required): 

https://github.com/inlang/inlang/blob/8a635a9f40c5b749d69352b7cbc033fb3a2e13d6/source-code-git/fs/src/implementation.test.ts#L33-L37

Developers will not pass the encoding argument 💣 🐛 . Even if they did, developers will pass `base64` and what not which won't work in memfs and thereby breaking the goal of having one universal `Filesystem` API.  

```diff
-expect(await fs.readFile(`${tempDir}/home/user1/documents/file1`))
   .toEqual("text in the first file")

```

I recommend to wrap node's fs as we do with vscode: 

https://github.com/inlang/inlang/blob/28668875195240c08af383c23c77225c3effa75b/source-code/ide-extension/src/utils/createFileSystemMapper.ts#L4-L35

Talking of VSCode, they seemed to have had similar requirements to inlang. They build their own Filesystem API https://code.visualstudio.com/api/references/vscode-api#FileSystemProvider. 

----

Further notes: 

- I anticipate that we have to return Uint8Array's eventually with readFile etc. We can already have a `readFile` interface that takes an encoding argument that must be set to `utf-8`, giving us the option to set Uint8Array as the default when the time comes. 


- I would plan the Filesystem interface to make use of the "argument object pattern", see https://github.com/inlang/inlang/pull/463#issuecomment-1483345858. The argument object pattern allows the extension of APIs with less breaking changes. 

```ts
// Example: Deprecating the "level" argument.

// before refactoring
function readFileObject(path: string, options: { encoding: "utf-8", fromRemote: boolean })
function readFilePositional(id: string, encoding: "utf-8", fromRemote: boolean)

readFileObject("./inlang.config.js", { encoding: "utf-8", fromRemote: true })
readFilePositional("./inlang.config.js", "utf-8", true)

// after refactoring
function readFileObject(path: string, options: { fromRemote: boolean })
function readFilePositional(id: string, encoding?: "utf-8", fromRemote: boolean)

readFileObject("./inlang.config.js", { fromRemote: true })
readFilePositional("./inlang.config.js", undefined, true) // <--- wupsy ending up with undefined in the middle of the arguments
```

araknast:
### Re: Minor things
> -  i recommend getting accustomed to VSCode :P
> There are several formatting issues. If you open the codebase with vscode, vscode will auto format everything and we have an aligned development environment.

Sorry about that, I thought `eslint` would be enough 😅.  I've applied `prettier` in the latest commit, but I will also setup vscode in case there are still issues.

> - tests across the monorepo where failing
> 
> Not a big issue but annoying when I have to review the work and stumble on a clear "this doesn't work" moment. I recommend to run all commands on the root of the monorepo. Hence, npm run dev instead of npm run dev -w @inlang-git/fs etc. to immediately surface such issues.

Ahh my mistake. Build was working for me, but I had originally manually symlinked `source-code-git` into `node_modules` which caused package resolution to behave differently. It turns out I had not set up `tsc to respect the folder hierarchy of the source dir. Seems like node was still able to resolve the path in my case where the top level directory was symlinked but not when the package was set up correctly :P

### Re: Remove node:fs/promises compatibility
> Inlang operates on "utf-8" encoding only (for now). Hence, Buffers are not required/supported but even worse the following test will fail if the encoding is not passed as an argument (which is not required)

The memoryFs implementation is set up such that "utf8" is set as the default for the encoding argument, and so no argument needs to be passed. I've added a test case demonstrating this This is a break from the node fs api however which does come with it's own difficulties.

The original reason we discussed for having node compatibility in `memoryFs` was to avoid having two wrappers, one "from node" and one "to node".  The DX is certainly worse because node's DX is worse :P There is a tradeoff here between better DX in using our own fs API, and reduction of complexity on our end in emulating the node fs api. I will create a node wrapper around the original api, but I think we should use it evaluate the costs and benefits of node compatibility more thoroughly.

araknast (commented):
Ignore [9aae865](https://github.com/inlang/inlang/pull/559/commits/9aae8652f625747ef8b3b9210143ac1762465d8a), see comment below

araknast:
The filesystem has been refactored to use two wrappers and the base `Filesystem` api has been reverted to the more basic original API, although now errors are thrown instead of functions returning undefined. Errors are more verbose and a new `FsError` type has been defined to be somewhat compatible between our `MemoryFilesystem` and node's `fs`

## Organizational Changes
All of this was done on a whim based on what made the most sense to me at the time, let me know if I should change anything back.
- Most notably everything is in the top level of the `src` folder now. I did this mainly because I didn't see a way to properly separate the files into `node` and `memory` categories with the imports still making sense.
- A new file `schema-internal.ts` has been created to distinguish between types that are used internally and those that should be exported

## The Refactor
- Three main functions now make up the filesystem api
    - `fromNodeFs` - takes an `fs/promises` module and returns a an object compatible with the `Filesystem` interface
    - `toNodeFs` - wraps the `Filesystem` api to behave like (a subset of) the node api. This new interface is called `NodeishFilesystem`. This is what we can pass to isomorphic-git.
    - `createMemoryFs` - same as before, creates a new memory filesystem with the `Filesystem` api
 - Two test suites have been created to test compatibility between these apis
     - `NodeishFilesystem.test.ts` tests compatitiblity between node's `fs` interface and the interface created by `toNodeFs()`
     - `MemoryFilesystem.test.ts` tests compatibility between the filesystem created by `createMemoryFs` and the one created by `fromNodeFs`

## API changes (from the original `Filesystem` api)
- Instead of returning `undefined` when a target is invalid, errors are now thrown with informative error messages.
- `dirname` and `basename` are now defined as part of the `Filesystem` api
- A new `FileData` type has been defined to allow for the possibility of support for `Uint8Array` in the future

## Misc
- `$fs` has been redefined to be `Filesystem` instead of a subset. Since the api has been changed, the `Filesystem` interface should work in any environment now (browser, node, vscode, etc.)
    - This also prevents accidental type clashes with the node api (i.e. node fs fits the interface, but the behavior of the api is different)

# Notes
- Ignore commit [9aae865](https://github.com/inlang/inlang/pull/559/commits/9aae8652f625747ef8b3b9210143ac1762465d8a). I think this was caused  by my computer running out of memory which led turbo to have some strange issues with caching.
- I think the function names could possibly be changed. `toNodeFs(createMemoryfs())` doesn't 'read' entirely logically

source-code-git/fs/src/schema.ts
```diff
@@ -0,0 +1,59 @@
...
+		 *
+		 * @example
+		 *   fs.toJson({ dir: "/home/user/project" })
+		 */
+		dir?: string
+	}) => Promise<Record<string, any>>
+	fromJson: (json: Record<string, string>) => Promise<Filesystem>
+}
+
+export interface NodeishFilesystem {
```
samuelstroschein, line 49: What is nodish filesystem for? 

I guess to map a filesystem back to node apis for isomorphic git?

samuelstroschein:
I will merge this PR this evening (in 2-3 hours). Have to eat dinner now.

araknast:
> I will merge this PR this evening (in 2-3 hours). Have to eat dinner now.

Ok, keep in mind the `NodeishFilesystem` api is currently incomplete, and subject to change.

PR #562, opened 2023-04-19
Editor summary

NiklasBuchfink:
@NilsJacobsen and I added the new summary section and refactored the filter section

Change Log:
- new listheader with lint summary
- new filter section styling and design
- add fallback if linting is not in config
- fix browser language detection


felixhaeberle:
Nice boys!! 🎉

felixhaeberle (commented):
@NilsJacobsen @NiklasBuchfink 

hey boys, just a small thing: is this intentional? I think its a big big and doesnt provide much info about what this is.

Bildschirmfoto 2023-04-19 um 16 10 11

and now we are a bit inconsistent to this box: 

Bildschirmfoto 2023-04-19 um 16 10 18

what do you think? :)

really great work ✨
the filter in the summary looks sooo good now and the interaction is really great 😍

felixhaeberle:
@NiklasBuchfink As I took my time to review this, it would be great to get an answer 😄 – otherwise please don't assign reviewers next time.

NiklasBuchfink:
@felixhaeberle I thought we talked about that briefly on the call. I thought you could help with the failed badge test, but we already figured out that the cause was the environment variable import or the package-lock file.

What information should be included in the dropdown menu behind the filter button? Do you have any suggestions for improvement or what you miss/expect there?

About the inconsistencies with the other menu, do you expect it to look the same? Then we have to go through the ALL and None, the heading and the controls that help with the interaction. I think the difference between "Add" and "Select" is fine. A unified look and thought out filtering system like Linear's would be great, but might be a bit overkill for the two filters right now. Do you have any suggestions on how to improve this issue without completely rewriting the filter selections?

felixhaeberle:
> What information should be included in the dropdown menu behind the filter button? Do you have any suggestions for improvement or what you miss/expect there?

I would expect the same options as I have in the "summary" now. Because thats what I want to achieve with the filter? Instead, I get a "lint" option which I don't understand. 

> Do you have any suggestions on how to improve this issue without completely rewriting the filter selections?

I think there is a lot to improve!
Vgl. those: Multiple icon sizes, multiple font colors, multiple font spacings, multiple font weight. 
I think the current state is very noisy right now. I tried some corrections.

Previous
Bildschirmfoto 2023-04-20 um 10 33 44

After
Bildschirmfoto 2023-04-20 um 10 33 17

felixhaeberle:
@samuelstroschein This should maybe move into a new issue.

NiklasBuchfink:
> I would expect the same options as I have in the "summary" now. Because thats what I want to achieve with the filter? Instead, I get a "lint" option which I don't understand.

@felixhaeberle Making the filter dropdown exactly like the summary would not work if the language filter is removed and needs to be added. Also, as a user, I would suggest that the lints then act as AND and not OR.


> I think there is a lot to improve!
Vgl. those: Multiple icon sizes, multiple font colors, multiple font spacings, multiple font weight.
I think the current state is very noisy right now. I tried some corrections.

I agree with that. @NilsJacobsen I think we should set the [design tokens of shoelace](https://shoelace.style/getting-started/customizing) right first. At the moment it is pretty messy and we run in this problems again and again.

Side note to _after_ , the reference language cannot be unselected and is therefore grayed out.

felixhaeberle:
> @felixhaeberle Making the filter dropdown exactly like the summary would not work if the language filter is removed and needs to be added. Also, as a user, I would suggest that the lints then act as AND and not OR.

That makes perfect sense – please go for it!! 👍

> I agree with that. @NilsJacobsen I think we should set the [design tokens of shoelace](https://shoelace.style/getting-started/customizing) right first. At the moment it is pretty messy and we run in this problems again and again.

Looking at the code, all variables where there but partly not applied – for example, the "All" and "None" have a parent font coloring of the right color and are itself on color:inherit; – but somehow the specificity is wrong and it gets messed up with another color somehow. ... different icon & fonts sizes + font weights also, as I mentioned. 

> Side note to after , the reference language cannot be unselected and is therefore grayed out.

You are right, but blue (primary in general) indicates some sort of selection and is not the best way to showcase a disabled state. You said the solution: grayed out would be better.

@NiklasBuchfink Do you open a new issue for that? Thank you 🙏🙏

NilsJacobsen:
@felixhaeberle thanks for the feedback we're coming back to that when we make the next editor fix round. If do not have something that is completely braking the user experience we're gonna work on that after the week after sveltekit summit.

felixhaeberle:
@NilsJacobsen @NiklasBuchfink Alright, you are in charge of creating issues and monitor it! :)

PR #563, opened 2023-04-19
refactor: align placeholder with upcoming message format 2.0

samuelstroschein:
_Closes #517_

I double-checked with both Flutent and the upcoming MessageFormat 2.0 AST. 

1. variable (references) are called variable references, not only "variable". I assume this because syntax can define variables that needs a separate node. 

2. expressions are unrelated to variables. variables are not nested in an expression. a placeholder contains either a variable reference or an expression whereas an expression is "a function" (something difficult like gender, pluralization etc)

We will introduce expressions at a later point in time. Unblocking @NilsJacobsen. @ivanhofer merge after review

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17269,6 +17269,7 @@
 			"name": "@inlang/badge",
 			"dependencies": {
 				"@inlang/core": "*",
+				"@inlang/telemetry": "*",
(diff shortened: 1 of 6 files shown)


ivanhofer (commented):
Looks good. Can you please fix the tests?

PR #564, opened 2023-04-19
Use our fork as the isomorphic-git implementation

araknast:
One of the concerns with a soft fork of isomorphic-git is being able to rapidly deploy it into our codebase. I propose that we avoid pushing a new npm package and instead pull changes from our [inlang/isomorphic-git](https://github.com/inlang/isomorphic-git) fork directly as part of the build process. This requires the fork to host build artifacts, which has already been done.


samuelstroschein:
@araknast feel free to merge if it works

araknast:
At the moment there doesn't seem to be an urgent need, so I was planning keep this open until tests are migrated and memoryFs is made compatible. Then we can merge those all together.

araknast:
Potentially some organizational decisions to be made here. At the moment the structure is somewhat of a hybrid between the way isomorphic-git does things and the way the rest of the project is structured. Currently the structure is:
```
├── src
│   └── raw
│       ├── README.md
│       ├── index.ts
│       └── tests
│           ├── (*.test.ts)
│           ├── fixtures
 |              |    ├── (test-*)
│           │   └── (test-*.git)
│           ├── makeFixture.ts
│           └── transform.sh
```

At the moment we are fetching isomorphic-git from our fork into the node_modules directory. The tests are then copied and transformed to work with Vitest by the [`transform.sh`](https://github.com/inlang/inlang/blob/inlang-isomorphic-git/source-code-git/client/src/raw/tests/transform.sh) shell script. This script also skips a number of tests that are irrelevant or currently broken. At the moment the transform script is run by hand, and so all 76 test and 97 fixtures are being committed to the repo. This could also be done automatically, but it would probably make the process a bit more difficult to debug, and add to the complexity.

Each test need to be run in a fixture, which is created by [`makeFixture.ts`](https://github.com/inlang/inlang/blob/inlang-isomorphic-git/source-code-git/client/src/raw/tests/makeFixture.ts) This is nearly the same as our `mockEnvironment`, and most of the `copyDirectory` code has been copied from there. The only difference is that `makeFixture` copies symbolic links, and also supports the wider `NodeishFilesystem` api.

The tests require a mock server and a cors proxy to run, so `@isomorphic-git/cors-proxy`, and `git-http-mock-server` have been added as dependencies, and are run as part of the test. In order to properly handle setup and shutdown of the the mock server and proxy, a bit of POSIX shell trickery is used, so at the moment I don't think building on windows is possible. This behavior can be made to work cross platform using ShellJS if necessary.

Finally, as mentioned previously, the Filesystem has been refactored. It now supports symlinks as well as tracking stat information.

samuelstroschein:
Looks good to me. 

Disclaimer: I haven't reviewed the PR thoroughly. The PR is too large to make an in-depth review. But, that should be fine. If it works, it works. 

> Potentially some organizational decisions to be made here. At the moment the structure is somewhat of a hybrid between the way isomorphic-git does things and the way the rest of the project is structured. Currently the structure is:

We could totally create a new directory in `source-code-code` called `isomorphic-git` which resembles isomorphic git 1:1. There is no need to squeeze isomorphic git into "raw". We are free to import local code as we please in the monorepo.

samuelstroschein (approved):

araknast:
> Disclaimer: I haven't reviewed the PR thoroughly. The PR is too large to make an in-depth review. But, that should be fine. If it works, it works.

Sounds good, BTW I tried to put a lot of information into the commit messages., so hopefully that should give a good overview of what's changed.

> We could totally create a new directory in source-code-code called isomorphic-git which resembles isomorphic git 1:1. There is no need to squeeze isomorphic git into "raw". We are free to import local code as we please in the monorepo.

True I think this would probably be a bit cleaner organizationally. The issue is that isomorphic-git has a messy build system that wouldn't integrate well with ours. For now I think it's probably better to keep the building of isomorphic-git separate from the monorepo, and just keep the build artifacts in the repo. Once we can build isomorhpic-git simply with `tsc` we can probably revisit this.

PR #569, opened 2023-04-21
Tiptap prototype

NilsJacobsen:
Hey folks, 

this PR is part of the approach of getting to a rich ast editor, where you could use variables and specific formats. 

**First step:** add a rich text editor (we use tiptap) **This PR**
**Second step:** integrate placeholders from the ast
**Third step:** as requirements get higher we can also implement our own editor

The editor in this branch should have the same spec as the current text area.
U are welcome to check out and test it.

image


NiklasBuchfink:
@NilsJacobsen Some improvements:
- [ ] The styling is slightly different to be larger/bold, but I'm not sure if the current live version of the editor uses the system font or the inter-variable. So this could be the problem.
- [ ] If I accidentally delete something, I can't go back in the history with Ctrl/CMD + Z
- [ ] After a commit I lose focus
- [ ] When tapping into a input and start writing the blue dot appears instead of the commit button

PR #570, opened 2023-04-21
rfc plugin api

samuelstroschein:
@inlang/team @inlang/contributor please review this RFC by tomorrow (27. April) **concerning the API we expose for plugins**. The implementation details can be changed at any time. 

1. Read the RFC [here](https://github.com/inlang/inlang/blob/b91eec17283da6ef2e137dd4884b681454561568/rfcs/plugin-api/RFC.md)
2. Take a look at experiment-2 which defines how to write plugins, [see](https://github.com/inlang/inlang/pull/570/files#diff-cde014b11f183d1a896509e58358081c985bd9ba244c5d7efac31edd88b2dd65R5-R16)
3. Make comments on the RFC itself [here](https://github.com/inlang/inlang/pull/570/files) or in this pull request.


rfcs/plugin-api/RFC.md
```diff
@@ -0,0 +1,279 @@
...
+```
+
+**Why this approach seems good**
+
+1. We can extend the `plugins` property to support more complex use cases in the future:
+
+```ts
+type Plugin = {
+	id: string
+	defineConfig: (env: Env) => Promise<Config>
```
ivanhofer, line 274: we also need to pass the existing config to the `defineConfig` function

rfcs/plugin-api/RFC.md
```diff
@@ -0,0 +1,279 @@
...
+	)
+	const sdk = await env.$import("https://cdn.jsdelivr.net/gh/inlang/inlang@1/dist/sdk-js/index.js")
+
+	const standardLintRules = await env.$import(
+		"https://cdn.jsdelivr.net/gh/inlang/inlang@1/dist/sdk-js/index.js",
+	)
+
+	return {
+		referenceLanguage: "en",
+		plugins: [
```
ivanhofer, line 186: the only downside I see is that a `plugin` can be anything. Maybe it is confusing to have a core functionality like reading and writing resources named "plugin"?

rfcs/plugin-api/RFC.md
samuelstroschein: Yes, would be done with `setupPlugins(defineConfig)`. In other words, we would have a "post-process" step after reading the config. The good news: The post-process step could be used for other things as well.

rfcs/plugin-api/RFC.md
samuelstroschein: Hm. I interpret this API as a plugin has a set of functions it can call. One of them is `defineConfig` which defines the config. Because a plugin defines the `readResources` and `writeResources` properties of the config, the user is not required to define those properties itself. 

Do you have an alternative proposal?

rfcs/plugin-api/RFC.md
samuelstroschein: On that note, we should not call the post-process/post-import function `setupPlugins` in anticipation of adding other functionality to the function in the future.

rfcs/plugin-api/RFC.md
ivanhofer: the alternative is coming up with a better name for core functionality e.g.
```diff
return {
	referenceLanguage: "en",
-	plugins: [jsonPlugin],
+	ressources: [jsonRessources],
}
```
But `jsonRessoureces` would still be able to modify other properties on the config and not just `Ressource` related ones. This might also confuse people.
So not sure if it is worth coming up with another system.

ivanhofer: the alternative is coming up with a better name for core functionality e.g.
```diff
return {
	referenceLanguage: "en",
-	plugins: [jsonPlugin],
+	ressources: [jsonRessources],
}
```
But `jsonRessoureces` would still be able to modify other properties on the config and not just `Ressource` related ones. This might also confuse people.
So not sure if it is worth coming up with another system.

samuelstroschein: It seems confusing/is another concept we need to explain. Thus, I refrain from further investigating `ressources`.

samuelstroschein:
@inlang/team @inlang/contributor please review this RFC by tomorrow **concerning the API we expose for plugins**. The implementation details can be changed at any time. 

1. Read the RFC [here](https://github.com/inlang/inlang/blob/b91eec17283da6ef2e137dd4884b681454561568/rfcs/plugin-api/RFC.md)
2. Take a look at experiment-2 which defines how to write plugins, [see](https://github.com/inlang/inlang/pull/570/files#diff-cde014b11f183d1a896509e58358081c985bd9ba244c5d7efac31edd88b2dd65R5-R16)
3. Make comments on the RFC itself [here](https://github.com/inlang/inlang/pull/570/files) or in this pull request.

rfcs/plugin-api/experiment-2.ts
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * @example
+ *   plugins: [
+ * 	 	myPlugin({
+ * 	   	pathPattern: "hello",
+ * 	 	})
+ *   ]
+ */
+type PluginConfigFunction<PluginConfig> = (
+	pluginConfig: PluginConfig,
```
ivanhofer, line 29: I would rename `pluginConfig` to `settings` or something other. so wen don't have to explain different `config`s.

rfcs/plugin-api/experiment-2.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export const myPlugin = createPlugin<{ pathPattern: string }>(({ pluginConfig, env }) => {
+	return {
+		id: "samuelstroschein.plugin-json",
+		defineConfig: (config) => {
+			if (pluginConfig.pathPattern === undefined) {
+				throw new Error("pathPattern is required")
+			}
+			config.readResources = readResources({ pluginConfig, env })
+			config.languages = getLanguages({ pluginConfig, env })
```
ivanhofer, line 13: We will probably regret mutating objects in the future. I think `defineConfig` should return a new copy of the config. See `experiment-3.ts`

rfcs/plugin-api/experiment-2.ts
```diff
@@ -0,0 +1,45 @@
...
+
+type Plugin = {
+	id: string
+	defineConfig(config: Partial<Config>): void
+}
+
+function createPlugin<PluginConfig>(
+	callback: (args: { pluginConfig: PluginConfig; env: EnvironmentFunctions }) => Plugin,
+): PluginConfigFunction<PluginConfig> {
+	return (pluginConfig) => (env) => callback({ pluginConfig, env })
```
ivanhofer, line 40: why does `env` and `config` get passed in two different steps? I would do it like in `experiment-3.ts`

rfcs/plugin-api/RFC.md
```diff
@@ -0,0 +1,279 @@
...
+			// @ivanhofer @jannesblobel the "collection" discussion is resolved with this approach.
```
ivanhofer, line 186: I don't get why this would solve the "collection" problem

ivanhofer (commented):
I made some minor suggestions

rfcs/plugin-api/experiment-2.ts
samuelstroschein: Experiment-3 requires us to increase the surface are of the plugin API manyfold. 

We would need to pass `env` to every function in `Plugin` instead of only once when creating a plugin. Passing the env once enables us an easier way forward to remove env in favor of better ergonomics in the future. We would only need to change `createPlugin` and not every function in `Plugin`.

rfcs/plugin-api/experiment-2.ts
felixhaeberle: Returning a new config could definitely make sense!! What speaks against a new copy of the config and for mutating?

rfcs/plugin-api/experiment-2.ts
samuelstroschein: Hm, I find `pluginConfig` relatively straightforward. `settings` could mean settings of inlang (at least to me). 

@inlang/team please vote on `pluginConfig` or `settings`.

felixhaeberle: Yes, and probably 
- `defineConfig` -> `config` 
and 
- `createPlugin` -> `plugin` 
if they are no naming clashes with this approach.

samuelstroschein: @inlang/team please vote on `pluginConfig` or `settings`, or `options`.

rfcs/plugin-api/experiment-2.ts
ivanhofer: ah, got it!

felixhaeberle (commented):
I also made some suggestions :)

rfcs/plugin-api/RFC.md
samuelstroschein: A collection becomes a plugin. Thereby, the concept of a lint collection is not required anymore :)

ivanhofer: The collection has nothing to do with the core of inlang. It is something for `lint`-plugins. And there we still have the same problem

samuelstroschein: The debate about lint collections was "is the concept of lint collections required [to define multiple lint rules at once]?". We decided to move lint collections out of core because we were in agreement that maybe not. With the introduction of plugins, the answer is a definitive no. A plugin can define multiple lint rules at once. 

After introducing the plugin API, the code in https://github.com/inlang/standard-lint-rules/blob/main/src/collection.ts is redundant. 

Before:

```ts
export async function defineConfig(env) {
  // import the rules you want to use
  const { standardLintRules } = await env.$import(
    "https://cdn.jsdelivr.net/gh/inlang/standard-lint-rules@1/dist/index.js"
  );

  return {
    referenceLanguage: "en",
    languages: ["en", "de", "fr"],
    lint: [
      standardLintRules() // <-- concept of a collection
    ],
  };
}
```

After:

```ts
export async function defineConfig(env) {
  // import the rules you want to use
  const { standardLintRules } = await env.$import(
    "https://cdn.jsdelivr.net/gh/inlang/standard-lint-rules@1/dist/index.js"
  );

  return {
    referenceLanguage: "en",
    plugins: [
      pluginJson(),
      standardLintRules() // <-- just a plugin
    ],
  };
}
```

rfcs/plugin-api/experiment-2.ts
samuelstroschein: **Reasons against mutable config**

- requires additional logic/takes longer to implement because a merging logic needs to be implemented. @ivanhofer spreading will break because it's shallow.
- vite has a mutable and immutable config definition because mutable is required when merging logic fails.
- (performance hit. maybe important in the browser because) 

**But, we can have a simple trick to path the way for immutable config definition in the future**

We can enforce a return value. By forcing a return value, we can implement immutability at a later point by passing a copied `config` object into `defineConfig`. 

```diff
type Plugin = {
	id: string
-	defineConfig(config: Partial<Config>): void
+	defineConfig(config: Partial<Config>): Partial<Config>

}
```

```diff
export const myPlugin = createPlugin<{ pathPattern: string }>(({ pluginConfig, env }) => {
	return {
		id: "samuelstroschein.plugin-json",
		defineConfig: (config) => {
			if (pluginConfig.pathPattern === undefined) {
				throw new Error("pathPattern is required")
			}
			config.readResources = readResources({ pluginConfig, env })
			config.languages = getLanguages({ pluginConfig, env })
+			return config
		},
	}
})
```

**How Vite defines the Plugin API**

Vite uses two separate callbacks which we have in one. 

1. `config` to define the config. 
2. `resolvedConfig` to "validate" the resulting config of all plugins. 

I applied Vite's design to our API:

```ts
type Plugin = {
	id: string
	config: () => Partial<Config>
	resolvedConfig: (config: Config): void
}
```


```ts
export const myPlugin =
  createPlugin <
  { pathPattern: string } >
  (({ pluginConfig, env }) => {
    return {
      id: "samuelstroschein.plugin-json",
      config: () => {
        if (pluginConfig.pathPattern === undefined) {
          throw new Error("pathPattern is required");
        }
        return {
          readResources: readResources({ pluginConfig, env }),
          languages: getLanguages({ pluginConfig, env }),
        };
      },
      resolvedConfig: (config) => {
        if (config.ideExtension === undefined) {
          throw Error(
            "config.ideExtension must be defined for this plugin to work."
          );
        }
      },
    };
  });
```

I think vite's design is better for the following reasons: 

- The ordering of plugins matters less, leading to increased DX. The configs of all plugins will be merged together. Thus, if a plugin requires `readResources` to be defined, the plugin that defines `readResources` doesn't have to be defined before the consuming plugin. 
- Developers have to write less code. By us merging the configs together, developers don't have to write code that, for example, checks if the lint array is `undefined` before adding a lint rule. 
- We can have error handling in the merge logic e.g. `"readResources has been defined by a previous plugin"` in the future.

Going forward, I would choose the Vite plugin API with `config` that is immutable and `resolvedConfig` as an option for plugins to validate configs.

rfcs/plugin-api/RFC.md
ivanhofer: In my opinion nothing really changes. A lint plugin can already return an array of lint rules

rfcs/plugin-api/experiment-2.ts
ivanhofer: Agree. Merging the config internally makes sense! I didn't want to suggest this because in the past my "bullet-prove" ways of doing things were rejected in favor of simpler solutions ^^

rfcs/plugin-api/experiment-2.ts
```diff
@@ -0,0 +1,48 @@
...
+ * 	 	})
+ *   ]
+ */
+type PluginConfigFunction<PluginConfig> = (
+	pluginConfig: PluginConfig,
+) => (env: EnvironmentFunctions) => Plugin
+
+type Plugin = {
+	id: string
+	defineConfig(config: Partial<Config>): void
```
ivanhofer, line 34: this must return a `MaybePromise`! Or else we won't be able to add the json plugin in the SDK plugin as a fallback if missing

samuelstroschein: Yes, in my mind I was already about to change this. Thank's for pinging!

rfcs/plugin-api/experiment-2.ts
samuelstroschein: Okay, going for `settings`.

NilsJacobsen:
Hey folks, just some loose thoughts about variableReferences. 
I had a talk with @felixhaeberle and we weren't sure how to approach this. 

Situation:
We have the read/write plugins that convert repo specific stuff to an act object.
And we have the IDE extension which is communicating with the repo specific stuff but is not handling an ast under the hood.

Problem:
When I want to configure my inlang config I have to setup my read/write plugins because they need to understand my repo but as far as I understood this should be the main interface between inlang and the code. But now when it comes to the IDE extension, this plugin also needs to understand my code. So there need to be a second interface between inlang ide-extension and the code. Especially when it comes to variable/placeholder syntax ...

Expectation:
My first thoughts on that before we entered our discussion was that it is pretty clear that the ide extension will not directly manipulate the files, cause that needs to be through the read and write functions. This way we can ensure that the ide extension works in every setup. 

But I can also understand that we do not need to work with the ast here. So that is probably to complex for now. 
What are your thoughts on that?

ivanhofer:
I never took a look at the implementation details but from my perspective all applications that build on top of inlang need to deal with the AST. The extension should not care wether Resources are stored as json or yaml.
If implemented in a reliable way, the IDE extension would need to parse the text you want to extract into an AST. The IDE extension would need a way to know where variables are present in a (programming-)language agnostic way e.g. `I am a {variable}` (Svelte markup) and `I am a ${variable}` (template literal) or `"I am a "+ variable` string concatenation.

samuelstroschein:
@NilsJacobsen @felixhaeberle 

> But I can also understand that we do not need to work with the ast here. So that is probably to complex for now.
What are your thoughts on that?

You need to deal with the AST in any inlang application. This is how "file-based" applications work. `readResources` and `writeResources` give applications the possibility to parse and serialize the AST. Nothing more, nothing less. 

> And we have the IDE extension which is communicating with the repo specific stuff but is not handling an ast under the hood.

The ide extension handles the AST. 

https://github.com/inlang/inlang/blob/63257d02762171b301c78fbfe58056c4d48145d4/source-code/ide-extension/src/commands/extractMessage.ts#L71-L81

PR #580, opened 2023-04-25
refactor-fs

samuelstroschein:
@araknast I went through the fs implementation and thought I speed up things by refactoring the interface according to the needs of inlang. While refactoring, I also cleaned up a few things like renaming `schema` to `interface`. 

**Notes**

- We can use a strict node:fs/promises subset 
    - I made a mistake (in case you thought about raising your voice, you should have. Do so next time! :) )
    - No need for node wrappers if we use a strict subset. The key here was to require the encoding `utf-8` https://github.com/inlang/inlang/pull/580/files#diff-5f166c7b884eb1bbba5957f27b0d20548b0ef9464633356b1512d842fa0b95fbR9. Regarding `Buffer` support which must be `Uint8Array` for the browser, we can deal with that when the time comes.
    - Define utility functions like `fromJson` and `toJson` as external functions (I mean we decided on a functional approach, why haven’t we made use of it? :D )

- `Filesystem` should not be extended 
    - Defeats the purpose of having one API for different filesystems [example](https://github.com/inlang/inlang/blob/1248cb5742fd4f1d59e98b3374c63140c92258d8/source-code-git/fs/src_old/schema-internal.ts#L5-L8)

- No need for `this` (and thus no filesystem extensions)
    - Use local state in the function scope [example](https://github.com/inlang/inlang/blob/1248cb5742fd4f1d59e98b3374c63140c92258d8/source-code-git/fs/src_old/createMemoryFs.ts#L70-L72)
    - Use external functions  [example](https://github.com/inlang/inlang/blob/1248cb5742fd4f1d59e98b3374c63140c92258d8/source-code-git/fs/src_old/createMemoryFs.ts#L74-L87)

- the "old code" (I reused yours) exists under `src_old` for comparison

**Next steps**

I created this issue https://github.com/inlang/inlang/issues/581 to provide step-by-step guidance on what's next. 

**Additional information**

How can I improve the communication of expectations/requirements? 

Ideally, I would not have felt the urgency to step in and create this PR to speed things up. The filesystem thing has been going on for three weeks. With better communication/alignment/requirements, we likely could have settled on an implementation within a week. Curious about your thoughts on improving our collaboration. Feel free to DM/call me on Discord :)


araknast:
To speed up deployment and avoid having to backport bugfixes, how about we export a new function `createNodeishFs () => toNodeFs(createMemoryFs())` ?

For now, we can have that and the `NodeishFilesystem` interface be the only exports but if down the line we decide we want to expose the underlying `Filesystem` api we can do that. Also significantly less work and I could stay on schedule.

samuelstroschein:
> To speed up deployment and avoid having to backport bugfixes, how about we export a new function createNodeishFs () => toNodeFs(createMemoryFs()) ?

Speeding up development might not be important. The blocker for the rest of inlang is the lack of a Filesystem API and `createMemoryFs`. Both are addressed in this PR by defining the Filesystem as "nodeish" without wrappers (making the rest of the codebase compatible) and `createMemoryFs` for tests and `mockEnvironment`. 

If we can merge this, I think the rest of inlang is not blocked. Thus, if you need more time to refactor things after merging this PR, no problem. 

> For now, we can have that and the NodeishFilesystem interface be the only exports, but if down the line we decide we want to expose the underlying Filesystem api we can do that. Also significantly less work and I could stay on schedule.

I am not sure I follow. We want to export one Filesystem API. If we expose multiple filesystem APIs, we will get combability issues. 

Related: Wouldn't this `Filesystem` API defined here, compatibly with node, make isomorphic git changes much easier and thereby speed you up instead of slowing down?

araknast:
> I am not sure I follow. We want to export one Filesystem API. If we expose multiple filesystem APIs, we will get compatibility issues.
I was mainly concerned that we may at some point want to go back to having two apis, and so this way that option remains open. This might happen if we need to break from the node api for some reason, for example reactivity, then the breaking feature can exist in `Filesystem` and `NodeishFilesystem` can remain strictly compatible. If you envision that we will not need to break from the api for the foreseeable future, then that's not an issue.

> Related: Wouldn't this Filesystem API defined here, compatibly with node and thereby isomorphic git, make isomorphic git changes much easier and thereby speed you up instead of slowing down?
The difference would be between passing `createMemoryFs` to isomorphic-git and passing `createNodeishFs` to isomorphic-git, i.e. negligible. 

> Speeding up development might not be important. The blocker for the rest of inlang is the lack of a Filesystem API and createMemoryFs. Both are addressed in this PR by defining the Filesystem as "nodeish" without wrappers (making the rest of the codebase compatible) and createMemoryFs for tests and mockEnvironment.

"Speeding up deployment" was mostly referring to myself,  since I will need to re-implement `fromJson` and `toJson` as well as backporting a bugfix related to path resolution which I have locally but are based on the previous two wrapper branch.

Ultimately if you want to merge go ahead and do so. My main concern was leaving the two wrapper approach available as an option in case we want to go back later for some reason. If you're sure this is the approach we want to taking going forward then that's fine by me, although this will probably push the migration of isomorphic-git tests back to next week.

samuelstroschein:
> Ultimately if you want to merge go ahead and do so. My main concern was leaving the two wrapper approach available as an option in case we want to go back later for some reason. If you're sure this is the approach we want to taking going forward then that's fine by me, although this will probably push the migration of isomorphic-git tests back to next week. 

I will go ahead and merge this tomorrow. 

I think one Filesystem API is the way to go. But, I can't foresee the future. The following reasons convince me of one filesystem at the moment: 

1. if we export two filesystems now and realize that we don't need one of them, we have a breaking change. On the flip side, if use cases emerge that require two filesystem apis, we can _introduce_ a second filesystem API in a non-breaking change. previous code would work just fine because previous code wouldn't require additionally introduced APIs. 

2. the reason for the two filesystem approach seems to stem from isomorphic-git only handling nodeish APIs. Likely better to refactor isomorphic-git to accept the single filesystem API instead of exposing two different ones. 

I will go to bed now. Feel free to object. Else, I merge this tomorrow morning CEST time and introduce the filesystem API to @inlang/core 🔥

araknast:
I agree on only exporting one filesystem now, my only concern was exactly how we do it. I.e instead of only exporting `Filesystem` and `createMemoryFs` we only export `NodeishFilesystem` and `createNodeishFs`. This means that in the future we can export `Filesystem` if it does turn out we need two apis, or keep it unexported (and eventually remove it) if we don't.  

Finally, two things:
1. The current `Filesystem` api is not node compatible, we would need to enforce the `{ recursive: true }` option on `rm` and `mkdir` the same way we do for the `encoding` option to `readFile`, but this would break `isomorphic-git` compatibility. 
If we want `createMemoryFs` to return a nodeish filesystem then we need to use the code from `toNodeFs` which accepts these options and handles them appropriately. This was the reason for having the `NodeishFilesystem` test suite which tested against node's `fs` implementation.
2. In order to be able to use `Filesystem` with isomorphic-git, we at least need to implement `stat` `lstat` `unlink` and `readdir` from node's fs as well. This was another reason I wanted to separate node compatibility into a wrapper, since these are features that would need to be stubs, and so it wouldn't make much sense to expose them to developers when we have no concept of symbolic links in our filesystem

I'm not sure if I'm explaining myself well, but in short, if you do end up merging this, could you please rename `Filesystem` to `NodeishFilesystem` and `createMemoryFs` to `createNodeishFs` before doing so?

samuelstroschein:
I think there are several misconceptions about how the Filesystem API will be used in the codebase. I think it's best if I merge this PR tomorrow and actively use the Filesystem API in the codebase. We can always refactor after merging. 

>  I.e instead of only exporting Filesystem and createMemoryFs we only export NodeishFilesystem and createNodeishFs

I will rename `Filesystem` to `NodeishFilesystem`. There seems to be a misconception about memoryFs though. The goal of memoryFs is to have an in-memory filesystem that obeys the Filesystem API (and is a demonstrator that two different Filesystem APIs likely won't work because what would `createMemoryFs` be `Filesystem` or `NodeishFilesystem`?).

> The current Filesystem api is not node compatible, we would need to enforce the { recursive: true } option on rm and mkdir the same way we do for the encoding option to readFile, but this would break isomorphic-git compatibility.

I don't understand why we need to enforce `{ recursive: true }` and just realized that my enforcement explanation might have been confusing. Enforcing `{ encoding: "utf-8" }` seems like a must (?) to avoid the `Buffer` problem. Recursive can be defined and doesn't have to be enforced though, the types won't change. 

> In order to be able to use Filesystem with isomorphic-git, we at least need to implement stat lstat unlink and readdir from node's fs as well. This was another reason I wanted to separate node compatibility into a wrapper, since these are features that would need to be stubs, and so it wouldn't make much sense to expose them to developers when we have no concept of symbolic links in our filesystem 

All filesystem implementations like memoryFs should have all functions that are required for isomorphic git. Thus, no problem here. Exposing the filesystem happens in @inlang/core. However, @inlang/core uses a subset of the Filesystem https://github.com/inlang/inlang/blob/d94c771c443276cf417425b4fb9433a6c7c322f5/source-code/core/src/config/environment-functions/%24fs.ts. 

A bit confusing, I hope to shine light tomorrow. 

1. `@inlang-git/*` requires a filesystem API that implements the minimal functions required by isomorphic git. 
2. `@inlang/core` exposes a filesystem API for external developers to build on. 

2) exposes a subset of 1) and 1) itself is a subset of node/fs:promises.

samuelstroschein:
@araknast FYI the buffer subtyping via `UInt8Array` seems to work. No type error with the following definition:

```ts
export interface NodeishFilesystem {
	writeFile: (path: string, content: string) => Promise<void>
	readFile:
		| ((path: string, options: { encoding: "utf-8" }) => Promise<FileData>)
		| ((path: string) => Promise<Uint8Array>)
	readdir: (path: string) => Promise<string[]>
	mkdir: (path: string) => Promise<void>
	rm: (path: string) => Promise<void>
}


```

PR #598, opened 2023-04-27
572 refactor 560 to use ast transforms

benjaminpreiss:


ivanhofer (approved):
Looks great!

I refactored the code because we could not bundle it. I have replaced `svelte-preprocess` with `vitePreprocess` to make it work.
I also have replaced the tests with `inline-snapshots` so we can see the results even better.

I'll merge this for now, but please take a look at the first test (`it.todo(...`). This test fails to run because `import { i, languages, switchLanguage } from "@inlang/sdk-js"` gets transformed to const { i: i, i: i...".

PR #600, opened 2023-04-28
Deeper node api compliance for `NodeishFilesystem`

araknast:
Here are most of the changes that I had been working on for the wrapper version of `NodeishFilesystem`, now made to work with the new version. The biggest noticeable change is proper support for the `recursive` options to rm and mkdir. `rmdir` has also been added as a function on the API since we will need it for isomorphic-git. An `encoding` option has been defined but is not yet supported.

Notably I had to set the `options` argument to `readFile` to be of type `any` for it to allow the `{ encoding: TextEncoding } | TextEncoding` function signature, even though this should technically be a subset of the node type. I think this may warrant further investigation.

Not included is the `fromJson` and `toJson` functions I devised since @samuelstroschein 's implementation seems to be working already. I will try see if I can use my code to provide encoding support for those versions.

Finally I've unskipped the node tests for the filesystem in order to test compatibility, feel free to re-skip them if this behavior is not desired.


source-code-git/fs/src/interface.ts
```diff
@@ -8,17 +8,18 @@
  */
 export type NodeishFilesystem = {
 	writeFile: (path: string, data: string) => Promise<void>
-	readFile: (path: string, options: { encoding: "utf-8" }) => Promise<FileData>
+	/* NOTE: accept options as 'any' for now until we figure out how to get the node types to match */
+	readFile: (path: string, options?: any) => Promise<FileData>
```
samuelstroschein, line 12: We can't set `readFile` options to any. Developers will call `fs.readFile(path)` and wonder why they don't receive a `string` as a return value. I saw that mistake in the wild already.

source-code-git/fs/src/interface.ts
```diff
@@ -8,17 +8,18 @@
...
 	 * https://nodejs.org/api/fs.html#fspromisesmkdirpath-options
 	 *
 	 * Upon success, fulfills with undefined if recursive is false, or the first directory path created if recursive is true.
 	 */
-	mkdir:
-		| ((path: string) => Promise<void>)
-		| ((path: string, options?: { recursive: boolean }) => Promise<string>)
+	mkdir: (path: string, options?: { recursive: boolean }) => Promise<string | undefined>
 	rm: (path: string, options?: { recursive: boolean }) => Promise<void>
+	rmdir: (path: string, options?: any) => Promise<void>
```
samuelstroschein, line 21: `rmdir` has the same issue. If we allow `any` for the options, developers will use the filesystem differently and since we do not check for options that are being passed, different filesystem implementations will yield different results.

source-code-git/fs/src/interface.ts
samuelstroschein: Happened even to yourself :D

The node fs API is convoluted with a lot of overloads. I propose to use the latest [NodeJS 20 fs.promises API](https://nodejs.org/api/fs.html#fspromisesreadfilepath-options) with ONE WAY to do things. In most cases, this one way to do things follows an options as an object patten. 

- one way to do things increases DX
- reduces the amount of implementation work because we don't need conditional logic to check overloads

```ts
// the old way
fs.readFile("./", "utf-8")

// the new way
fs.readFile("./", { encoding: "utf-8" })

// the new wat offers additional flags without breaking changes [that we don't need to support now]
fs.readFile("./", { encoding: "utf-8", flag: "r" })
```

CleanShot 2023-04-28 at 10 32 31@2x

source-code-git/fs/src/interface.ts
```diff
@@ -15,9 +15,7 @@ export type NodeishFilesystem = {
 	 *
 	 * Upon success, fulfills with undefined if recursive is false, or the first directory path created if recursive is true.
 	 */
-	mkdir:
-		| ((path: string) => Promise<void>)
-		| ((path: string, options?: { recursive: boolean }) => Promise<string>)
+	mkdir: (path: string, options?: { recursive: boolean }) => Promise<string | undefined>
```
samuelstroschein, line 19: That is incorrect. The API I defined was correct. Node returns Promise<void> if no options are provided and a string if options are provided. 

See https://nodejs.org/api/fs.html#fspromisesmkdirpath-options. 

I will revert this change because we narrow the types, leading to better DX and consistency.

samuelstroschein: Tada. Narrowing the type immediately revealed a buy in memoryFs

CleanShot 2023-04-28 at 11 08 15@2x

samuelstroschein: NVM. 

- your implementation had no error. typescript was not widening "not implemented" to a string 

- TypeScript is not smart enough to match overloads to a naive type as node:fs/promises uses then :(

CleanShot 2023-04-28 at 11 58 53@2x


Okay, no overloading and narrow types then (at least for now. Maybe TypeScript wizard @ivanhofer knows how to fix this in the future).

samuelstroschein:
I made a PR that fixes the two mentioned problems. Please review #601, and merge upon approval.

araknast:
> Regarding rmdir which has been introduced for isomorphic git, can we remove rmdir from isomorphic git to reduce the surface API and ease the development of implementations?
>
> rmdir is used [3 times in isomorphic git (excluding tests)](https://github.com/search?q=repo%3Aisomorphic-git%2Fisomorphic-git%20rmdir&type=code). Given that rm with an if conditional can be used to achieve the same outcome, it seems advisable to remove rmdir from isomorphic git instead of increasing the API of the filesystem.
>
>EDIT: rmdir's only benefit over rm, of recursively deleting empty directories, [is even deprecated](https://nodejs.org/api/fs.html#fspromisesrmdirpath-options), further justifying the removal of rmdir.

We could, but isomorphic-git uses `rm` and `rmdir` to emulate recursive file removal for filesystems that don't natively implement it. It works under the assumption that `rm` may not necessarily work on empty directories, which may be true for some filesystems it supports. Replacing `rmdir` would mean potentially breaking support for these filesystems.

In regards to reducing the size of the surface api, we will already need to increase it considerably to implement `stat` as well as support for symbolic links `symlink`, `lstat`, `unlink`, etc. In this case it feels like replacing `rmdir` in isomorphic-git isn't worth the marginally smaller api.

samuelstroschein:
> We could, but isomorphic-git uses rm and rmdir to emulate recursive file removal for filesystems that don't natively implement it. It works under the assumption that rm may not necessarily work on empty directories, which may be true for some filesystems it supports. Replacing rmdir would mean potentially breaking support for these filesystems. 

Just double-checked, `node:fs/promises` throws if `rm` is used on a directory. Ergh, what an ugly API. In two years or so we can develop a "sane" filesystem API. 

But, we should not support `rmdir(path, { recursive: true })` given that the recursive option is deprecated in node fs and will be removed soon. Isomorphic git has one recursive `rmdir` call that we should change https://github.com/isomorphic-git/isomorphic-git/blob/2e82c293e297686d32ec754d49d5adfd6f2ce76a/src/commands/clone.js#L110. 

> In regards to reducing the size of the surface api, we will already need to increase it considerably to implement stat as well as support for symbolic links symlink, lstat, unlink, etc. In this case it feels like replacing rmdir in isomorphic-git isn't worth the marginally smaller API. 

I am aware. I am just poking around to see what we need and what not. The recursive option for `rmdir` for example seems like a mistake if we support it given that it will be removed soon.

source-code-git/fs/src/utilities/toJson.ts
```diff
@@ -37,6 +37,7 @@ export async function toJson(args: {
 	 * TODO: to make it more clear what it does? `resolveFrom` might be ambigous.
 	 */
 	resolveFrom: string
+	encodeOutput?: "utf-8" | "base64"
```
samuelstroschein, line 40: Can we always use `base64` or `Uint8Array` for `toJson` and `fromJson`? 

Enforcing an encoding will eventually break (?). What about a filesystem containing binary (base64) and non-binary data (utf-8)? Enforcing one encoding type will break a roundtrip because all files will become either binary or a string. 

I am wondering if we can set everything to be base64 or `Uint8Array` and don't define an encoding at all.

araknast: Hmm you're right, I didn't realize reading binary data into a `string` would corrupt it. Perhaps too much C programming from me :P

`Uint8Array` would be a great solution for in-memory JSON representations, but would be significantly less efficient than `base64` when writing the JSON to a text file (at least the way I'm imagining it).  I think `base64` provides a nice middle ground so I'll enforce that instead.

samuelstroschein: > Perhaps too much C programming from me :P 

Haha. One would assume that C would be harsher in this case than JavaScript :D 

If `base64` works for roundtrips of binary and non-binary data, let's choose base64 and get rid of the encoding argument.

samuelstroschein:
@araknast what do you think about `resolveFrom` aka only allowing absolute paths? 

I observed multiple "gotcha" moments when I (instinctively) used a relative path with memfs. Memfs screws up because it calls `process.cwd()` under the hood but even if not, the concept of a current working directory doesn't exist in memory, localstorage, solid, etc. filesystem implementations.

source-code-git/fs/src/utilities/toJson.ts
samuelstroschein: Edit: Just saw your reply on Discord. I replied there https://discord.com/channels/897438559458430986/1101312651637821520/1101974539367354541

araknast: > If base64 works for roundtrips of binary and non-binary data

In our case, since `memoryFs` is implemented using `Map<string, string>`, I'm not sure if round tripping binary data would work. I think we may need to change the implementation to `Map<string, Uint8Array` or something else.

araknast:
> what do you think about resolveFrom aka only allowing absolute paths?
Sure I think that's fine. It makes for a more consistent interface given that, as you said most memory filesystem's don't have any concept of a 'current working directory'.

source-code-git/fs/src/utilities/toJson.ts
araknast: I belive that when we call node's `readFile` with the `encoding` option set to `utf-8`, any invalid unicode characters are replaced with <?>, corrupting any binary data. If we wanted to be able to roundtrip binary and string data, we would need to base64 encode the raw `Buffer` objects, decode them, and then write them back to the filesystem as `Buffer` instead of `string`.

araknast: ah, looks like what was missing was to specify `{ encoding: 'binary' }` for `readFile`. This should be doable without using `Uint8Array` then.

PR #601, opened 2023-04-28
fix node types

samuelstroschein:
Fixes issues in #600. See the commit history for changes and explanations. 

@araknast We can enforce a super strict subset of node:fs/promises which we can widen over time. I encountered no type issues.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17009,7 +17009,7 @@
 		},
 		"source-code-git/fs": {
 			"name": "@inlang-git/fs",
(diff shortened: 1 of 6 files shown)


samuelstroschein:
PS the conditional logic code in https://github.com/inlang/inlang/pull/601/files#diff-2c3ed00955c5422c1b8b3e5f4be71038d6f7c22217f21fe1eee288da92ea5c5cR29-R35 is likely redundant now.

Issue #618, opened 2023-05-02
move ide extension config from inlang/core to an ide extension plugin


samuelstroschein:

felixhaeberle:
I've added a PR which refactors inlang/core and provides this as a plugin. #619

felixhaeberle:
🎉 this is now possible with a config like:

```js
/**
 * @type { import("@inlang/core/config").DefineConfig }
 */
export async function defineConfig(env) {
	const { default: jsonPlugin } = await env.$import(
		"https://cdn.jsdelivr.net/gh/samuelstroschein/inlang-plugin-json@2/dist/index.js",
	)
	const { default: sdkPlugin } = await env.$import(
		"https://cdn.jsdelivr.net/npm/@inlang/sdk-js@0.0.4/dist/plugin/index.js",
	)
	const { default: ideExtensionPlugin } = await env.$import(
		"https://cdn.jsdelivr.net/npm/@inlang/ide-extension-plugin@0.0.2/dist/index.js",
	)

	return {
		referenceLanguage: "en",
		plugins: [
			jsonPlugin({ pathPattern: "./languages/{language}.json" }),
			sdkPlugin({
				languageNegotiation: {
					strategies: [{ type: "localStorage" }, { type: "navigator" }],
				},
			}),
			ideExtensionPlugin({
				/* config */
			}),
		],
	}
}

```

PR #619, opened 2023-05-02
refactor: add ide extension plugin

felixhaeberle:
this refactors the ide extension functionality out of inlang/core and into a plugin. ref #618


felixhaeberle:
I've tested this with the sdk-js/adapter-sveltekit/examples/test-app/inlang.config.js with the following config:

```js
import ideExtensionPlugin from "@inlang/ide-extension/plugin"

/**
 * @type { import("@inlang/core/config").DefineConfig }
 */
export async function defineConfig(env) {
	const { default: jsonPlugin } = await env.$import(
		"https://cdn.jsdelivr.net/gh/samuelstroschein/inlang-plugin-json@2/dist/index.js"
	)
	const { default: sdkPlugin } = await env.$import(
		"https://cdn.jsdelivr.net/npm/@inlang/sdk-js@0.0.4/dist/plugin/index.js"
	)
	// const { default: ideExtensionPlugin } = await env.$import(
	// 	"https://cdn.jsdelivr.net/npm/@inlang/ide-extension@x.x.x/dist/plugin/index.js"
	// )

	return {
		referenceLanguage: "en",
		plugins: [
			jsonPlugin({ pathPattern: "./languages/{language}.json" }),
			sdkPlugin({
				languageNegotiation: {
					strategies: [{ type: "localStorage" }, { type: "navigator" }]
				}
			}),
			ideExtensionPlugin({
				messageReferenceMatchers: [
				  async ( /** @type {{ "documentText": string; }} */ args) => {
					const regex = /(?<!\w){?t\(['"](?<messageId>\S+)['"]\)}?/gm;
					const str = args.documentText;
					let match;
					const result = [];
		  
					while ((match = regex.exec(str)) !== null) {
					  const startLine =
						(str.slice(0, Math.max(0, match.index)).match(/\n/g) || [])
						.length + 1;
					  const startPos =
						match.index - str.lastIndexOf("\n", match.index - 1);
					  const endPos =
						match.index +
						match[0].length -
						str.lastIndexOf("\n", match.index + match[0].length - 1);
					  const endLine =
						(
						  str
						  .slice(0, Math.max(0, match.index + match[0].length))
						  .match(/\n/g) || []
						).length + 1;
		  
					  if (match.groups && "messageId" in match.groups) {
						result.push({
						  messageId: match.groups["messageId"],
						  position: {
							start: {
							  line: startLine,
							  character: startPos,
							},
							end: {
							  line: endLine,
							  character: endPos,
							},
						  },
						});
					  }
					}
					return result;
				  },
				],
				extractMessageOptions: [{
					callback: (messageId) => `{i("${messageId}")}`,
				  }
				],
			  })

		],
	}
}

```

felixhaeberle:
@ivanhofer As @samuelstroschein has limited availability, I tagged you as reviewer. 🙏😇

felixhaeberle:
@samuelstroschein @ivanhofer For bringing this to the example repo (and releasing it) – we have to publish a npm package for the ide extension too.

ivanhofer:
@felixhaeberle we need a public url where someone can access the bundled file.
Either build it and commit it in the inlang repository, release it as an npm package, or include it in a repository that can create a "Release" works. I would prefer the `npm` package.
What do you think @samuelstroschein?

source-code/ide-extension/src/plugin/schema.ts
```diff
@@ -0,0 +1,67 @@
...
+
+const ideExtensionSchema = z
+	.object({
+		/**
+		 * Defines matchers for message references inside the code.
+		 *
+		 * @param args represents the data to conduct the search on
+		 * @returns a promise with matched message references
+		 */
+		messageReferenceMatchers: z.array(
```
ivanhofer, line 33: this seems wrong. Previously it was a function that returned an array

source-code/ide-extension/src/plugin/schema.ts
```diff
@@ -0,0 +1,67 @@
...
+		 * An array of VSCode DocumentSelectors.
+		 *
+		 * The document selectors specify for which files/programming languages
+		 * (typescript, svelte, etc.) the extension should be activated.
+		 *
+		 * See https://code.visualstudio.com/api/references/document-selector
+		 */
+		// documentSelectors: DocumentSelector[];
+	})
+	.optional()
```
ivanhofer, line 65: I would mark it as required, because it doe not make sense to use the plugin and that plugin then does not define the property

ivanhofer (changes requested):
I have fixed a few smaller things. Please take a look at the `zodObject`.
It would also be great to have tests that fail if we pass a wrong config

felixhaeberle:
@ivanhofer Thank you, will do the changes. 🙏

source-code/core/src/config/schema.ts
```diff
@@ -78,54 +78,6 @@ export type InlangConfig = {
...
-				/**
-				 * The position from where to where the reference can be found.
-				 */
-				position: {
-					start: { line: number; character: number }
-					end: { line: number; character: number }
-				}
-			}>
-		>)[]
-
```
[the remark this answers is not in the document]
ivanhofer: I never have worked with `z.function` before, but it seems correct. If you add some tests we will know it for sure ;)

felixhaeberle:
@ivanhofer I added tests. but I have to check because currently while writing tests I have some strange type errors.
the zod type doesn't resolve correctly. can you reproduce this?

source-code/ide-extension/tsup.config.ts
```diff
@@ -0,0 +1,16 @@
+import { defineConfig } from "tsup"
```
samuelstroschein, line 1: I would use pluginBuildConfig from  @inlang/core/plugin to stay compliant.

source-code/ide-extension/package.json
```diff
@@ -19,7 +19,11 @@
 		"Linters",
 		"Visualization"
 	],
-	"main": "./dist/main.cjs",
+	"exports": {
+		".": "./dist/ide/main.cjs",
+		"./plugin": "./dist/plugin/index.js"
+	},
```
samuelstroschein, line 25: no exports are needed and might confuse vscode (we are not using npm for plugins and vscode only uses main)

source-code/ide-extension/src/plugin/plugin.ts
```diff
@@ -0,0 +1,16 @@
+import { type IdeExtensionConfig, validateIdeExtensionConfig } from "./schema.js"
+import type { InlangConfig } from "@inlang/core/config"
+import { createPlugin } from "@inlang/core/plugin"
+
+// ------------------------------------------------------------------------------------------------
+
+export const ideExtensionPlugin = createPlugin<IdeExtensionConfig>(({ settings }) => ({
```
samuelstroschein, line 7: hey, you told me to not use `pluginConfig` and use `settings` instead :D I would expect `IdeExtensionConfig` to be named `IdeExtensionSettings` ;)

source-code/ide-extension/tsup.config.ts
felixhaeberle: @samuelstroschein 😄 You missed to export the types correctly from package.json, therefore the import couldn't be found and we didn't use it "in that way already" in the codebase, you imported it from the file directly for the tests. I will fix that tomorrow. 🧼

source-code/ide-extension/package.json
```diff
@@ -28,9 +32,23 @@
 			}
 		]
 	},
+	"typesVersions": {
+		"*": {
+			"plugin": [
+				"./dist/plugin/index.d.ts"
+			]
+		}
+	},
```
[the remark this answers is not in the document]
felixhaeberle: @samuelstroschein How do we publish the ide-extension plugin?

source-code/ide-extension/tsup.config.ts
samuelstroschein: The types are exported correctly for node16 module resolution^^

I guess the ide extension (as well as sveltekit) don't use node16 module resolution. All of this module discrepancy is such a pain.

source-code/ide-extension/package.json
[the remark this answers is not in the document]
samuelstroschein: Just as we publish other plugins, directly via github. 

But, other plugins should build on top of the ide extension plugin. Thus, publishing to npm might be a good idea after all. We can't use `@inlang/ide-extension` as a namespace though because vscode doesn't support `@organization` syntax. 

Might be easier to have two separate package.json files/projects for the ide extension and the ide extension plugin? Should also ease bundling and typing things as the plugin-template can be used. 

```
source-code/ide-extension-plugin [@inlang/plugin-ide-extension] -> this would need to be imported by developers
source-code/ide-extension [vs-code-extension]
```

source-code/ide-extension/tsup.config.ts
ivanhofer: I can take a look at this. I think I now have seen all edge cases when it comes to exporting types for all possible formats ^^

ivanhofer (commented):
@felixhaeberle I have fixed the failing test for you. Regarding the type issue, I assume you already fixed it yourself, right?
Then the only thing that is missing is to release the plugin as a npm package.

Once the nom package exists, you can fix the failign tests. Currently we can't specify the types, because they don't get exported anywhere.

felixhaeberle:
@ivanhofer thank you!! will have a look on it tomorrow morning ☺️🙏

source-code/ide-extension/package.json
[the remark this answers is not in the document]
felixhaeberle: @samuelstroschein or even it's own repository like the other plugins / standard lint rules etc? 

then it's really clear that this is some standalone plugin code. 

published under the inlang organization (not my account).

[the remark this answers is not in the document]
felixhaeberle: @samuelstroschein off topic here but regarding the ide extension marketing in the marketplace I would suggest something more common like translation intellisense or something with -"lens", 
-> inlang - i18n translation intellisense, not only "inlang"

we can A/B test this once it's out.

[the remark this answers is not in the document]
felixhaeberle: @samuelstroschein I have moved it into it's own folder for now. can be further extracted.

source-code/ide-extension/src/plugin/plugin.ts
felixhaeberle: @samuelstroschein Done.

felixhaeberle:
@ivanhofer I moved the plugin into it's own folder for an npm release. (@inlang/ide-extension-plugin)
-> source-code/ide-extension-plugin
How can I release it from the command line, or should I use github. I remember we had a command line tool for that?
  
Currently the build / tests are broken by the`sdk-js` workspace, do you know the cause? 
Has it something to do with the missing release?

felixhaeberle:
@ivanhofer After the tests are fixed for the `sdk-js`, this should be merged in order that I can continue preparing the example repo for @samuelstroschein.

ivanhofer (approved):
I made some changes and think we can release it now

PR #623, opened 2023-05-02
Validate config schema with zod

stepanorda:


ivanhofer:
Thanks for the PR. I made some minor improvements in a commit. I have also commented out some types, we don't support for now.

I also have added some default values, but `zod` seems to have issues with it.
What we want is to allow to pass

```ts
{
	strict: false
}
```
and
```ts
{ }
```
and when calling `parse` in both cases the result should be
```ts
{
	strict: false
}
```
But if we use `.default(false)`, the type requires us to pass `strict` anyways and if we use `.default(false).optional()`, then the default value does not get applied to the `strict` property.
Can you please investigate if there is a possibility to have the behavior described above?

stepanorda:
@ivanhofer accomplished it by splitting zod-generated TS into 2 interfaces. Input and output.

ivanhofer:
> @ivanhofer accomplished it by splitting zod-generated TS into 2 interfaces. Input and output.

Cool solution! How did you come up with it? Did you find something on their docs?

stepanorda:
@ivanhofer just googled how to do it :) It's not really described in docs, just briefly mentioned that you can have infare, input and output.

ivanhofer (approved):
Thanks again for your help!

PR #636, opened 2023-05-05
Merge changes from inlang-git-fs into main

araknast:
A few changes I made to `memoryFilesystem` whilst working on migrating the isomorphic-git tests and compatibility.

The first commit adds extends `FilesystemError` to also include the name of the command that produced the error, in line with node's FS implementation. I had originally implemented for my own personal use but I found it to be useful enough to warrant including in the actual implementation.

The second commit fixes a bug where calling `fs.mkdir('/')` would create a directory with `undefined` as it's key instead of a string. This was original due in part to a non-null assertion in `getBasename`, which has been replaced with a nullish coalesce to `path`. This also brings the behavior of `getBasename` and `getDirname` in line with their POSIX counterparts, now returning the path instead of an empty string when a basename is given. "/" has also been added to `specialPaths` since we can no longer count on it being filtered out by `getBasename` and `getDirname`.

### DIFF
--- a/source-code-git/fs/src/errors/FilesystemError.ts
+++ b/source-code-git/fs/src/errors/FilesystemError.ts
@@ -2,23 +2,23 @@ export class FilesystemError extends Error {
 	code: string
 	path: string
 
-	constructor(code: string, path: string) {
+	constructor(code: string, path: string, syscall: string) {
 		let message
 		switch (code) {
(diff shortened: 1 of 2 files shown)


araknast:
Ended up rewriting `memoryFs` anyways, going to merge it with inlang-isomorphic-git instead.

Issue #666, opened 2023-05-08
IDE extension improvements


felixhaeberle:
- [x] https://github.com/inlang/inlang/issues/348
- [x] https://github.com/inlang/inlang/issues/383
- [x] https://github.com/inlang/inlang/issues/451
- [x] https://github.com/inlang/inlang/issues/633

felixhaeberle:
this can be closed in favor of #688 – work continues there 🚧

PR #688, opened 2023-05-09
Ide extension improvements

felixhaeberle:
hi, this PR adds the functionality from #666.
this should be merged & tested with the new release of `ide-extension-plugin` then.

### DIFF
--- a/source-code/ide-extension/src/commands/extractMessage.ts
+++ b/source-code/ide-extension/src/commands/extractMessage.ts
@@ -1,3 +1,4 @@
+import { telemetryNode } from "./../../../telemetry/src/implementation/node.js"
 import * as vscode from "vscode"
 import { setState, state } from "../state.js"
 import { query } from "@inlang/core/query"
@@ -94,6 +95,13 @@ export const extractMessageCommand = {
 		}
 		await textEditor.edit((editor) => {
 			editor.replace(textEditor.selection, preparedExtractOption)
+		})
+		telemetryNode.capture({
+			distinctId: "unknown",
+			event: "VS CODE EXTENSION message extracted",
+			properties: {
+				config: state().config,
+			},
 		})
 		return msg("Message extracted.")
 	},
--- a/source-code/ide-extension/src/commands/extractMessage.ts
+++ b/source-code/ide-extension/src/commands/extractMessage.ts
@@ -2,6 +2,7 @@ import * as vscode from "vscode"
 import { setState, state } from "../state.js"
 import { query } from "@inlang/core/query"
 import type { Message } from "@inlang/core/ast"
+import { msg } from "../utils/message.js"
 
 /**
  * Helps the user to extract messages from the active text editor.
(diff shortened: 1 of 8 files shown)


source-code/ide-extension/src/commands/extractMessage.ts
```diff
@@ -14,17 +16,26 @@ export const extractMessageCommand = {
...
+			return msg(
 				"There is no `ideExtension` object in the inlang.config.json configured.",
+				"warn",
+				"notification",
 			)
 		}
 		if (ideExtension.extractMessageOptions === undefined) {
-			return vscode.window.showWarningMessage(
-				"The `extractMessageReplacementOptions` are not defined in the inlang.config.json but required to extract a message.",
-			)
```
samuelstroschein, line 24: I would keep this in and not set a default. The default will break, leading to frustrated developers. A concise error message on why the IDE Extension doesn't work seems much better.

samuelstroschein (changes requested):
Minor changes requested. Else go and merge this. 

FYI haven't tested the ide extension because you mentioned some commonjs problems that need to be fixed

source-code/ide-extension/src/commands/extractMessage.ts
felixhaeberle: @samuelstroschein So people have to paste in a huge config when using the inlang extension when there's no plugin (like it is now?) and no sdk usage?

samuelstroschein: > So people have to paste in a huge config when using the inlang extension 

Yes. 

Don't you think the DX is better when the ide extension nudges a user to install a plugin that works rather than pretending to work but not work? The latter leads to "this extension is buggy and crap".

source-code/ide-extension/src/main.ts
```diff
@@ -96,12 +114,15 @@ async function main(args: { context: vscode.ExtensionContext }): Promise<void> {
 	args.context.subscriptions.push(
 		vscode.languages.registerCodeActionsProvider(
 			[
-				// TODO: improve with #348
```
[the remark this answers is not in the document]
felixhaeberle: @samuelstroschein The TODO was because of the activation event of the extension, not the code action. We also need a smart default here, as simply "every file" (which is also possible) is too broad. Should probably also be set by ide-extension-plugin and could be overridden by the user.

felixhaeberle:
@samuelstroschein extension works now that you can test it again. but still telemetry isn't working. I don't know why. So strange! Spend the whole morning figuring this out

felixhaeberle:
@NiklasBuchfink & @NilsJacobsen I tagged you as reviewer because it's best when a lot of people test it in their environment first. Please use the @inlang/example for this and see where you stumble over things. 

Please read the README of the IDE extension to set you up.

samuelstroschein:
@felixhaeberle 

1. I can't get the ide extension working with the example. I assume the reason to be the removal of default settings. I tried to use the ide extension config 

2. The telemetry issue is resolved. I replaced rollup with esbuild which solved bundling issues along with other benefits. See commit bebd1a6e490ddc1c83cd1945b1736c5184c74fe2

---

PS The nudge below is great UX

CleanShot 2023-05-10 at 16 53 03@2x

felixhaeberle:
@samuelstroschein I just looked at the code and everything works so far, but it doesn't track any events in my case. Also when @NilsJacobsen started the extension on his machine, no events were being registered. Do we run it in a wrong way?

felixhaeberle:
Which environment should I use? I chose the `monorepo` and then `development` but doesnt work. Now trying to use `development_internal`.

felixhaeberle:
@NilsJacobsen & @NiklasBuchfink can you test it again with the instructions above?

NilsJacobsen:
@felixhaeberle give me a sec

samuelstroschein:
@felixhaeberle merge this?

felixhaeberle:
@samuelstroschein will do!

PR #713, opened 2023-05-10
705: fix getFunctionOrDeclarationValue and test

### LINKED ISSUE #705: Multiple exports with the same name `handle`
The issue occurs when someone uses the `sequence` helper function.

Example repo: https://github.com/jmandzik/inlang-repro/

Code:
```ts
import { sequence } from "@sveltejs/kit/hooks";

const seq1 = async ({ event, resolve }) => {
  return resolve(event);
};

export const handle = sequence(seq1);
```

We need to improve the AST detection.

benjaminpreiss:
closes #705


benjaminpreiss:
@ivanhofer you want to do the honours?

PR #759, opened 2023-05-15
Force symlinks to be absolute

### LINKED ISSUE #753: Fix `clone.test.ts` – can't find file `test-walk/folder/3.txt/`
## Problem

A clone test seems to be broken and can't find a file.

```sh
@inlang-git/client:test:  ✓ src/raw/tests/GitRefManager.test.ts  (4 tests) 33ms
@inlang-git/client:test: stdout | src/raw/tests/clone.test.ts > clone > should set up the remote tracking branch by default
@inlang-git/client:test: FilesystemError: ENOENT: No such file or directory, symlink '/src/raw/tests/fixtures/isomorphic-git/__tests__/__fixtures__/test-walk/folder/3.txt/'
@inlang-git/client:test:     at Object.symlink (file:///home/runner/work/inlang/inlang/source-code-git/fs/dist/implementations/memoryFs.js:132:23)
@inlang-git/client:test:     at FileSystem.writelink (file:///home/runner/work/inlang/inlang/source-code-git/client/node_modules/isomorphic-git/internal-apis.js:4841:17)
@inlang-git/client:test:     at /home/runner/work/inlang/inlang/source-code-git/client/node_modules/isomorphic-git/index.cjs:6146:28
@inlang-git/client:test:     at processTicksAndRejections (node:internal/process/task_queues:95:5)
@inlang-git/client:test:     at async Promise.all (index 1404)
@inlang-git/client:test:     at /home/runner/work/inlang/inlang/source-code-git/client/node_modules/isomorphic-git/index.cjs:6118:7
@inlang-git/client:test:     at /home/runner/work/inlang/inlang/source-code-git/client/node_modules/isomorphic-git/index.cjs:987:16 {
@inlang-git/client:test:   code: 'ENOENT',
@inlang-git/client:test:   path: '/src/raw/tests/fixtures/isomorphic-git/__tests__/__fixtures__/test-walk/folder/3.txt/',
@inlang-git/client:test:   syscall: 'symlink',
@inlang-git/client:test:   target: undefined
@inlang-git/client:test: }
@inlang-git/client:test: 
```

### Expected behavior

Test should find the file.

Maybe it's because the 3.txt with a slash behind it is registered as a folder and not file? 
Only a wild guess.

## Steps to reproduce

run `npm run test`.

araknast:
All paths in memoryFs should be absolute, but isomorphic-git's `walk` fixture includes a relative symlink that causes subtle errors. This change ensures that all symlinks targets that do not begin with `/` are normalized relative to the basedir of the symlink.

Fixes #753

### DIFF
--- a/source-code-git/fs/src/implementations/memoryFs.ts
+++ b/source-code-git/fs/src/implementations/memoryFs.ts
@@ -157,15 +157,17 @@ export function createMemoryFs(): NodeishFilesystem {
 			path: Parameters<NodeishFilesystem["symlink"]>[1],
 		) {
 			path = normalPath(path)
+			target = target.startsWith("/") ? target : `${path}/../${target}`
 			const targetInode: Inode | undefined = fsMap.get(normalPath(target))
 			const parentDir: Inode | undefined = fsMap.get(getDirname(path))
 
 			if (fsMap.get(path)) throw new FilesystemError("EEXIST", path, "symlink", target)
 
-			if (parentDir instanceof Uint8Array) throw new FilesystemError("ENOTDIR", path, "symlink")
+			if (parentDir instanceof Uint8Array)
+				throw new FilesystemError("ENOTDIR", path, "symlink", target)
 
 			if (targetInode === undefined || parentDir === undefined)
-				throw new FilesystemError("ENOENT", path, "symlink")
+				throw new FilesystemError("ENOENT", path, "symlink", target)
 
 			parentDir.add(getBasename(path))
 			newStatEntry(path, fsStats, 2, 0o777, target)
--- a/source-code-git/client/src/raw/tests/add.test.ts
+++ b/source-code-git/client/src/raw/tests/add.test.ts
@@ -129,7 +129,6 @@ describe("add", () => {
 		if (symlinkTargetStr.startsWith("./")) {
 			symlinkTargetStr = symlinkTargetStr.substr(2)
 		}
-		expect(symlinkTargetStr).toEqual("c/e.txt")
 	})
 	it("ignored file", async () => {
 		// Setup
--- a/source-code-git/client/src/raw/tests/fixtures/test-add/c/special-pattern.js
+++ b/source-code-git/client/src/raw/tests/fixtures/test-add/c/special-pattern.js
@@ -1,3 +1,3 @@
 module.exports = {
-	ignore: true,
+  ignore: true
 }
--- a/source-code-git/client/src/raw/tests/fixtures/test-add/js_modules/awesome/index.js
+++ b/source-code-git/client/src/raw/tests/fixtures/test-add/js_modules/awesome/index.js
@@ -1,5 +1,5 @@
 module.exports = () => {
-	return {
-		happiness: true,
-	}
+  return {
+    happiness: true
+  }
 }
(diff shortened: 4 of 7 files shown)


samuelstroschein:
@araknast Are you confident that forcing absolute paths is the right decision?

source-code-git/fs/src/implementations/memoryFs.ts
```diff
@@ -157,15 +157,17 @@ export function createMemoryFs(): NodeishFilesystem {
 			path: Parameters<NodeishFilesystem["symlink"]>[1],
 		) {
 			path = normalPath(path)
+			target = target.startsWith("/") ? target : `${path}/../${target}`
```
samuelstroschein, line 160: If we force absolute paths, wouldn't it make sense to use `assertIsAbsolutePath` and explicitly provide developers with a clear error message instead of implicitly fixing the path (that might lead to confusion because relative paths can sometimes be used and sometimes not)? 

Like in https://github.com/inlang/inlang/blob/3cc7b69936b393dfa6e3fc13a22b39e7734944ed/source-code-git/fs/src/utilities/fromJson.ts#L14

samuelstroschein (changes requested):

source-code-git/fs/src/implementations/memoryFs.ts
samuelstroschein: @araknast Are you confident that forcing absolute paths is the right decision?

EDIT: Absolute paths are less ambiguous for sure and cwd doesn't exist in the browser. Seems to me that trading off "works in every environment" for slight DX improvements for node (by not requiring absolute paths in node) is a good one.

araknast: > If we force absolute paths, wouldn't it make sense to use assertIsAbsolutePath and explicitly provide developers with a clear error message instead of implicitly fixing the path (that might lead to confusion because relative paths can sometimes be used and sometimes not)?

Ideally we would be able to use `assertIsAbsolutePath` to ensure all paths begin with `/` but I don't think it would be worth the trouble of making it work with isomorphic-git. There is also the issue of `mockEnvironment`, and the reason this issue exists in the first place in that symlinks on 'real' filesystems often use relative paths. Using `assertIsAbsolutePath` would mean that we could not create a mock environment with a directory that contains a relative symbolic link (the reason why the error in #753 was occurring.

> Are you confident that forcing absolute paths is the right decision?
>
> EDIT: Absolute paths are less ambiguous for sure and cwd doesn't exist in the browser. Seems to me that trading off "works in every environment" for slight DX improvements for node (by not requiring absolute paths in node) is a good one.

All paths in memoryFs are 'absolute' anyways since they are always relative to the root of the filesystem. Internally a '/' is prepended to the path as part of the `normalPath` function. The behavior is identical to node as if `cwd` was set to `/`. 

To be clear, I don't necessarily want to enforce use of absolute paths in the api (for the reasons outlined above), only to ensure all paths that are passed around internally in memoryFs are absolute. This was the intended behavior before, and this patch simply extends it to the `symlink` command.

samuelstroschein: @araknast Got it, makes sense!

samuelstroschein (approved):
@araknast feel free to merge!

Issue #760, opened 2023-05-15
CLI `inlang machine translate` command


samuelstroschein:
## Context

Users are requesting a `machine translate` command for the CLI that can be embedded in CI/CD. 

## Proposal 

Have a `machine translate` command that: 

1. calls `readResources` to get all resources
2. iterate over the resources and finds all messages that are missing (I propose to not rely on the missingMessages rule because not every project is 100% certain to use the missingMessages rule). 
3. machine translates all missing translations
4. calls `writeResources` to persist updated resources   

## Blocked by

- [x] #76 
- [x] #777 

## Tasks after complete

- [ ] #782

jonator:
@samuelstroschein will be waiting for a PR

samuelstroschein:
@jonator Yes, for now at least via our API keys (no need for you to set up anything)

jonator:
Tagging @maxmillington

jonator:
We also have a chaggpt API key we can pass instead of a google API key already.

Looking great @samuelstroschein ! Keep me posted

This will be me for now:
<img width="855" alt="Screenshot 2023-05-16 at 3 45 28 PM" src="https://github.com/inlang/inlang/assets/4606373/1d634f7b-bf8e-4013-a20f-f7f94030c1fb">

felixhaeberle:
@samuelstroschein Is there a mode needed where we "machine translate --hard" translate all messages, not respecting already translated keys? Could be with an additional prompt.

samuelstroschein:
@felixhaeberle please no `--hard` flag :D Replacing human reviewed (hopefully) translations with machine translation = bad.

felixhaeberle:
@samuelstroschein unfortunately, we have to fix the env setup in prod #777 😅

samuelstroschein:
Fixed #777. @jonator PR is on my todo list for tomorrow in #782

PR #766, opened 2023-05-16
chore: add first draft of machine translation command

felixhaeberle:
This is the pull request for the machine translation feature in #760 
It's very basic at the moment but overall the functionality is established.

*Exclaimer: This implementation is not ready, it's just there to already move ourselves into a feedback cycle and discuss further ideas where this could end and which functionality we want to provide through this command.*

At the moment, it looks like this:
- Printing out the used config file
- Print out the total languages and messages
- Print out to which language the CLI is currently translating to.
- Success message

Bildschirmfoto 2023-05-16 um 15 27 34


source-code/cli/src/commands/machine/translate.ts
```diff
@@ -0,0 +1,151 @@
...
+export const translate = new Command()
+	.command("translate")
+	.description("Machine translate all resources.")
+	.action(translateCommandAction)
+
+async function translateCommandAction() {
+	// Set up the environment functions
+	const env: InlangEnvironment = {
+		$import: initialize$import({
+			// @ts-ignore TODO: use @inlang-git/fs
```
samuelstroschein, line 19: @ts-ingore should not be required. Is this a copy & paste error?

source-code/cli/src/commands/machine/translate.ts
```diff
@@ -0,0 +1,151 @@
...
+			fs: fs.promises,
+			fetch,
+		}),
+		// @ts-ignore TODO: use @inlang-git/fs
```
samuelstroschein, line 23: @ts-ingore should not be required. Is this a copy & paste error?

source-code/cli/src/commands/machine/translate.ts
```diff
@@ -0,0 +1,151 @@
...
+	} else {
+		log.info("✅ Using inlang config file at `" + filePath + "`")
+	}
+
+	// Get the content of the inlang.config.js file
+	const file = fs.readFileSync(filePath, "utf-8")
+
+	const config = await setupConfig({
+		module: await import("data:application/javascript;base64," + btoa(file.toString())),
+		env,
```
samuelstroschein, line 41: You can directly import the config instead of a dynamic import (which is a workaround for the editor). 

```ts

const module = await import(filePath)
```

samuelstroschein (commented):
Can you check if `@ts-ignore` is required?

source-code/cli/src/commands/machine/translate.ts
felixhaeberle: works!

samuelstroschein:
Also don;t forget to add a changeset with `npx changeset`

source-code/cli/src/commands/machine/translate.ts
```diff
@@ -16,11 +18,11 @@ async function translateCommandAction() {
 	// Set up the environment functions
 	const env: InlangEnvironment = {
 		$import: initialize$import({
-			// @ts-ignore TODO: use @inlang-git/fs
+			// TODO: use @inlang-git/fs
```
samuelstroschein, line 21: You don't need to use `@inlang-git/fs`. `@inlang-git/fs` is for browser environments. The CLI operates in node

felixhaeberle:
@inlang/team 
This PR introduces the new `inlang machine translate` command. 
This translates all missing keys for all languages (except of the reference one -> could be an improvement)

Furthermore this adds the `inlang open editor` command too from #775 to further improve the DX and let people quickly jump to the editor from their project.

felixhaeberle:
@samuelstroschein Do you have more feedback?

source-code/cli/src/commands/machine/translate.ts
```diff
@@ -0,0 +1,168 @@
...
+		if (error) {
+			console.error("Failed to get the remote URL.", error.message)
+			return
+		}
+
+		const remoteUrl = stdout.trim()
+
+		telemetryNode.capture({
+			distinctId: "CLI",
+			event: "CLI: machine translate",
```
samuelstroschein, line 157: Let's name it `CLI command executed [machine translate]` to have a common way of identifying that a command has been executed?

felixhaeberle: done

samuelstroschein (approved):
lets go baby

PR #773, opened 2023-05-16
Don't use `--sparse` option when cloning isomorphic-git

araknast:
The version of git used for deployments is out of date and does not support the `--sparse` option to `git clone`.


samuelstroschein:
@araknast feel free to merge such "low impact" PRs yourself. no need for me to review. If the changes work, they work. If not, reverse the changes :D

PR #803, opened 2023-05-19
Landingpage redesign 05 23

NilsJacobsen:
Hey folks, 

implementation of the new landing page redesign is out.
Feedback is welcome.

Take a look at the wording pls. I'm not super sure about the pricing section. There may be a better title.
Thanks guys.

localhost_3000_ (1)


samuelstroschein:
1) I made a PR to improve the pricing wording https://github.com/inlang/inlang/pull/804.

### 2) The tagline for the config is unclear. 

CleanShot 2023-05-19 at 15 31 41@2x

I understand that you want to communicate that inlang is powered by a config but I assume that a user that lands on the landing page couldn't care less _if he doesn't know why he should care_. 

Word by word analyzation: 

```
config -> uh no i don't want a config
bridge -> i don't care about bridges (needs imagination, open for interpretation, ambiguous)
inlang's infrastructure -> why would I care about inlang's infrastructure? What even is inlang's infrastructure?
```

### Proposal 

> Multiple apps, endless plugins, one config. 

```
multiple apps -> hell yes, I want apps
endless plugins -> nice inlang has plugins. this will work for me! (completely missing in the current tagline)
one config -> wow so simple, i get multiple apps with only one config  
```

The proposed tagline "hooks" better because we don't try to explain how inlang works but what the users gets and what the user has to do [implicitly explaining inlang]. The proposal turns the config into a good thing: The user gets multiple apps with just one config!

NilsJacobsen:
I like your proposals. Gonna change it.

samuelstroschein:
@NilsJacobsen Felix (on vacation now) asked why the framework logos (Svelte, Python, etc.) are greyscale. To increase "wiedererkennungswert", making the logos colorful seems like a good change?

yesWecanAcess:
> @NilsJacobsen Felix (on vacation now) asked why the framework logos (Svelte, Python, etc.) are greyscale. To increase "wiedererkennungswert", making the logos colorful seems like a good change to increase "widererkennungswert"?

And it gives them a bit more pop.  I like the idea

jannesblobel:
> @NilsJacobsen Felix (on vacation now) asked why the framework logos (Svelte, Python, etc.) are greyscale. To increase "wiedererkennungswert", making the logos colorful seems like a good change to increase "widererkennungswert"?

And it gives them a bit more pop. I like the idea

NilsJacobsen:
So I would say the aligned colors definitely make it more calm and adds more hierarchy to it. I'm pretty sure adding the original colors will make each logo fight for attention. -> then inlang will not be the focus anymore. 

I have to admit that the "wiedererkennungswert" will grow, but yeah I guess these two standings are not compatible so we have to decide for one way. I think make them grey scale is a common thing.
 
https://vercel.com/
https://linear.app/
https://astro.build/
https://turbo.build/
to name a few

The UI should only showcase that you can use every tool not really a specific one. Cause otherwise we would need to make every possible integration visible.

jannesblobel:
screencapture-localhost-3000-2023-05-19-16_29_29
What are your thoughts?

samuelstroschein:
@jannesblobel it does provide a better wiedererkennungswert but also distracts a bit. 

What about using opacity for framework logos? Seems like a middle ground between wiedererkennungswert and hierarchy.

PR #842, opened 2023-05-25
Groups for posthog

jannesblobel:


source-code/telemetry/src/implementation/gitOriginCheck.ts
```diff
@@ -0,0 +1,8 @@
+export function originChecked(getOrigin) {
+	if (getOrigin.startsWith("https://" || getOrigin.endsWith(".git"))) {
+		const checkedOrigin = getOrigin
+		return checkedOrigin
+	} else {
+		return "wrong Origin"
+	}
+}
```
samuelstroschein, line 8: 1. name the function `validateOrigin`
2. place the function in `@inlang/telemetry` 
3. use the function in every app where the origin is used. 

It would probably be even better if you write a function `getOrigin` that is used in every inlang app, ensuring that the origin is always "get" in the same way.

source-code/website/src/pages/editor/@host/@owner/@repository/State.tsx
```diff
@@ -221,6 +222,11 @@ export function EditorStateProvider(props: { children: JSXElement }) {
 			const result = await cloneRepository(args)
 			// not blocking the execution by using the callback pattern
 			// the user does not need to wait for the response
+			// checks whether the gitOrigin corresponds to the pattern.
+			const gitOrigin = originChecked(await getGitOrigin(args))
+			telemetryBrowser.group("repository", gitOrigin, {
+				name: gitOrigin,
```
samuelstroschein, line 228: why does the name need to be set?

source-code/cli/src/main.ts
```diff
@@ -36,10 +36,11 @@ export const cli = new Command()
 	.hook("postAction", (command) => {
 		telemetry.capture({
 			distinctId: "unknown",
+			groups: { repository: checkedOrigin! },
 			event: `CLI command executed`,
 			properties: {
 				args: command.args.join(" "),
-				gitOrigin,
+				checkedOrigin,
```
samuelstroschein, line 43: we don't need the property here annymore because we identify the group

source-code/cli/src/main.ts
```diff
@@ -50,9 +51,10 @@ export const cli = new Command()
 
 telemetry.capture({
 	distinctId: "unknown",
+	groups: { repository: checkedOrigin! },
 	event: "CLI started",
 	properties: {
 		version,
-		gitOrigin,
+		checkedOrigin,
```
samuelstroschein, line 58: we don't need the property here annymore because we identify the group

source-code/website/src/pages/editor/@host/@owner/@repository/State.tsx
```diff
@@ -230,21 +236,21 @@ export function EditorStateProvider(props: { children: JSXElement }) {
 					telemetryBrowser.capture("EDITOR cloned repository", {
 						owner: args.routeParams.owner,
 						repository: args.routeParams.repository,
-						isPrivate: response.data.private,
+						type: response.data.private ? "isPrivate" : "isPublic",
```
samuelstroschein, line 239: I would only use "private" or "public" without the `is` prefix. The `is` prefix is used for boolean values, making it easier to identify that a variable named `isPrivate` is a boolean that "is" either true or false.

samuelstroschein (changes requested):
1. I would move `getOrigin` to the `@inlang/telemetry` service and use it in every app. That ensures that the origin is always "get" in the same way in every app and makes `originChecked` redundant. 
2. The `BADGE created` event should have a group property too.

source-code/website/src/pages/editor/@host/@owner/@repository/State.tsx
```diff
@@ -236,7 +236,7 @@ export function EditorStateProvider(props: { children: JSXElement }) {
 					telemetryBrowser.capture("EDITOR cloned repository", {
 						owner: args.routeParams.owner,
 						repository: args.routeParams.repository,
-						type: response.data.private ? "isPrivate" : "isPublic",
+						type: response.data.private ? "Private" : "Public",
```
samuelstroschein, line 239: minor: I would make the strings lowercase because "unknown" is lowercase. Really minor though

source-code/website/src/pages/editor/@host/@owner/@repository/State.tsx
jannesblobel: You must set both to create a new group/repository. 
Otherwise, Posthog will not create a new group.

samuelstroschein (approved):

Issue #835, opened 2023-05-23
don't require to commit every little change


samuelstroschein:
## Problem 

The editor requires every change in a message to be committed. This workflow slows down users, is annoying, and not even how git workflows "work".

@araknast remarked this problem in [using git more effectively in the editor](https://gist.github.com/araknast/d80f8b64b2777e42b49fd9e89e5a1b13#using-git-more-effectively-in-the-editor). 

<img width="686" alt="CleanShot 2023-05-23 at 10 08 02@2x" src="https://github.com/inlang/inlang/assets/35429197/e1d63af3-2b8c-4a49-99dd-ae2c8c4b8c25">

## Shannon's proposal

> "Everything is a [branch], and we have a lot of tools to work with [branches]."

In [using git more effectively in the editor](https://gist.github.com/araknast/d80f8b64b2777e42b49fd9e89e5a1b13#using-git-more-effectively-in-the-editor), @araknast  proposes to introduce a concept to inlang-git of everything being a branch to mitigate performance and the UX issue of being required to commit everything. 

**Pros**

- fewer concepts that seem to make git easier to use

**Cons**

- Mitigates but doesn't solve the root cause of the stated UX performance and "reason why everything is a commit" issues
- Higher risk of building the wrong thing because the proposed concept of everything is a branch doesn't stem from a requirement 
- The concept of everything being a branch needs to be explained. We'll get there eventually. But, I think it is too early to make drastic changes to git. 

## Samuel's counter-proposal 

Don't introduce a new concept. Leave git as is for now to avoid building the wrong thing. 

The UX issue of being required to commit everything can be solved by 1) changing the commit button to a save button and 2) have a dedicated commit button. A dedicated commit button resembles other git-based apps.

**Pros**

- Commits can have batched changes (solving the UX issue to 90%)
- Git concepts remain as is, for now, avoiding us to build the wrong thing
- Being forced to solve the root causes of issues. (change detection = we need semantic diffing; status matrix performance issues = we need a filesystem that communicates with inlang-git)

**Cons**

- Change detection doesn't work anymore. 

The reason why everything has been a commit was to show "you changed X messages" by using `statusMatrix`. This concept is gone. We can't use git anymore to detect Message related changes. Changes would by detected on a file/line level, not a Message level. To re-initiate message-based "change detection", we would need semantic diffing. Great news: Semantic diffing seems like a no-brainer quality-of-life feature for software engineering too. Fewer merge conflicts, and more fine-grained changes. 

[...]

PR #888, opened 2023-06-05
#835 Reduce commits

NiklasBuchfink:
Reduce commits for better performance (https://github.com/inlang/inlang/issues/835)

**Changes**:
- "commit" button into a local "save"
- "push" is now "commit & push"
- "commit counter" is now a "save counter"


NilsJacobsen (changes requested):
We need to handle all the changes, until you commit them not only the last saved one.

NilsJacobsen (approved):
Now it works, great improvement @NiklasBuchfink

PR #900, opened 2023-06-08
Fix internal issue 15 public private forx

jannesblobel:
add group properties


source-code/website/src/pages/editor/@host/@owner/@repository/State.tsx
```diff
@@ -236,10 +236,14 @@ export function EditorStateProvider(props: { children: JSXElement }) {
 					repo: args.routeParams.repository,
 				})
 				.then((response) => {
+					telemetryBrowser.group("repository", gitOrigin, {
+						visibility: response.data.private ? "Private" : "Public",
+						isFork: response.data.fork ? "Fork" : "isNotFork",
+						parentGitOrigin: response.data.parent?.git_url.replace("git://", ""),
```
samuelstroschein, line 242: why don't you use parseOrigin for the gitorigin?

You can plug the url into `parseOrigin({remotes: [response.data.parent.git_url]})`

jannesblobel: parseOrigin required remote="origin"  and the URL 
[{remote: 'origin', url: response.data.parent.git_url}]
and we also need to adjust the function.  response.data.parent.gut_url return git://github.com/inlang/example.git
We are not covering git:// in this function. 
That's the reason.

jannesblobel: It looks clearer and more readable if we do this with the split function.

samuelstroschein: then change `parseOrigin` to only take the origin aka `parseOrigin(origin)`. 

We should use parseOrigin to ensure that the origin is always identical.

jannesblobel:
```suggestion
							? parseOrigin({ remotes: [{ remote: "origin", url: response.data.parent.git_url }] })
```
If we rewrite the function, I have to handle the remote on a other place
```suggestion
							? parseOrigin({ remotes: [{ remote: "origin", url: response.data.parent.git_url }] })
```

jannesblobel: Therefore, I would set the "remote" like suggested,  if we want to use the same function

samuelstroschein: Okay sounds good.

PR #910, opened 2023-06-09
Improve-feedback-reaction

samuelstroschein:
## Problem

1. The feedback thing needs to be added manually to each site. 
2. The feedback is hard to read because the telemetry event maps emojis to some strings. 

## Solutions

1. Add the feedback component to the documentation layout 
2. use the emojis directly in posthog.

### DIFF
--- a/documentation/apps/cli.md
+++ b/documentation/apps/cli.md
@@ -7,5 +7,3 @@ description: This is the CLI description
 # {% $frontmatter.title %}
 
 This is a text about the CLI
-
-{% Feedback /%}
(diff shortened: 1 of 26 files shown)


samuelstroschein:
@NiklasBuchfink I would change the number of emojis to two "good" and "bad". It makes graph building in posthog easier because we can calculate a number of the good and bad ratio that indicates if a documentation site needs improvements. 

For example a simple good / bad calculation 

5 good, 23 bad = 0.21
1 good, 1 bad = 1
30 good, 4 bad = 7,5

What do you think?

PR #937, opened 2023-06-15
make i18next plugin work for cal.com

samuelstroschein:
## Goal 

Make the plugin work for cal.com. 

The setup config function throws an error. Upon investigation, the plugin was refactored to increase the quality. Squashing multiple other bugs along the way. 

## Changes

- Unbundled several monster functions to increase code comprehension.
- Increased performance by removing O(n^3) computations, see https://github.com/inlang/inlang/issues/902#issuecomment-1594592866. 
- Fixed redundant reads and writes.
- Fixed multiple smaller bugs

### DIFF
--- a/.changeset/eight-knives-hear.md
+++ b/.changeset/eight-knives-hear.md
@@ -1,0 +1,5 @@
+---
+"@inlang/plugin-i18next": minor
+---
+
+Substantial internal refactorings to increase the quality of the plugin.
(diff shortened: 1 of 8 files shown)


samuelstroschein:
Here is an example for the performance issues. 

The resource body gets iterated over 4x times instead of only once. 

image

samuelstroschein:
@NilsJacobsen feel free to look at the individual commits to figure out what I improved.

PR #959, opened 2023-06-19
Implement objectStoreFs to improve cloning in large repos

araknast:
### functional 
- recursive directory creation :heavy_check_mark: 
- store objects un-deflated when caching :heavy_check_mark: 
### aesthetic
- split `setOidRecursive` :heavy_check_mark: 
- refactor/replace `setOidRecusive :heavy_check_mark: 
- all functions as top level :heavy_check_mark: 
- move functions to separate files :heavy_check_mark: 
- proper tests :heavy_check_mark: 
- lint :heavy_check_mark: 
- prettier :heavy_check_mark: 
- remove miscellaneous "test" files :heavy_check_mark: 
- clean up branch history :heavy_check_mark: 
- ~~`inlang-git` -> `project-lisa`~~ do this later as a quick commit to avoid complex conflicts
- `client/raw` -> `client/git` (same as above)


araknast:
@samuelstroschein 
Good news: `objectStoreFs` has been integrated into the editor.

Bad news: loading takes the exact same amount of time.

Not sure what it is, but there is something causing significant slowdown in the browser. In a bench test using node, reading a file from objectStoreFs was instant, now it takes about the same time `checkout` does. My question is, how do we proceed? Do we:
1. Implement `remoteObjectStoreFs` to try and offset the performance hit by reducing network operations 
2. Implement a reactive filesystem in hopes that this is the cause of the performance issues
3. Write a hopefully faster filesystem using something like IndexedDB, or even host fs apis, then implement reactivity at the `objectStoreFs` level
4. something else

Obviously 1 & 2 should both be done anyway, so in that case the question is in what order, i.e. what should we focus on first.

samuelstroschein:
@araknast Let's write performance tests for project lisa.

I understand that we are now wondering where performance slowdowns are coming from and, more importantly, if we are improving. Hence, having a suite of performance tests e.g. `PASS: cloning test repo X took 4s. TARGET: 5s` should give us more guidance and confidence. 

1. Do you agree that we should have a suite of performance tests to provide us insights and confidence that we are improving the performance of project lisa?

> Obviously 1 & 2 should both be done anyway, so in that case the question is in what order, i.e. what should we focus on first.

The most performance gain is to be expected in lazy cloning/loading of files. Hence, I would focus on implementing lazy loading of files. I understood that objectstorefs makes lazy loading of files easier. 

2. Lazy loading of files next?

araknast:
> Do you agree that we should have a suite of performance tests to provide us insights and confidence that we are improving the performance of project lisa?

Yes, that would be good, one question is how do we test the editor in general?  Is this a case where something like selenium/nightmare would be necessary?

> The most performance gain is to be expected in lazy cloning/loading of files. Hence, I would focus on implementing lazy loading of files. I understood that objectstorefs makes lazy loading of files easier.

Yes, `objectStoreFs` should make it easier to implement lazy loading.  I'll do that then, and hopefully we see performance gains there.

samuelstroschein:
> Yes, that would be good, one question is how do we test the editor in general? Is this a case where something like selenium/nightmare would be necessary? 

Hmm I guess so. The question is whether project lisa needs to test the editor. If the bench tests for project lisa are passing, and the editor is slow, the editor has an issue that is likely not lisa related. 

Hence, end-to-end tests might not be required. Bench tests that lead to an investigation and one-off fix seem sufficient. 

> Yes, objectStoreFs should make it easier to implement lazy loading. I'll do that then, and hopefully we see performance gains there.

👍

araknast:
Closed, moved to: https://github.com/inlang/inlang/pull/1187

PR #965, opened 2023-06-20
fix: placeolder parsing creates empty Text elements

NilsJacobsen:
Added the filter because we do not need to parse anything that is empty.
```ts
const elements: ast.Pattern["elements"] = text
		.split(placeholder)
		.filter((element) => element !== "")
		.map((element) => {
			if (placeholder.test(element)) {
				return {
					type: "Placeholder",
					body: {
						type: "VariableReference",
						name: variableReferencePattern[1]
							? element.slice(
									variableReferencePattern[0].length,
									// negative index, removing the trailing pattern
									-variableReferencePattern[1].length,
							  )
							: element.slice(variableReferencePattern[0].length),
					},
				}
			} else {
				return {
					type: "Text",
					value: element,
				}
			}
		})
```

### DIFF
--- a/source-code/plugins/i18next/src/plugin.test.ts
+++ b/source-code/plugins/i18next/src/plugin.test.ts
@@ -157,9 +157,9 @@ it("should remember if a file has a new line at the end or not", async () => {
 	expect(file2).toBe(withoutNewLine)
 })
 
-it("should correctly identify placeholders", async () => {
+it("should correctly identify placeholders without adding Text elements around it", async () => {
 	const enResource = `{
-    "test": "Hello {{username}}"
+    "test": "{{username}}"
 }`
 
 	const env = await mockEnvironment({})
@@ -175,6 +175,6 @@ it("should correctly identify placeholders", async () => {
 	const resources = await config.readResources!({
 		config: config as any,
 	})
-	expect(resources[0]?.body[0]?.pattern?.elements[0]?.type).toBe("Text")
-	expect(resources[0]?.body[0]?.pattern?.elements[1]?.type).toBe("Placeholder")
+	expect(resources[0]?.body[0]?.pattern?.elements[0]?.type).toBe("Placeholder")
+	expect(resources[0]?.body[0]?.pattern?.elements[1]).toBe(undefined)
 })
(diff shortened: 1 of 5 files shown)


NilsJacobsen:
https://github.com/inlang/inlang/issues/955

source-code/plugins/i18next/src/plugin.test.ts
```diff
@@ -175,6 +175,6 @@ it("should correctly identify placeholders", async () => {
 	const resources = await config.readResources!({
 		config: config as any,
 	})
-	expect(resources[0]?.body[0]?.pattern?.elements[0]?.type).toBe("Text")
-	expect(resources[0]?.body[0]?.pattern?.elements[1]?.type).toBe("Placeholder")
+	expect(resources[0]?.body[0]?.pattern?.elements[0]?.type).toBe("Placeholder")
+	expect(resources[0]?.body[0]?.pattern?.elements[1]).toBe(undefined)
```
samuelstroschein, line 179: @NilsJacobsen you just broke the test. 

How do you know if `Hello {{placeholder}}` will return a text element now and a placeholder? You don't :D 

Instead, either/or: 

- create a new unit test testing in isolation what you want to test
- update the test without changing the message and test that pattern.elements[2] is undefined

samuelstroschein (changes requested):

source-code/plugins/i18next/src/plugin.test.ts
samuelstroschein: In other words, you added a new unit test by removing another unit test. But, both are relevant. Thus, you should create a new unit test

NilsJacobsen: Alright got it. Give me a sec.

NilsJacobsen: Added a second test for it.

samuelstroschein (approved):
Perfect. 

@NilsJacobsen For my information, is this kind of unit testing easier than creating examples?

samuelstroschein (changes requested):
Oh wait, the changes need to be copy & pasted to the json plugin too :D

NilsJacobsen:
I like it better 👍

NilsJacobsen:
On my way :D

NilsJacobsen:
Done ✅

samuelstroschein (approved):

samuelstroschein:
@NilsJacobsen you forgot to create a changeset. The plugins won't be published. Please create a changeset on main with the `npx changeset` command and create a patch for `plugin-json` and `plugin-i18next`

PR #979, opened 2023-06-21
chore: add github actions extensions

openscript:
My reason to add them is to write better Github Actions in a more comfortable way.


felixhaeberle (approved):
Everything fine!

PR #986, opened 2023-06-22
update project lisa design principles

samuelstroschein:
### Changes 

- simplified the design principles (no fluff, easy to understand)
- added git compatibility (must have; wasn't mentioned)


araknast (changes requested):
I think the "fluff" was fine actually. It gave a more concrete view of what the principles mean in practice. All that's needed is to add the 'Git compatible' principle and corresponding explanation.

samuelstroschein:
@araknast Okay, will change it tomorrow to add more "fluff"

araknast (approved):

Issue #976, opened 2023-06-21
Can't commit with `plugin-i18next`


NiklasBuchfink:
## Problem

If I change something in the editor and commit the translation:
1. The change is reverted back to the old copy
2. An empty commit is pushed to the remote repository

Related to: https://github.com/samuelstroschein/inlang-plugin-json/issues/26

As works for the latest version of the `inlang/plugin-json` I suspect that it is a "plugin-i18next" and not an editor.

### Expected behavior

After push:
1. The commit is created correctly with the changes
2. By pulling after this commit, the copy remains unchanged

## Steps to reproduce

1. Open editor with a project using i18next (https://inlang.com/editor/github.com/NiklasBuchfink/nextjs-i18next)
2. Make a change & save
3. Push and commit
4. Review commit in your fork

Working `plugin-json` repository:
https://inlang.com/editor/github.com/NiklasBuchfink/example

samuelstroschein:
Investigation pending. 

A test case has been added to the plugin in https://github.com/inlang/inlang/commit/9dbf79c6583d034eb2845d8ca782c51d739a316d which does not indicate that the problem is plugin related.

samuelstroschein:
@NiklasBuchfink The bug seems editor-state-related. 

A few indicators: 

1. `setupConfig` roundtrip tests pass, meaning serialization works as expected.
2. The bug occurs the moment push changes is pressed. (On that note, rename the button to commit & push?)

Can you investigate the editor state?

https://github.com/inlang/inlang/assets/35429197/447ff6a4-d2ca-4184-92dc-4b31de478b76

NiklasBuchfink:
I'm on it 👍

NiklasBuchfink:
@samuelstroschein Results of my investigation:

This issue also happens with the `plugin-json` for path patterns like `app/i18n/locales/{language}/*.json`.

Everything is correct till the `writeResources` function in the State.tsx is executed. The `resources` argument contains the changes.

After that the `pushChanges` function is triggered and the value `filesWithUncommittedChanges` after executing `raw.statusMatrix` is empty for the mentioned path pattern.

samuelstroschein:
Okay, gotta have to add unit tests for paths like `app/i18n/locales/{language}/*.json` tomorrow.

samuelstroschein:
@NiklasBuchfink as discussed on the call, would be great of you can fix this

NiklasBuchfink:
Closed with #987

Issue #966, opened 2023-06-20
`npx inlang lint` creates empty `undefined.json` files


NiklasBuchfink:
## Problem

Executing `npx inlang lint` creates an empty `undefined.json` file in each language folder with the content:

```js
{}
```

**Additional information:**
- pathPattern: `public/locales/{language}/*.json` (folder structure with * structure)
- Operating system: Windows 11

<img width="704" alt="image" src="https://github.com/inlang/inlang/assets/59048346/67815c8f-4b45-4e12-9546-7ce8d38d58fa">

### Expected behavior

When running the command, no empty json file should be created.

## Steps to reproduce

1. Setup `@inlang/cli` with a i18next project
2. Make shure it's a project with a pathPattern like `public/locales/{language}/*.json`
3. Execute `npx inlang lint`

felixhaeberle:
@NiklasBuchfink Thanks for submitting this bug!

felixhaeberle:
@samuelstroschein Does lint write at all? 😅 Maybe that's only a bug inside the CLI but most of the time, the CLI only uses functions exposed by inlang/core.

samuelstroschein:
@felixhaeberle `setupConfig` writes to the filesystem to make a roundtrip test. 

- we should probably use a memory fs for roundtrip testing
- something is wrong in a plugin then i guess

felixhaeberle:
@NiklasBuchfink Can you provide the **config file** and the output of `npx @inlang/cli config validate` for reproduction?

felixhaeberle:
Probably related to #976

samuelstroschein:
@felixhaeberle true, could be related to #976 

I'll investigate today

NiklasBuchfink:
@felixhaeberle 

> @NiklasBuchfink Can you provide the **config file** and the output of `npx @inlang/cli config validate` for reproduction?

Example repo: https://github.com/NiklasBuchfink/nextjs-i18next

**Config:**
```js
/**
 * @type { import("@inlang/core/config").DefineConfig }
 */
export async function defineConfig(env) {
  const { default: pluginJson } = await env.$import(
    "https://cdn.jsdelivr.net/npm/@inlang/plugin-json@3/dist/index.js"
  )

  const { default: standardLintRules } = await env.$import(
    "https://cdn.jsdelivr.net/npm/@inlang/plugin-standard-lint-rules@3/dist/index.js",
  );

  return {
    referenceLanguage: "en",
    plugins: [
[...]

PR #987, opened 2023-06-22
fix language folder support

NiklasBuchfink:
This should tackle #966 & #976 for `plugin-i18next` and `plugin-json`

### DIFF
--- a/.changeset/clever-coats-hang.md
+++ b/.changeset/clever-coats-hang.md
@@ -1,0 +1,6 @@
+---
+"@inlang/plugin-i18next": patch
+"@inlang/plugin-json": patch
+---
+
+support language folders and addLanguage button
--- a/package-lock.json
+++ b/package-lock.json
@@ -18317,7 +18317,7 @@
 		},
 		"source-code/cli": {
 			"name": "@inlang/cli",
-			"version": "0.11.3",
+			"version": "0.11.5",
 			"license": "Apache-2.0",
 			"bin": {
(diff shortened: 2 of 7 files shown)


NiklasBuchfink:
@samuelstroschein Metadata is deprecated and still used inside the plugin. As it was empty inside writeResources, the fileName is undefined and that's why there was always the file creation of  `undefined.json`.

Please review the changes and let me know, if the fix over `message.id.name!.split(".")[0])` instead of `message.metadata.fileName` is valid.

**Note:** The same fix still needs to be made within the `plugin-json`.

samuelstroschein (approved):
LGTM. 

Why did the other unit tests not catch the bug?

samuelstroschein:
@NiklasBuchfink feel free to mark the PR as ready for review and immediately merge :)

BUT don't forget to: 

1. update the json plugin too
2. create patch notes with `npx changeset`

NiklasBuchfink:
@samuelstroschein I tested it locally. 

#966 is fixed with the change.

#976 not:
- correct in the local state of the editor (writing to fs seems to be correct.)
- commit is still empty (`filesWithUncommittedChanges` is still empty)

See https://inlang.com/editor/github.com/NiklasBuchfink/nextjs-i18next

I dig deeper

NiklasBuchfink:
Empty commits bug is fixed.

Last thing to be fixed: AddLanguage support. This is broken since the PR Reduce commits.

PR #988, opened 2023-06-22
Fix internal 72

jannesblobel:
Add group identifier to IDE and CLI


source-code/ide-extension/src/services/telemetry/implementation.ts
```diff
@@ -38,6 +38,15 @@ async function capture(args: CaptureEventArguments) {
...
 	}
+	if (args.event === "IDE-EXTENSION activated") {
+		telemetry.groupIdentify({
+			groupType: "repository",
+			groupKey: gitOrigin,
+			properties: {
+				name: gitOrigin,
+			},
+		})
+	}
```
samuelstroschein, line 49: Same here. I would implement this directly after https://github.com/inlang/inlang/blob/84116139da02753ba3c69864ae439a3e43fc84e6/source-code/ide-extension/src/main.ts#L20-L27

samuelstroschein (changes requested):
Furthermore, don't forget to create a changeset with `npx changeset`. This should be a patch note

source-code/ide-extension/src/services/telemetry/implementation.ts
jannesblobel: That was my first approach, but I had to import GitOrigin in this case. I assume that it is not so pretty to import the GitOrign here. 
But i will look into it again :)

samuelstroschein: Ah fair. Looks cleaner regardless i would assume :)

samuelstroschein (approved):

samuelstroschein:
@jannesblobel good to merge. just run `npx changeset` and create a changeset please

jannesblobel:
@samuelstroschein  nur nochmal kurz um sicher zu gehen, dass das changeset auch richtig ist. 
habe auch noch ganz kurz den groupIdentify in die badge gepackt  nicht unbedingt notwendig aber sollte auch nicht schaden

samuelstroschein:
@jannesblobel sieht richtig aus (patch note)

PR #1003, opened 2023-06-27
Lisa: Update design principles

araknast:
Add the "Host Agnostic" principle, as well as some re-wording and re-organization of the document.


samuelstroschein (approved):
Good changes! 

I like that things are becoming more clear, manifested by design principles. 

cc @inlang/team because you should know the design principles of lisa

PR #1024, opened 2023-07-02
fix(ide-extension): crash without git origin

openscript:
Uses `groupKey` "without-git", when there is no git origin

### DIFF
--- a/source-code/ide-extension/src/main.ts
+++ b/source-code/ide-extension/src/main.ts
@@ -32,13 +32,15 @@ export async function activate(context: vscode.ExtensionContext): Promise<void> 
 			},
 		})
 		const gitOrigin = await getGitOrigin()
-		telemetry.groupIdentify({
-			groupType: "repository",
-			groupKey: gitOrigin,
-			properties: {
-				name: gitOrigin,
-			},
-		})
+		if (gitOrigin) {
+			telemetry.groupIdentify({
+				groupType: "repository",
+				groupKey: gitOrigin,
+				properties: {
+					name: gitOrigin,
+				},
+			})
+		}
 		msg("Inlang extension activated.", "info")
 		// start the ide extension
 		main({ context })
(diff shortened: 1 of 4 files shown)


source-code/ide-extension/src/services/telemetry/implementation.ts
```diff
@@ -60,14 +60,19 @@ async function capture(args: CaptureEventArguments) {
...
+			dir: await raw.findRoot({
+				fs,
+				filepath: workspaceRoot ?? process.cwd(),
+			}),
+		})
+		const gitOrigin = parseOrigin({ remotes })
+		return gitOrigin
+	} catch (e) {
+		console.warn(e)
+		return 'without-git'
```
samuelstroschein, line 76: return `undefined`

source-code/ide-extension/src/main.ts
```diff
@@ -31,18 +31,14 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
...
-			console.warn(error)
-		}
+		const gitOrigin = await getGitOrigin()
+		telemetry.groupIdentify({
+			groupType: "repository",
+			groupKey: gitOrigin,
+			properties: {
+				name: gitOrigin,
+			},
+		})
```
samuelstroschein, line 41: don't group identify if the origin is `undefined`. 

```ts

if (gitOrigin) {
telemetry.groupIdentify({
			groupType: "repository",
			groupKey: gitOrigin,
			properties: {
				name: gitOrigin,
			},
		})
}

File: source-code/ide-extension/src/services/telemetry/implementation.ts [hunk 3]
```diff
@@ -60,14 +60,19 @@ async function capture(args: CaptureEventArguments) {
...
-	return gitOrigin
+			dir: await raw.findRoot({
+				fs,
+				filepath: workspaceRoot ?? process.cwd(),
+			}),
+		})
+		const gitOrigin = parseOrigin({ remotes })
+		return gitOrigin
+	} catch (e) {
+		console.warn(e)
```
samuelstroschein, line 75: I would remove the console.warn. There is nothing "bad" going on if a directory has not git repo.

samuelstroschein (changes requested):
Seems like you went the complicated route which would lead to bad telemetry data. 

### Request 

Simply don't perform a group identify if the origin is undefined, and don't use a stub ("without-git").

felixhaeberle (approved):
lgtm 👍

PR #1046, opened 2023-07-09
feat(ide-extension): add context tooltip

openscript:
[Screencast from 2023-07-09 23-06-54.webm](https://github.com/inlang/inlang/assets/1105080/328724ce-0b18-47ce-b4c5-44b62f91fc8c)

### DIFF
--- a/source-code/ide-extension/src/commands/editMessage.ts
+++ b/source-code/ide-extension/src/commands/editMessage.ts
@@ -37,7 +37,7 @@ export const editMessageCommand = {
       },
     }
 
-    const [newResource, exception] = query(currentResource).update({ id: messageId, with: newMessage })
+    const [newResource, exception] = query(currentResource).upsert({ message: newMessage })
 
     if (exception) {
       return window.showErrorMessage("Couldn't update message. ", exception.message)
--- a/source-code/ide-extension/src/commands/editMessage.ts
+++ b/source-code/ide-extension/src/commands/editMessage.ts
@@ -1,7 +1,58 @@
+import { query } from '@inlang/core/query'
+import { setState, state } from '../state.js'
+import { msg } from '../utilities/message.js'
+import { EventEmitter, window } from 'vscode'
+import type { Message } from '@inlang/core/ast'
+
+const onDidEditMessageEmitter = new EventEmitter<void>()
+export const onDidEditMessage = onDidEditMessageEmitter.event
+
 export const editMessageCommand = {
   id: "inlang.editMessage",
   title: "Inlang: Edit Message",
-  callback: async function ({ messageId, resource }: { messageId?: string, resource?: string }) {
-    return undefined
+  callback: async function ({ messageId, resource }: { messageId: string, resource: string }) {
+    const { writeResources } = state().config
+    const currentResource = state().resources.find((r) => r.languageTag.name === resource)
+    if (!currentResource) {
+      return msg("Couldn't retrieve resource", "warn", "notification")
+    }
+
+    const message = query(currentResource).get({ id: messageId })
+
+    const newValue = await window.showInputBox({
+      title: "Enter new value:",
+      placeHolder: message?.pattern.elements[0]!.value as string
+    })
+
+    if (newValue === undefined) {
+      return
+    }
+
+    const newMessage: Message = {
+      type: "Message",
+      id: { type: "Identifier", name: messageId },
+      pattern: {
+        type: "Pattern",
+        elements: [{ type: "Text", value: newValue }],
+      },
+    }
+
+    const [newResource, exception] = query(currentResource).update({ id: messageId, with: newMessage })
+
+    if (exception) {
+      return window.showErrorMessage("Couldn't update message. ", exception.message)
+    }
+    const resources = state().resources.map((r) =>
+      r.languageTag.name === resource ? newResource : r,
+    )
+    await writeResources({
+      config: state().config,
+      resources,
+    })
+    setState({ ...state(), resources })
+
+    onDidEditMessageEmitter.fire()
+
+    return msg("Message updated.")
   }
 } as const
--- a/source-code/ide-extension/src/commands/editMessage.ts
+++ b/source-code/ide-extension/src/commands/editMessage.ts
@@ -1,0 +1,7 @@
+export const editMessageCommand = {
+  id: "inlang.editMessage",
+  title: "Inlang: Edit Message",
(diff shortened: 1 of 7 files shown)


samuelstroschein:
@felixhaeberle thanks for taking me off the review request. @openscript since felix is leading the ide extension, no need to put me in review :)

source-code/ide-extension/src/decorations/contextTooltip.ts
```diff
@@ -0,0 +1,43 @@
...
+    const args = encodeURIComponent(JSON.stringify([{ messageId: message.messageId, resource: r.languageTag.name }]));
+
+    return [...acc, {
+      language: r.languageTag.name,
+      message: m?.pattern.elements[0]!.value ? m.pattern.elements[0].value as string : MISSING_TRANSLATION_MESSAGE,
+      editCommand: Uri.parse(`command:inlang.editMessage?${args}`),
+      openInEditorCommand: Uri.parse(`command:inlang.openInEditor?${args}`)
+    }]
+  }, [])
+  const contextTable = `<table>${messages.map((m) => renderTranslationRow(m)).join('')}</table>`
```
felixhaeberle, line 35: @openscript Can we have the messages sorted alphabetically from top to bottom? 

It might be interesting to now also think about highlighting the reference language message, but as we already dispay this message in the inline decorator, I think this is sufficient for now.

source-code/ide-extension/src/decorations/messagePreview.ts
```diff
@@ -111,4 +113,7 @@ export async function messagePreview(args: { context: vscode.ExtensionContext })
 		undefined,
 		args.context.subscriptions,
 	)
+
+	// update decorations, when message was edited
+	onDidEditMessage(() => updateDecorations());
```
felixhaeberle, line 118: Nice! Not listenting to all of the files but only update if we have a change event 👍

source-code/ide-extension/src/commands/openInEditor.ts
```diff
@@ -0,0 +1,17 @@
+import { getGitOrigin } from '../services/telemetry/implementation.js'
+import * as vscode from "vscode"
+
+const EDITOR_BASE_PATH = 'https://inlang.com/editor/'
+
+export const openInEditorCommand = {
+  id: "inlang.openInEditor",
+  title: "Inlang: Open in editor",
+  callback: async function ({ messageId }: { messageId?: string }) {
+    // TODO: Probably the origin should be configurable via the config.
```
felixhaeberle, line 10: @openscript Nice, you included a comment for this to be configurable. Let's see how it envolves.

source-code/ide-extension/src/commands/editMessage.ts
```diff
@@ -0,0 +1,58 @@
...
+    const currentResource = state().resources.find((r) => r.languageTag.name === resource)
+    if (!currentResource) {
+      return msg("Couldn't retrieve resource", "warn", "notification")
+    }
+
+    const message = query(currentResource).get({ id: messageId })
+
+    const newValue = await window.showInputBox({
+      title: "Enter new value:",
+      placeHolder: message?.pattern.elements[0]!.value as string
```
felixhaeberle, line 24: @openscript Can we also set the exisiting value here and not only the placeholder? But only if a value exists ofc. This could help resolve typos in long strings and not have to type everything again.

source-code/ide-extension/src/commands/editMessage.ts
```diff
@@ -0,0 +1,58 @@
...
+    )
+    await writeResources({
+      config: state().config,
+      resources,
+    })
+    setState({ ...state(), resources })
+
+    onDidEditMessageEmitter.fire()
+
+    return msg("Message updated.")
```
felixhaeberle, line 56: @openscript I once created the message helper function but the API of it needs improvements in order that we pass an object (because of optional arguments) and do not rely on argument chaining. We should refactor this soon.

source-code/ide-extension/src/decorations/contextTooltip.ts
```diff
@@ -0,0 +1,43 @@
...
+  message: string
+  editCommand?: Uri
+  openInEditorCommand?: Uri
+}
+
+function renderTranslationRow(row: ContextTableRow) {
+  const messageListing = `<td><strong>${row.language}</strong></td><td>${row.message}</td>`
+  const editCommand = row.editCommand ? `<td><a href="${row.editCommand}">$(edit)</a></td>` : ''
+  const openInEditorCommand = row.openInEditorCommand ? `<td><a href="${row.openInEditorCommand}">$(link-external)</a></td>` : ''
+  return `<tr>${messageListing}${editCommand}${openInEditorCommand}</tr>`
```
felixhaeberle, line 19: @openscript Can we get a littlebit of whitespace in the table cells? 

For example, additional 2px on each side of the column and 4px on the outside of the table? I recognized that the window easily closes (that was a concern we decided to go with) and it could probably be a bit less jumpy if we increase the space in the tooltip by a littlebit.

source-code/ide-extension/src/commands/editMessage.ts
openscript: Sounds good!

source-code/ide-extension/src/decorations/contextTooltip.ts
openscript: Yes, that should be easy :)

source-code/ide-extension/src/commands/editMessage.ts
openscript: @felixhaeberle Can you create an issue for this and describe how you would refactor it?

felixhaeberle: @openscript Will do!

source-code/ide-extension/src/decorations/contextTooltip.ts
openscript: Apart from `&nbsp;` I was not able to take an influence on the padding. I think the styling options are limited (for a good reason in my opinion).

Before:
image

After:
image

The tooltip looks better to me, even though I consider `&nbsp;` as an ugly hack.

felixhaeberle: @openscript That makes it much better in my opinion! Let's stay with the hack for now and see how the tooltip envolves.

felixhaeberle:
lgtm 👍 @openscript If you think there is nothing left, let's merge 🎉

https://media.giphy.com/media/vMNoKKznOrUJi/giphy.gif

PR #1065, opened 2023-07-11
Fix internal #104

jannesblobel:
I Fixed the issue https://github.com/inlang/internal/issues/104

### DIFF
--- a/source-code/ide-extension/src/main.ts
+++ b/source-code/ide-extension/src/main.ts
@@ -34,8 +34,8 @@ export async function activate(context: vscode.ExtensionContext): Promise<void> 
 				$set: {
 					// this information are user based, it will help us to create cohorts and to get a understanding why some projects are exclude
 					"user-IDE-settings": {
-						"workspace-recommendation-setting": !(await isDisabledRecommendation()),
-						"auto-config-file-creation-setting": !(await isDisabledConfigFileCreation()),
+						workspaceRecommendationSettingIsEnabled: !(await isDisabledRecommendation()),
+						autoConfigRecommendationSettingIsEnabled: !(await isDisabledConfigFileCreation()),
 					},
 				},
 			},
--- a/source-code/ide-extension/src/main.ts
+++ b/source-code/ide-extension/src/main.ts
@@ -15,12 +15,12 @@ import { promptToReloadWindow } from "./utilities/promptToReload.js"
 import { coreUsedConfigEvent } from "@inlang/telemetry"
 import {
 	recommendation,
-	isDisabledRecommendation,
+	// isDisabledRecommendation,
 	isInWorkspaceRecommendation,
 } from "./utilities/recommendation.js"
 import {
 	createInlangConfigFile,
-	isDisabledConfigFileCreation,
+	// isDisabledConfigFileCreation,
 } from "./utilities/createInlangConfigFile.js"
 
 export async function activate(context: vscode.ExtensionContext): Promise<void> {
@@ -42,13 +42,6 @@ export async function activate(context: vscode.ExtensionContext): Promise<void> 
 			properties: {
 				vscode_version: vscode.version,
 				version: version,
-				$set: {
-					// this information are user based, it will help us to create cohorts and to get a understanding why some projects are exclude
-					"user-IDE-settings": {
-						workspaceRecommendationSettingIsEnabled: !(await isDisabledRecommendation()),
-						autoConfigRecommendationSettingIsEnabled: !(await isDisabledConfigFileCreation()),
-					},
-				},
 			},
 		})
 		msg("Inlang extension activated.", "info")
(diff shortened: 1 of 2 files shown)


source-code/ide-extension/src/main.ts
```diff
@@ -27,20 +31,15 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
 			properties: {
 				vscode_version: vscode.version,
 				version: version,
-				workspaceRecommendation: !(await isDisabledRecommendation()),
-				autoConfigFileCreation: !(await isDisabledConfigFileCreation()),
+				$set: {
+					// this information are user based, it will help us to create cohorts and to get a understanding why some projects are exclude
+					"user-IDE-settings": {
+						"workspace-recommendation-setting": !(await isDisabledRecommendation()),
```
samuelstroschein, line 37: use boolean property names here. 

`workspaceRecommendationSettingIsEnabled`: true | false  
`autoConfigRecommendationSettingIsEnabled`: true | false  

**Why?**

- the `is` indicates that the property is a boolean and thereby increases comprehension 
- i was confused why you negate the values with `!` until i realized that the properties should be interpreted as  "default truthy". this is ambiguous. You can take the ambiguity out by stating the default case `isEnabled`

source-code/ide-extension/src/main.ts
```diff
@@ -89,6 +88,18 @@ async function main(args: { context: vscode.ExtensionContext }): Promise<void> {
...
+	if (gitOrigin) {
+		telemetry.groupIdentify({
+			groupType: "repository",
+			groupKey: gitOrigin,
+			properties: {
+				name: gitOrigin,
+				"in-workspace-recommendation": await inWorkspacerecommendation({ workspaceFolder }),
+			},
+		})
+	}
```
samuelstroschein, line 102: Doesn't the  group identify has to come before `IDE-EXTENSION activated`? 

Otherwise, the ide extension event will always be sent without the group information wouldn't it?

source-code/ide-extension/src/utilities/recommendation.ts
```diff
@@ -9,14 +9,7 @@ import { getGitOrigin, telemetry } from "../services/telemetry/implementation.js
...
  */
-export const recommendation = async (args: {
-	workspaceFolder: vscode.WorkspaceFolder
-}): Promise<void> => {
-	// check if the showRecommendation setting is set to false
-	if (await isDisabledRecommendation()) {
-		return
-	}
-
+export async function inWorkspacerecommendation(args: { workspaceFolder: vscode.WorkspaceFolder }) {
```
samuelstroschein, line 12: 1. If the function returns a boolean, I would improve the naming with `isInWorkspaceRecommendation`. On that note, the function contains a typo. The `r` in `recommendation` should be capitalized. 

2. Make the return type explicit with `boolean | undefined`. I didn't see the `return` statement at the very end.

samuelstroschein (changes requested):

felixhaeberle:
@jannesblobel Please test those settings & behavior with a real project. 

We currently do not have tests in place & especially since testing for telemetry is hard, so please have an extra look to see if it works the way it's intended. We are heavily relying on the telemetry provided by the IDE extension, so be extra careful 😅

samuelstroschein:
Hey sorry @jannesblobel, how does this fix https://github.com/inlang/internal/issues/104? 

We don't care about the user settings? We care if a repo has the workspace recommendation or not. Completely unrelated to the "settings". 

Your proposal clearly says "check if the IDE is in the workspace Recommendation and phrase is like isInWorkspaceRecommendation." why do you make the same mistake again now?

samuelstroschein:
@jannesblobel no, issue #104 is about something unrelated to cohorts and whatnot. fix issue 104 please

source-code/ide-extension/src/main.ts
```diff
@@ -89,6 +92,18 @@ async function main(args: { context: vscode.ExtensionContext }): Promise<void> {
...
+	if (gitOrigin) {
+		telemetry.groupIdentify({
+			groupType: "repository",
+			groupKey: gitOrigin,
+			properties: {
+				name: gitOrigin,
+				isInWorkspaceRecommendation: await isInWorkspaceRecommendation({ workspaceFolder }),
+			},
+		})
+	}
```
samuelstroschein, line 106: Let's do the group identify after the config file has been loaded. 

- less false data for workspace recommendations (if a repo doesn't have a config, obviously the workspace recommendation doesn't exist)

**Todo**

1. Move group identify beneath config setup. 
2. Attach the property to a repo if the config exists (then we can better filter)?

samuelstroschein (changes requested):

source-code/ide-extension/src/main.ts
jannesblobel: I think we agreed that we do not want to move this function and add another  group identify if a config exist after the config load. ;)
CleanShot 2023-07-20 at 22 53 47@2x

samuelstroschein:
@jannesblobel let's close this PR. #1149 will change the whole startup of an app. In other words, this code here will become outdated. No need to spam Posthog for 5 days. 

Agree? Then close please.

samuelstroschein (approved):

PR #1102, opened 2023-07-14
[RFC] refactor core to be "message-first"

samuelstroschein:
@inlang/team Read and discuss the RFC https://github.com/inlang/inlang/pull/1102/files. Here is a 19:59 minute long [loom](https://www.loom.com/share/8ff46ee202374e52b376d4d029f10fe5?sid=61303cce-bc07-445b-85f3-21d564c53a2a) about the RFC. Grab popcorn 🍿

1. Does "message-first" data structure make sense?
2. Is a (massive) breaking change worth the upside?
3. Can we make/should we invest resources into making the transition smooth for existing users?

> inlang apps think in messages, users think in messages, and yet inlang apps have to think in resources that can be queried to get messages.
> 
> that's wrong

---

#### Off-topic

Who would have thought that @felixhaeberle was right with his comment in https://github.com/inlang/inlang/issues/1072#issuecomment-1632699004: 

CleanShot 2023-07-14 at 20 44 58@2x

### DIFF
--- a/rfcs/message-first-architecture/RFC.md
+++ b/rfcs/message-first-architecture/RFC.md
@@ -185,7 +185,7 @@ export const missingMessage = createLintRule({
 export const missingMessage = createLintRule({
 	id: "inlang.missingMessage",
 	message: ({ message, messages, config }) => {
-		if (message.languageTag === config.referenceLanguageTag) {
+		if (message.languageTag !== config.referenceLanguageTag) {
 			return
 		}
 		const result: LintReport[] = []
--- a/rfcs/message-first-architecture/RFC.md
+++ b/rfcs/message-first-architecture/RFC.md
@@ -14,6 +14,11 @@ Inlang core, and thereby inlang itself, is architected "resource-first". Among t
 - Complicated application logic because apps care about messages, not resources.
 - Potentially, technical debt when extending inlang (core) beyong messages to other types of content (e.g. markdown, images, etc.).
 
+### Wrong assumptions with core's "resource-first" architecture
+
+- Resources are the primary data structure of inlang.
+- Plugins need exact knowledge of AST which turned out to be false. Every plugin, regardless of the format, treats resources as key-value data structure (e.g. JSON, YAML, etc.).
+
 ## Proposal
 
 Inlang core should be refactored to be "message-first".
@@ -75,6 +80,10 @@ messages = messages.delete({
 ```ts
 await config.saveMessages(messages)
 ```
+
+### Plugins
+
+- nothing changes for plugins except that they need to expose `saveMessages()` and `getMessages()`.
 
 ### Linting
 
--- a/rfcs/message-first-architecture/RFC.md
+++ b/rfcs/message-first-architecture/RFC.md
@@ -1,0 +1,179 @@
+## Discussion angle
+
+1. Does "message-first" data structure make sense?
+2. Is a (massive) breaking change worth the upside?
+3. Can we make/should we invest resources into making the transition smooth for existing users?
+
+Here is a 19:59 minute loom https://www.loom.com/share/8ff46ee202374e52b376d4d029f10fe5 about this RFC.
+
+## Problem
+
+Inlang core, and thereby inlang itself, is architected "resource-first". Among things, "resource-first" leads to:
+
+- Design flaws in inlang core [https://github.com/inlang/inlang/issues/1096, https://github.com/inlang/inlang/issues/945].
+- Complicated application logic because apps care about messages, not resources.
+- Potentially, technical debt when extending inlang (core) beyong messages to other types of content (e.g. markdown, images, etc.).
+
+## Proposal
+
+Inlang core should be refactored to be "message-first".
+
+- Simplifies the application logic of inlang core and inlang apps.
+- Simplify communication of inlang because the concept of "resources" is abstracted away/only plugins care about resources.
+- Future-proofs inlang. Messages can be retrieved from anywhere (e.g. resources files, remote APIs, embedded databases).
+
+### Application flow
+
+> Don't take the API design too seriously. It's a sketch to illustrate the simplification.
+
+- The general application flow is vastly simplified: No more resources, query, etc. Just `messages`.
+- The `messages` object can be performance optimized internally.
+
+#### 1. `config.getMessages()`
+
+- Plugins can retrieve messages from anywhere (e.g. resources files, remote APIs, embedded databases).
+- Future optimization: Only retrieve messages that are actually used in the app by passing a filter object.
+
+```ts
+const messages = await config.getMessages()
+```
+
+#### 2. CRUD operations
+
+- No more `resources` or `query` objects
+- Way simpler API.
+
+```ts
+messages = messages.create({
+	id: "hello",
+	languageTag: "en-US",
+	body: "how are you?",
+})
+
+const message = messages.get({
+	id: "hello",
+	languageTag: "en-US",
+	body: "how are you?",
+})
+
+messages = messages.update({
+	id: "hello",
+	languageTag: "en-US",
+	body: "how are you doing?",
+})
+
+messages = messages.delete({
+	id: "hello",
+	languageTag: "en-US",
+})
+```
+
+#### 3. `config.saveMessages()`
+
+- A plugin can save messages anywhere (e.g. resources files, remote APIs, embedded databases).
+
+```ts
+await config.saveMessages(messages)
+```
+
+### Linting
+
+- No more resources.
+  -- Lint "messages" makes much more sense than linting "resources".
+  -- Linting "messages" is what our lint system is about.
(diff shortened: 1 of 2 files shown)


samuelstroschein:
FYI I feel confident about the proposal. I just don't know how/if we should ease the breaking changes.

rfcs/message-first-architecture/RFC.md
```diff
@@ -0,0 +1,211 @@
...
+	id: "hello",
+	languageTag: "en-US",
+})
+```
+
+```ts
+// For comparison, here is the current API:
+
+const refResource = resources.find((resource) => resource.languageTag.name === "en-US")
+const message = query(refResource).get({
```
ivanhofer, line 89: To be fair this API could also be simplified to the example below ;)

rfcs/message-first-architecture/RFC.md
```diff
@@ -0,0 +1,211 @@
...
+		}
+	},
+})
+```
+
+##### Proposal
+
+> Note: This is a sketch. The API design should foster a discussion about the general direction.
+
+- No more `Resource` nodes. Just `Message` linting.
```
ivanhofer, line 184: What about parts inside a `Message` like `Variables` and `PluralRules`?

rfcs/message-first-architecture/RFC.md
```diff
@@ -0,0 +1,211 @@
...
+- No more `target` and `reference` nodes. Just a `message` and all the `messages`.
+- No more `"skip"` concept because performance is optimized via the `messages` object.
```
ivanhofer, line 186: If it is a flat structure, then it makes sense. If it's nested then we should keep it.

rfcs/message-first-architecture/RFC.md
```diff
@@ -0,0 +1,211 @@
...
+	},
+})
+```
+
+##### Proposal
+
+> Note: This is a sketch. The API design should foster a discussion about the general direction.
+
+- No more `Resource` nodes. Just `Message` linting.
+- No more `target` and `reference` nodes. Just a `message` and all the `messages`.
```
ivanhofer, line 185: It is harder to query the matching `Variable` if you need to find the message first.

rfcs/message-first-architecture/RFC.md
```diff
@@ -0,0 +1,211 @@
...
+					// because the message is missing.
+					targetLanguage = target?.languageTag.name
+				},
+				Message: ({ target, reference }) => {
+					if (target === undefined && reference) {
+						report({
+							node: reference,
+							message: `Message with id '${reference.id.name}' is missing for '${targetLanguage}'.`,
+						})
+						return {
```
ivanhofer, line 167: The return is currently not needed

rfcs/message-first-architecture/RFC.md
```diff
@@ -0,0 +1,211 @@
...
+- No more `Resource` nodes. Just `Message` linting.
+- No more `target` and `reference` nodes. Just a `message` and all the `messages`.
+- No more `"skip"` concept because performance is optimized via the `messages` object.
+- The mental model is simple: "I have a message and I can query all the messages to derive is something is wrong"
+
+```ts
+export const missingMessage = createLintRule({
+	id: "inlang.missingMessage",
+	message: ({ message, messages, config }) => {
+		if (message.languageTag === config.referenceLanguageTag) {
```
ivanhofer, line 193:
```suggestion
		if (message.languageTag !== config.referenceLanguageTag) {
```

You want to iterate over the messages of the referece language for this lint rule

ivanhofer (commented):
Getting rid of the concept of `Resources` makes sense. If we take a look at how an ideal implementation looks like, then we will have core primitives like `Message`, `Video`, `Text` etc.
I think we don't really need to find a concept to group those primitives.

`Resources` are the way how currently everything get's handled, but we are limiting ourselfs by sticking to legacy concepts.

I had a similar thought a few weeks ago when I was thinking about optimizing the data the SDK sends over the network to the clients (would be namespaces in current i18n concepts), but didn't spent time investigating further and left it for the future.

rfcs/message-first-architecture/RFC.md
samuelstroschein: A `Message` will have a pattern (or rather body which sounds more intuitive). Thus, not much changes. 

A lint rule needs to iterate over the pattern "manually" instead of a linter providing a pattern visitor. The difference is minor. I argue that manually iterating over the pattern is better because it's plain JS and therefore easier to understand.

rfcs/message-first-architecture/RFC.md
samuelstroschein: Flat structure. 

Nesting messages seems to stem from thinking in "resource-first" aka structering JSON files more "beautiful". A concept that is redundant if our ecosystem progresses. Using the IDE-Extension, Editor, etc. will be more convenient to modify messages and thereby resource files.

rfcs/message-first-architecture/RFC.md
samuelstroschein: Slightly but the target and reference concept breaks apart sooner or later. 

1. Sometimes both target and reference doesn't exist ("message not used in source code")
2. If we introduce variants for Messages (A/B testing), multiple references and targets exist. 

In both cases, it seems to be simpler to put the slight burden on rule writers to query the message(s) that want to lint directly.

rfcs/message-first-architecture/RFC.md
samuelstroschein: My bad, shouldn;'t be there. I fix it

rfcs/message-first-architecture/RFC.md
samuelstroschein: True but with more effort. messages would need to deal with resources under the hood and yikes everything becomes more complicated :D

samuelstroschein:
> @ivanhofer  If we take a look at how an ideal implementation looks like, then we will have core primitives like `Message`, `Video`, `Text` etc.

Exactly! 

Resources are just 'storage'. From an app standpoint, we care about Messages, Video, Audio, Documents, etc. Btw at this point it becomes clear that we are building a global-first CMS with extreme flexibility.

felixhaeberle:
For the editor I would say yes, but are we competing with something like Sanity then, where internationalization is just a feature? 

https://github.com/sanity-io/document-internationalization

We also build 1 an automation layer for globalized content namely the CLI and 2 Dev tools for the process of globalizing, which is a bit more loosely connected to the editor vision.

samuelstroschein:
> @felixhaeberle  For the editor I would say yes, but are we competing with something like Sanity then, where internationalization is just a feature? 

Application CMS, not "generic" CMS. 

inlang is globalization infrastructure.
inlang messages is a global-first application cms.

whatever. let's stick to globalization infrastructure for now. once we build more products, we eventually have to position inlang messages as dedicated product of that infrastructure ecosystem

samuelstroschein:
@inlang/team hey everyone, please discuss this issue today. 

this RFC blocks a lot of other issues. I am available for spontenous calls

NilsJacobsen:
## Opinions

> 1. Does "message-first" data structure make sense?

I agree. 
- Like the split (inlang doesn't care about the storage)
- In the editor we do exactly that message first structure before we can use it 
```ts
	const messages = createMemo(() => {
		const result: {
			[id: string]: {
				[language: Language]: LintedMessage | undefined
			}
		} = {}
		for (const resource of resources) {
			for (const id of query(resource).includedMessageIds()) {
				// defining the initial object, otherwise the following assignment will fail
				// with "cannot set property 'x' of undefined"
				if (result[id] === undefined) {
					result[id] = {}
				}
				const message = query(resource).get({ id }) as LintedMessage
				result[id]![resource.languageTag.name] = message
			}
		}
		return result
	})
```
- message.get just seams to be the better DX.

> 2. Is a (massive) breaking change worth the upside? 3. Can we make/should we invest resources into making the transition smooth for existing users?

I think we should make a good plan on how to do it. Losing our big repos would be very bad. But the most changes will be on our side right? We need tp prevent the weird situation where people use old plugins with the new core.

> Namespaces & Variants

- namespaces are doomed 😈 , but we need to support it -> let's not put it in core, let's keep it in the plugins and depricate it when most of the users are coming through sdks where we do not use namespaces.
- I wouldn't add variant for now. Could be nice in the future, but I don't see any project pushing for it.
- I hardly push for not adding any kind of metadata,namespaces ... in the past. You are right. It will always be a CRUD pain.

samuelstroschein:
Please vote with 👍 or 👎. 

If we agree that we should conduct a breaking change, the implications are: 

- focus this week, potentially into next week, on refactoring (feature/commit freeze)
- we can do other clean-ups at the same time (add `displayName` to lint rules, languageTag enforcements, etc.) 
- a good opportunity for everyone to get their hands dirty with `core`

If the proposal is accepted, I'll draft an issue (or a project?) with tasks that need to be conducted and invite everyone to add "drive-by refactorings" that can easily be conducted.

PR #1121, opened 2023-07-18
fix(ide-extension): display placeholders in message previews

openscript:


felixhaeberle (changes requested):
lgtm 👍 @openscript can you add a patch `npx changeset`?

PR #1084, opened 2023-07-13
docs: Adjustments and in-page navigation

floriandwt:
Based on https://github.com/inlang/internal/issues/90, I have added a in-page navigation to reach all the links.
List of updates:
- fixed bugs in anchor links
- added in-page navigation
- sticky navigation

PR #1136, opened 2023-07-19
fix: doc submenu rendering

floriandwt:
The headings of the newly added submenu weren't shown on the page (just when locally started).
Changed:
- Getting the headings now client-side from the markdown renderable tree
- As of the addition of https://inlang.com/documentation/sdk/sveltekit/advanced more bugs happened in the anchors, fixed them

@NilsJacobsen I'm assigning you because you already looked into #1084 and it's basically a fix for that.


NilsJacobsen (approved):
Lgtm. I tested it in production mode this time. SHould work. Feel free to merge.

PR #1142, opened 2023-07-19
Armageddon: Core refactoring 

samuelstroschein:
work in progress

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -9,7 +9,7 @@ jobs:
   ci:
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest-32gb, macos-latest]
+        os: [ubuntu-latest, macos-latest]
         version: [18]
     runs-on: ${{ matrix.os }}
     steps:


source-code/core/plugin/src/parsePlugin.ts
```diff
@@ -0,0 +1,86 @@
...
+	"inlang.plugin-standard-lint-rules",
+	"inlang.plugin-i18next",
+]
+
+export const parsePlugin = (args: {
+	config: InlangConfig
+	plugin: PluginApi
+	pluginInConfig: InlangConfig["plugins"][0]
+	result: ResolvePluginResult
+}) => {
```
samuelstroschein, line 29: @felixhaeberle I would not mix parsing with logic. 

```ts
function parsePlugin(maybeValidPlugin: Plugin): Result<Plugin, Error> {}
```

Hiding callbacks behind a setup function makes things complicated. Maybe we should go for the alternative `setup` pattern that we discussed: 

```ts
type Plugin = { 
  meta: Metadata,
  // called on resolving the plugin. can be used to initialize state/whatever. 
  setup: (args: { options: unknown } ) => void
  loadMessages(),
  saveMessages(),
  addAppSpecificApis(),
  // ...
  
}
```

@felixhaeberle what do you think?

samuelstroschein: Also makes `meta.usedApis` redundant because we know what APIs are used. In the future, we can re-add a meta.usedApis if we need to serialize the data.

felixhaeberle: yep, we should improve a few things there structure-wise.

samuelstroschein: > (because of multiple plugins in conjunction causing errors)

You are mixing parsing errors and resolution errors. 

`parsePlugin` = is a plugin valid
`resolvePlugins` = are all plugins in combination valid

felixhaeberle: oh, yes. we need to restructure the code there.

source-code/core/module/src/api.ts
```diff
@@ -34,8 +34,16 @@ export type ResolvedModulesFunction = (args: {
...
 		resolvedLintRules: LintRule[]
 	}
 	errors: Array<ModuleError | ModuleImportError>
 }>
+
+/**
+ * The API after resolving the modules.
+ */
+export type ResolvedModulesApi = {
+	resolvedPlugins: Awaited<ReturnType<ResolvePluginsFunction>>["data"]["plugins"]
```
samuelstroschein, line 47: @felixhaeberle why not return the return value of the resolve plugins function as we discussed?

source-code/core/module/src/resolveModules.ts
```diff
@@ -51,16 +49,27 @@ export const resolveModules: ResolvedModules = async (args) => {
...
+			 */
+
+			result.data.resolvedPlugins = {
+				...result.data.resolvedPlugins,
+				...resolvedPlugins.data.plugins,
+			}
+
+			result.data.resolvedLintRules = {
+				...result.data.resolvedLintRules,
+				...lintRules,
```
samuelstroschein, line 72: You can't spread here as the plugins and lint rules contain functions.

source-code/core/module/src/api.ts
samuelstroschein: What is `ResolvedModulesApi` for anyways? This is not an API?

source-code/core/module/src/resolveModules.ts
samuelstroschein: (the function would be lost aka this code doesn't work)

felixhaeberle: types & code are not ready, need to be improved - just fixed it for the build to pass

adding your comments to my list

source-code/plugins/i18next/src/plugin.test.ts
```diff
@@ -1,8 +1,9 @@
 import { expect, it, describe } from "vitest"
-import { plugin, type PluginOptions } from "./plugin.js"
-import { Variant, createMockEnvironment, getVariant } from "@inlang/plugin"
+import { plugin} from "./plugin.js"
+import type { PluginOptions } from "./options.js"
+import { type Variant, createMockEnvironment, getVariant } from "@inlang/plugin"
 
-describe("plugin options", () => {
+describe("option pathPattern", () => {
 	it("should throw if the path pattern does not include the {language} placeholder", async () => {
```
samuelstroschein, line 8: @inlang/plugins change `{language}` to `{languageTag}`?

NilsJacobsen: Done ✅

source-code/plugins/i18next/examples/example02/inlang.config.json
```diff
@@ -1,6 +1,6 @@
 {
 	"sourceLanguageTag": "en",
-	"languageTags": ["en"],
+	"languageTags": ["en", "de"],
 	"modules": ["./dist/index.js"],
 	"settings": {
 		"plugins": {
```
samuelstroschein: @inlang/plugins can we get rid of the "exampleXX" pattern? Testing what needs to be tested in unit tests should be possible. 

- nobody known what is tested if the example is called `example01`
- using regular unit test patterns make the test easier to follow

source-code/core/plugin/src/api.test-d.ts
```diff
@@ -25,26 +25,24 @@ const plugin1: Plugin<PluginOptions, AppSpecificApis> = {
...
-				messageReferenceMatcher: () => {},
+				messageReferenceMatcher: () => {
+					return
+				},
 			},
 			"inlang.cli": {
-				cliExtract: () => {},
+				cliExtract: () => {
+					return
+				},
```
samuelstroschein, line 45: the cli needs extract?

felixhaeberle: no whatever this is just a random test 😄

source-code/core/app/src/createInlang.ts
```diff
@@ -0,0 +1,205 @@
...
+
+/**
+ * Creates an inlang instance.
+ *
+ * - Use `_import` to pass a custom import function for testing,
+ *   and supporting legacy modules such as CJS.
+ *
+ */
+export const createInlang = async (args: {
+	configPath: string
```
samuelstroschein, line 28: @felixhaeberle @ivanhofer @NilsJacobsen @NiklasBuchfink

I think createInlang should figure out itself where the config is located and should only take the workingDirectory as argument. 

- apps wouldn't need own "where is the config logic" 
- all apps behave the same way/find the same config upon opening a repository (and defining the working directory) 

Agree? Vote pls.

felixhaeberle: Speaking about the CLI, how does it then work with self-defined --config paths in monorepo setups atm?

ivanhofer: We could make it optional and if set, the `resolveConfig` logic tries that file first

source-code/core/app/src/createInlang.ts
```diff
@@ -0,0 +1,194 @@
...
+		// -- messages ----------------------------------------------------------
+
+		const [messages, setMessages] = createSignal<Message[]>()
+		createEffect(() => {
+			console.log("0 effect load messages is triggered")
+			const _resolvedModules = resolvedModules()
+			if (!_resolvedModules) return
+			//console.log(_resolvedModules.data.plugins.data)
+			makeTrulyAsync(
+				_resolvedModules.data.plugins.data.loadMessages({ languageTags: config()!.languageTags }),
```
ivanhofer, line 70: should we wrap all plugin defined functions to return a `Result`?

samuelstroschein: How would an app fall back? Seems to me that an app should crash if loadMessages fails. But, also seems like we can change the behaviour in the future. We are not going to set @inlang/app to v1.0

source-code/core/app/src/createInlang.ts
samuelstroschein: > Speaking about the CLI, how does it then work with self-defined --config paths in monorepo setups atm? 

Set the working directory (which you can pass to createInlang) to the config path.

source-code/core/app/src/createInlang.ts
ivanhofer: We would add the error to the plugin errors array to make it more consistent. This would be the same as an `ErrorDuringLintError`.
`loadMessages` could throw an error on the 3rd call of that function if the API endpoint is temporary down. Should the editor crash then? I think the app should decide that.

samuelstroschein: > I think the app should decide that. 

Yeah, good point. 

@inlang/editor what do you think?

source-code/core/plugin/src/resolvePlugins.ts
```diff
@@ -0,0 +1,166 @@
...
+	PluginFunctionLoadMessagesAlreadyDefinedError,
+	PluginFunctionSaveMessagesAlreadyDefinedError,
+	PluginUsesInvalidIdError,
+	PluginUsesInvalidSchemaError,
+	PluginUsesReservedNamespaceError,
+} from "./errors.js"
+import { deepmerge } from "deepmerge-ts"
+import { TypeCompiler } from "@sinclair/typebox/compiler"
+
+const whitelistedPlugins = ["inlang.plugin-json", "inlang.plugin-i18next"]
```
ivanhofer, line 14: I don't think this is a smart idea ^^

source-code/core/app/src/createInlang.ts
NilsJacobsen: I think that feels more consistent. So that would be a task for plugins.

NiklasBuchfink: @NilsJacobsen and I agree that this is more consistent and is better handled within the app.

source-code/core/plugin/src/resolvePlugins.ts
samuelstroschein: How else would you reserve the `inlang` namespace?

ivanhofer: We will probably have more issues if we restrict the packages than if we trust the community to not use the inlang namespace.
If we always need to add a new plugin id to this array we
1. will probably forget it every second time and need to release a fix immediately afterwards
2. we require applications to always have the latest version of `@inlang/plugin` installed or else they won't be able to use the newly released plugin

source-code/core/config/src/schema.ts
```diff
@@ -0,0 +1,79 @@
...
+ *           └── ...
+ * ```
+ */
+export type PluginSettings = Static<typeof PluginSettings>
+export const PluginSettings = Type.Object({
+	options: Type.Optional(PluginOptions),
+})
+
+export type InlangConfig = Static<typeof InlangConfig>
+export const InlangConfig = Type.Object({
```
ivanhofer, line 60: we should probably invest some time into making error messages more human readable

ivanhofer: image

ivanhofer: Another one:
image

source-code/core/app/src/createInlang.ts
```diff
@@ -0,0 +1,215 @@
...
+				.then((messages) => {
+					setMessages(messages)
+					markInitAsComplete()
+				})
+				.catch((err) => {
+					console.error("Error in load messages ", err)
+				})
+		})
+
+		const query = createReactiveQuery(() => messages() ?? [])
```
ivanhofer, line 79: at this point of the execution `messages` will always be `undefined`

source-code/core/app/src/wrappers/withSolidReactivity.test.ts
```diff
@@ -0,0 +1,337 @@
...
+			})
+
+			createEffect(() => {
+				// 2 times because effect creation + set
+				inlang.meta.lintRules()
+				counterLint += 1
+			})
+
+			inlang.setConfig({ ...inlang.config(), languageTags: ["en", "fr"] })
+			setTimeout(() => {
```
ivanhofer, line 67: @NilsJacobsen FYI this was never awaited and therefore the tests didn't check anything.

source-code/cli/src/commands/machine/translate.ts
```diff
@@ -1,7 +1,7 @@
 // @ts-nocheck
 import { query } from "@inlang/core/query"
 import { Command } from "commander"
-import { countMessagesPerLanguage, getFlag, log } from "../../utilities.js"
+import { getFlag, log } from "../../utilities.js"
```
ivanhofer, line 4: please make sure that the build runs after refactoring code and before you push your code! I have disabled the `build` command for the CLI for now

samuelstroschein: sorry, moved old code before i ran off to dinner

source-code/plugins/json/src/settings.ts
```diff
@@ -1,47 +1,45 @@
...
+	ignore?: string[] // default is []
+}
+
 /**
- * Throws an error if the settings are invalid.
+ * Throws an error if the options are invalid.
  *
  * Not using zod becaue it's not worth the bundle size (2kb vs 14kb).
  */
 export function throwIfInvalidSettings(settings: PluginSettings) {
```
ivanhofer, line 13: I think it makes sense to use typebox here too

source-code/plugins/json/src/settings.ts
```diff
@@ -1,47 +1,45 @@
...
-	} else if (settings.pathPattern.includes("{language}") === false) {
-		throw new Error(
-			"The pathPattern setting must be defined and include the {language} placeholder. An example would be './resources/{language}.json'.",
-		)
-	} else if (settings.pathPattern.endsWith(".json") === false) {
-		throw new Error(
-			"The pathPattern setting must end with '.json'. An example would be './resources/{language}.json'.",
-		)
+	if (typeof settings.pathPattern === "string") {
+		if (settings.pathPattern.includes("{languageTag}") === false) {
```
ivanhofer, line 15: We should also throw an error if the pathPattern includes `{{languageTag}}` (with double brackets), since double brackets are the default for `variableReferencePattern` and they easily can be mixed up as it happened to me.

source-code/core/app/src/createInlang.ts
```diff
@@ -0,0 +1,220 @@
...
+		let configValue: InlangConfig
+		createEffect(() => configValue = config()!) // workaround to not run effects twice (e.g. config change + modules change) (I'm sure there exists a solid way of doing this, but I haven't found it yet)
+
+		const [messages, setMessages] = createSignal<Message[]>()
+		createEffect(() => {
+			const _resolvedModules = resolvedModules()
+			if (!_resolvedModules) return
+
+			makeTrulyAsync(
+				_resolvedModules.data.plugins.data.loadMessages({ languageTags: configValue!.languageTags }),
```
ivanhofer, line 73: should we output a warning if `loadMessages` does not return anything?

source-code/plugins/json/src/settings.ts
samuelstroschein: @ivanhofer you should ping @inlang/plugins if you want a reply ;)

source-code/plugins/json/src/settings.ts
samuelstroschein: Agree. cc @inlang/plugins

source-code/core/app/src/createInlang.ts
samuelstroschein: prob push an error to `errors` no?

source-code/core/app/src/createInlang.ts
```diff
@@ -0,0 +1,220 @@
...
+	]
+}
+
+// TODO: create global util type
+type MaybePromise<T> = T | Promise<T>
+
+const makeTrulyAsync = <T>(fn: MaybePromise<T>): Promise<T> => (async () => fn)()
+
+
+// TODO: how do we unsubscribe? Do we need that?
```
samuelstroschein, line 211: as i understand, no. we don't need to unsubscribe because the reactivity is handled within the inlang instance and is thereby auto tracked via scopes.

samuelstroschein: @ivanhofer i guess we will see if we run into memory leaks :D

source-code/plugins/json/src/settings.ts
NiklasBuchfink: Should double brackets or single brackets be the default for `inlang-plugin/json`? We can still change that. We didn't have a default before.

samuelstroschein: depends on what's most used

NiklasBuchfink: Let's use the single curly braces. They are the most common among JS, TS, Python and C# developers.

NiklasBuchfink: ✅ set `{variable}` set as default for `json`
✅ added error and test for `json` and `i18next`

source-code/plugins/json/src/settings.ts
NiklasBuchfink: I will add it 👍

source-code/core/messages/src/utilities.ts
```diff
@@ -0,0 +1,224 @@
...
+ *
+ * @example
+ * 	const variant = getVariant(message, { languageTag: "en", selectors: { gender: "male" }});
+ */
+export function getVariant(
+	message: Message,
+	options: {
+		languageTag: LanguageTag
+		selectors?: Record<string, string>
+	},
```
samuelstroschein, line 21: @NilsJacobsen this API is unexpected. I would have expected:

```ts
const variant = getVariant(message.body[languageTag], selectors: { gender: "male" })
```

- getting the language tag is straightforward via the body
- a variant is unrelated to the language tag (this api seems like a leaky design)
- variants only concern a specific message body

source-code/core/app/src/createInlang.ts
ivanhofer: It isn't really an error or is it?

samuelstroschein: Are you referring to empty array or undefined with "if does not return anything"? 

`loadMessages` must return an array of messages. If `undefined` is returned, I would consider the falsy return value a plugin error

ivanhofer: `undefined` => error
empty array => warning that the plugin is probably not set up correctly

source-code/plugins/json/src/settings.ts
ivanhofer: To be honest, I have troubles understanding the current validation logic. Some things get validated outside this function and something inside the function looks like code duplication.
It was just a suggestion, to have the same validatin pattern everywhere.

source-code/core/app/src/createInlang.ts
samuelstroschein: I see. I would not console.warn something (will be ignored). Apps should handle "hey, you don't have messages yet." as part of the app flow.

source-code/core/messages/src/utilities.ts
samuelstroschein: Also unexpected that the function returns a `Result`. The function can't crash. Either a variant exists or it doesn't (undefined). 

I will go ahead and change the API because it's obvious that we will change the API and I have to work with getVariant at the moment. Don't want to refactor it again. 

@NilsJacobsen use `Result` for failing code, not for code that can return `undefined`.

source-code/core/app/src/createReactiveQuery.ts
```diff
@@ -0,0 +1,48 @@
...
+		// TODO: update index
+		// q: replace all entries?
+		// q: just add new entries?
+		// q: remove old entries?
+		// setIndex(new ReactiveMap(messages().map((message) => [message.id, message])))
+	})
+
+	return {
+		create: ({ data }) => {
+			index.set(data.id, data)
```
ivanhofer, line 23: what do we want to do here if the id already exists? Currently we override the entry

source-code/core/messages/src/utilities.ts
samuelstroschein: NVM the variants need the message selectors to derive the ordering of the variants. Hence, passing the entire message is correct but unexpected. The return type of result is incorrect. I will change that.

source-code/core/app/src/createReactiveQuery.ts
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Creates a query API for messages.
+	@@ -9,41 +9,43 @@ import { createEffect } from "./solid.js"
+ */
+export function createReactiveQuery(messages: () => Array<Message>): MessageQueryApi {
+	//let index: ReactiveMap<string, Message>
+	const index = new ReactiveMap<string, Message>(messages().map((message) => [message.id, message]))
+
+	createEffect(() => {
+		// TODO: update index
```
ivanhofer, line 14: @inlang/team how should we update the map if new entries show up?

ivanhofer: - replace all entries?
- just add new entries?
- remove old entries?

source-code/core/app/src/createReactiveQuery.ts
samuelstroschein: Using `Result` for potentially failing queries like create, update?

source-code/core/app/src/createReactiveQuery.ts
samuelstroschein: > replace all entries? 

If it's fast enough :D We can optimize later

source-code/core/app/src/createReactiveQuery.ts
ivanhofer: so `create` should not override?

ivanhofer: and `delete` should "fail" if the id does not exist?

ivanhofer: Or maybe just return `true` or `false`?

ivanhofer: @inlang/team

samuelstroschein: I vote for `Result`

samuelstroschein: Ah you mean because the CRUD ops are `void`. Good point! Let's not leak `Result` into the ecosystem. Using a boolean that indicates whether an operation was successful seems indeed better!

- not leaking `Result` into the ecosystem -> better DX because "what is a result" is avoided
- boolean are easier to handle

samuelstroschein: Using a boolean is likely also good for `createVariant` and `updateVariantPattern`

ivanhofer: I did already implement it yesterday. I will add tests tomorrow.

source-code/core/app/inlang.config.json
```diff
@@ -0,0 +1 @@
+{"sourceLanguageTag":"en","languageTags":["en"],"modules":["./dist/index.js"],"settings":{"plugins":{"inlang.plugin.i18next":{"options":{"pathPattern":"./examples/example01/{languageTag}.json","variableReferencePattern":["{","}"]}}},"lintRules":{"inlang.missingMessage":{"level":"error"}}}}
```
ivanhofer, line 1: Why do we have this file here?

source-code/core/messages/src/query.test.ts
```diff
@@ -0,0 +1,115 @@
...
+		query.create({ data: mockMessage1 })
+		query.create({ data: mockMessage2 })
+
+		const messages = query.getAll()
+
+		expect(messages).toEqual([mockMessage1, mockMessage2])
+	})
+})
+
+describe("update", () => {
```
ivanhofer, line 62: @NiklasBuchfink what do you actually test here?
In the first test you check if the `get` after `update` returnes the passed object, but the description says something different.
In the other two tests you never call `update`.

source-code/core/app/src/createReactiveQuery.ts
ivanhofer: I wrote all tests for `query` but have nothing changed for variants yet.

> Using a boolean is likely also good for `createVariant` and `updateVariantPattern`

`updateVariantPattern` has multiple ways of failing. There returning `false` seems wrong.

source-code/core/app/inlang.config.json
samuelstroschein: Always wondered that. I think @NilsJacobsen used it for testing. Let's remove this file

source-code/core/messages/src/query.test.ts
NiklasBuchfink: @ivanhofer I got something mixed up there, but you have already corrected it. Thank you for pointing it out to me.

source-code/core/app/src/wrappers/withSolidReactivity.test.ts
```diff
@@ -0,0 +1,325 @@
+import { describe, it, expect } from "vitest"
+import { createInlang } from "../createInlang.js"
+import { createMockNodeishFs } from "@inlang/plugin/test"
```
ivanhofer, line 3: @samuelstroschein should this import not come from `@inlang/module/test` or `@inlang/app/test`?

ivanhofer: Ok, I'll re-export it and it should be fine. I just was confused that I need to import this form `plugin` inside the `sdk` app.

source-code/core/messages/src/variant.ts
```diff
@@ -0,0 +1,242 @@
...
+			selectors?: Variant["match"]
+		}
+	},
+): Variant | undefined {
+	const variant = matchMostSpecificVariant(message, args.where.languageTag, args.where.selectors)
+	if (variant) {
+		//! do not return a reference to the message in a resource
+		//! modifications to the returned message will leak into the
+		//! resource which is considered to be immutable.
+		return structuredClone(variant)
```
ivanhofer, line 33: Do we need to clone the object? We already do that when querying messages.

NilsJacobsen: I think we want the variant to me immutable right? The same way as we use the query utility functions in the code base before the refactoring

source-code/core/app/src/createInlang.ts
```diff
@@ -75,12 +75,20 @@ export const createInlang = async (args: {
 			if (!conf) return
 			loadModules({ config: conf, nodeishFs: args.nodeishFs, _import: args._import })
 				.then((resolvedModules) => {
+					if (resolvedModules.errors.length) {
+						throw new Error(resolvedModules.errors as any)
+					}
+					if (!resolvedModules.runtimePluginApi.loadMessages || !resolvedModules.runtimePluginApi.saveMessages) {
+						throw new Error("It seems you did not install any plugin that handles messages. Please add one to make inlang work.") // TODO: add link to docs
```
ivanhofer, line 83: @inlang/team do we agree to throw in these cases? If yes, I can add tests and a meaningful error description tomorrow.

source-code/core/json-serializable/src/schema.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * ---------------- UTILITIES ----------------
+ */
+
+import { Static, Type } from "@sinclair/typebox"
+
+const JSONValue = Type.Union([Type.String(), Type.Number(), Type.Boolean(), Type.Null()])
+const JSONArray = Type.Array(JSONValue)
+// avoiding recursive types in JSON object
+const NestedJSONObject = Type.Record(Type.String(), Type.Union([JSONValue, JSONArray]))
```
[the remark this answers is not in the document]
samuelstroschein: Can you point me to the error? All test are passing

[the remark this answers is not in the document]
ivanhofer: run `npm run dev` inside `source-code/sdk-js/src/adapter-sveltekit/examples/test-app`

source-code/core/app/src/createInlang.ts
samuelstroschein: I think we should never throw in `createInlang` and instead implement a mock readMessages, saveMessages, and push the error to `inlang.errors`: 

- more explicit error handling without a "gotcha" 
- apps can implement "create a new project" workflows easier by reacting to the pushed `inlang.errors` instead of handling try catch things.

source-code/core/json-serializable/src/schema.ts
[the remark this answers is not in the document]
ivanhofer: let's see if we run into issues with that implementation 😅

source-code/core/app/src/createInlang.ts
ivanhofer: I got confused by this line:
image

Yes, so we add a new error to the array 👍

ivanhofer: fixed, pushed but not tested yet

jannesblobel:
@samuelstroschein @ivanhofer 

What is the difference between these two lines? https://github.com/inlang/inlang/blob/91d088f1f9f77651957a422329b1b7b0e9d20e74/source-code/core/app/src/api.ts#L65-L66

ivanhofer:
> What is the difference between these two lines?

None, probably just a merge conflict

source-code/core/app/src/api.ts
```diff
@@ -90,11 +90,11 @@ export type Subscribable<Value> = {
 	subscribe: (callback: (value: Value) => void) => void
 }
 
-type MessageQueryApi = {
+export type MessageQueryApi = {
 	create: (args: { data: Message }) => boolean
 	get: (args: { where: { id: Message["id"] } }) => Message | undefined
```
samuelstroschein, line 95: Get needs to be subscribable

source-code/ide-extension/src/main.ts
```diff
@@ -20,6 +20,7 @@ import { openInEditorCommand } from "./commands/openInEditor.js"
 import { editMessageCommand } from "./commands/editMessage.js"
 import { openInlangProject, tryCatch } from "@inlang/app"
 import { createFileSystemMapper } from "./services/inlang-environment/src/createFileSystemMapper.js"
+import { _import } from "./services/inlang-environment/src/_import.js"
```
samuelstroschein, line 23: @felixhaeberle i would not classify `_import` as an (external) service. and neither does inlang-environment exist anymore

felixhaeberle: better?

source-code/core/app/src/api.ts
ivanhofer: I can take a look

source-code/core/module/src/errors.ts
```diff
@@ -0,0 +1,28 @@
+export class ModuleError extends Error {
+	public readonly module: string
+	constructor(message: string, options: { module: string; cause?: Error }) {
+		super(message)
+		this.name = "ModuleHasNoExportsError"
```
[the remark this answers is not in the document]
samuelstroschein: absolutely!

source-code/website/src/pages/editor/@host/@owner/@repository/index.page.tsx
```diff
@@ -33,6 +35,16 @@ function TheActualPage() {
...
+	const messages = createMemo(() => {
+		const result: {
+			[id: string]: MessageType
+		} = {}
+		for (const message of inlang()?.query.messages.getAll() || []) {
+			result[message.id] = message
+		}
+		return result
+	})
+
```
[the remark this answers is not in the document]
NilsJacobsen: Jap should be possible👍

source-code/plugins/json/src/plugin.ts
```diff
@@ -336,6 +336,7 @@ async function saveMessages(args: {
 		}
 		for (const [languageTag, value] of Object.entries(storage)) {
 			const pathWithLanguage = args.settings.pathPattern.replace("{languageTag}", languageTag)
+			await args.nodeishFs.mkdir(pathWithLanguage.split("/").slice(0, -1).join("/"), { recursive: true })
```
ivanhofer, line 339: I'm sure I already sayed it a few times:
**Plugins must work also if a file and folder does not exist yet!**
I fixed it quickly to be able to continue working on the SDK. @inlang/plugins please add a test and fix it in the other plugins too.

NilsJacobsen: I thought we do not want to have empty resource files any more. 
In the editor we look in the config (or add tag to it) to show editor then wenn u edited something u can push it with content. How is it for the SDK?

ivanhofer: > I thought we do not want to have empty resource files any more.

This has nothing to do with empty resource files..

In the SDK, if there are no messages yet, I will create some examples. And therefore `saveMessages` should not fail.

source-code/sdk-js/src/adapter-sveltekit/vite-plugin/config/config.ts
```diff
@@ -0,0 +1,158 @@
...
+		)
+
+		const svelteKitVersion = await getSvelteKitVersion()
+
+		resolve({
+			debug: settings.debug,
+
+			sourceLanguageTag: inlang.config().sourceLanguageTag,
+			languageTags: inlang.config().languageTags,
+			messages: () => Object.values(inlang.query.messages.getAll()),
```
NilsJacobsen, line 138: could you also use a combination of includedMessageIds and get here? We want to remove getAll. @ivanhofer

samuelstroschein: @NilsJacobsen should be doable. 

```ts
messages: () => inlang.query.message.includedIds().map((id) => inlang.query.get({ where: id } ))
```

lix/client/package.json
```diff
@@ -23,6 +23,7 @@
 	},
 	"dependencies": {
 		"@inlang-git/fs": "*",
+		"@inlang/app": "*",
```
samuelstroschein, line 26: @janfjohannes why does lix depent on inlang/app? 

If we want to share reactive primitives, lets extract the reactive primitives and adapters from inlang/app to lix/reactive-primitives or something. 

The logical education is that "inlang is a lix app". Not the other way around "lix is an inlang app"

samuelstroschein: Related to https://github.com/inlang/inlang/issues/1290

source-code/core/app/package.json
```diff
@@ -0,0 +1,53 @@
...
+		"---- LINT ----------------------------------------------------------": "",
+		"lint": "cd ../.. && npm run lint:base ./source-code/core",
+		"lint:fix": "cd ../.. && npm run lint:fix:base ./source-code/core",
+		"format": "cd ../.. && npm run format:base ./source-code/core",
+		"format:fix": "cd ../.. && npm run format:fix:base ./source-code/core",
+		"---- OTHER ---------------------------------------------------------": "",
+		"clean": "rm -rf ./dist ./.turbo ./coverage ./node_modules"
+	},
+	"engines": {
+		"node": ">=16.15.0"
```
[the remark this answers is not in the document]
samuelstroschein: Agree :)

source-code/badge/src/badge.ts
```diff
@@ -34,7 +34,7 @@ export const badge = async (url: string) => {
 	})
 
 	const inlang = await openInlangProject({
-		configPath: "./inlang.config.json",
+		configPath: "./project.inlang.json",
```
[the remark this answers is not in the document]
felixhaeberle: @samuelstroschein oh I still though we treat it as configuration file but give it a better name.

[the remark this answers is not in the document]
samuelstroschein: No, the concept of a config file is gone. We can name the path `projectFilePath`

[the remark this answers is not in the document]
felixhaeberle: @samuelstroschein Alright, then also rename `inlang.config()` & `inlang.setConfig()` to `inlang.project()` & `inlang.setProject()` ? 😂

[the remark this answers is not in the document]
felixhaeberle: > No, the concept of a config file is gone.
Just kidding.
Agree 👍 So keeping `config` & `setConfig`, but changing `configPath` > `projectFilePath`.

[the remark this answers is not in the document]
jannesblobel: Honestly, `project.config()` and `project.setConfig()` make so much more sense than `config` and `setConfig`

source-code/website/src/pages/marketplace/Select.tsx
```diff
@@ -106,3 +105,41 @@ export function SelectionWrapper(props: {
...
+			gitfloat?.classList.remove("animate-slideOut")
+			gitfloat?.classList.remove("opacity-0")
+			gitfloat?.classList.add("animate-slideIn")
+		} else {
+			gitfloat?.classList.add("animate-slideOut")
+			gitfloat?.classList.remove("animate-slideIn")
+			gitfloat?.classList.add("opacity-0")
+		}
+	})
+
```
samuelstroschein, line 129: @floriandwt ich verbiete dir effects! :D 

du kannst doch einfach auf dem gitfloat element dynamisch classen spezifizieren? https://www.solidjs.com/tutorial/bindings_classlist

floriandwt: Ich hatte die Logik einfach aus dem Editor kopiert und bin noch dran:
```
createEffect(() => {
		if (localChanges() > 0 && localStorage?.user !== undefined) {
			const gitfloat = document.querySelector(".gitfloat")
			gitfloat?.classList.add("animate-jump")
			setTimeout(() => {
				gitfloat?.classList.remove("animate-jump")
			}, 1000)
		}
	})
```

Werde es asap abändern, beim Editor ist es glaub wegen der localStorage so gesetzt :D Danke für den Link

source-code/website/src/pages/editor/@host/@owner/@repository/State.tsx
```diff
@@ -233,6 +234,19 @@ export function EditorStateProvider(props: { children: JSXElement }) {
...
+	createEffect(() => {
+		if (doesInlangConfigExist() === true) {
+			//Non-null assertion operator because gitOrigin/parseOrign always returns a string.
+			//Even if there is no origin, "unknown" is returned as a string.
+			telemetryBrowser.group("repository", gitOrigin()!, {
+				name: gitOrigin(),
+			})
+		}
+	})
+
```
samuelstroschein: @jannesblobel you don't need an effect. 

You can directly perform the group identify at the point where the repository is opened, in the createResource which calls `openRepository`. 

Your code likely also contains a bug. calling repo without a getter `repo()` is not reactive

source-code/paraglide/src/adapter-sveltekit/vite-plugin/config/config.ts
```diff
@@ -11,7 +11,7 @@ import { createBasicInlangConfig } from "./utils/createBasicInlangConfig.js"
 import { createDemoResourcesIfNoMessagesExistYet } from "./utils/createDemoResourcesIfNoMessagesExistYet.js"
 import { doesPathExist } from "./utils/utils.js"
 import { getNodeishFs } from "./utils/getNodeishFs.js"
-import { createImport } from "@inlang/module"
+import { createImport } from "../../../../../sdk/dist/resolve-modules/import.js"
```
samuelstroschein, line 14: @felixhaeberle @NiklasBuchfink wrong import path

Issue #1144, opened 2023-07-19
design flaws that should be addressed in core


samuelstroschein:
## Context

@inlang/team Participate in the referenced issues. I plan on doing a call tmr (20 July) to discuss things (if required). 

LOOM https://www.loom.com/share/4611de77b168423694fe53771ae6a49f 📹 

- during https://github.com/orgs/inlang/projects/35 more (highly likely) design flaws became apparent
- addressing those design flaws now is a unique opportunity because we can still manually update every config with PRs 

## Design flaws 

- [ ] https://github.com/inlang/inlang/issues/1143
- [ ] https://github.com/inlang/inlang/issues/1140

## question

Should we resolve the design flaws now at the risk of a big breaking change?

NilsJacobsen:
That's right, with the creation of the inlang instance, we can hide all that complex config stuff and work with an easy json. I also support that apps can change user settings.

My feeling: Prosal is better 👍 

My biggest concern is that this will result in a big braking change. 👎 
How can we make the transition smooth? Can we support both configs for some time? If we need maintainer input to make that work, I'm not sure if it's worth it losing superspreaders. (The best product will not help us in the next round, it's the users)

samuelstroschein:
> How can we make the transition smooth? Can we support both configs for some time? If we need maintainer input to make that work, I'm not sure if it's worth it losing superspreaders. (The best product will not help us in the next round, it's the users) 

The change is so big that I don't know if the effort is worth it. But: 

- the only requirement is that apps transform the config from JS to JSON. 
- for large repos, we can hardcode the config that an app should create based on their existing app :D
- for other repos, we can try to "auto migrate". not sure if it worth though

felixhaeberle:
We could write an codemod which transforms the js to a json, so this part would be "automated". 

We can even read the languages from the resource folder and add them to the json, because we know the resource path (and what should potentially be excluded).

But the codemod has to be executed manually. In each project. An appealing solution could be to introduce an temporary cli command `config codemod` which can be used by the maintainer and we tell this the maintainer via a setup error message in the new system.

samuelstroschein:
Let's create a dedicated issue on how we can migrate the config. @felixhaeberle could you do that? 

For this discussions, a very interesting question is: Do more design flaws exist that should be addressed?

NilsJacobsen:
[...]

Issue #1143, opened 2023-07-19
[design flaw] config as code is wrong


samuelstroschein:
## Problem 

The `js` config can't be modified by apps.

- users can't add languages in the editor 
- https://github.com/inlang/inlang/issues/1139
- the list goes on...

## Proposal 

Replace the JS config with a JSON config that can be modified by apps (similar to VSCode settings). 

- enables apps to be much more powerful 
- apps can work independently of developers (translator wants to add a plugin for example)
- only importing plugins is what user config do anyways, nobody uses the power of code now that plugins are available
- entails the requirement for #1140

```ts
{
  "referenceLanguageTag":"en",
  "languageTags":[
    "en",
    "de",
    "fr"
  ],
  "plugins":{
    "inlang.standardLintRules":{
      "module":"https://cdn.jsdelivr.net/npm/@inlang/plugin-standard-lint-rules@3/dist/index.js"
    },
    "inlang.plugin-json":{
      "module":"https://cdn.jsdelivr.net/npm/@inlang/plugin-json@3/dist/index.js",
      "options":{
        "pathPattern":"source-code/website/lang/{language}.json"
      }
    }
  },
  "lint":{
    "rules":{
      "inlang.missingMessage":"off"
    }
  }
}
```

samuelstroschein:
For plugins we could also make "where you host your plugin" easy: Any JSDelivr link works. NPM, GitHub, we don't care. 

In the future, we likely need our own registry.

samuelstroschein:
Also `languageTags` must always be provided. Plugins have an API that doesn't mix with config. For legacy purposes, plugins can have a `extendLanguageTags` callback that should phase out over time. 

We want app users to have the power to add languageTags and messages as they desire.

NilsJacobsen:
- In the future, we should be able to change settings without editing code. There should be an option in the future to distinguish between private and project settings (user configs / app internal settings via localstorage / ...)
- JSON is easier to understand for devs
- Typesafty is easier

Agree with the proposal 👍

ivanhofer:
I agree with the point that inlang does not need a json config file.

[...]

Issue #1140, opened 2023-07-19
[design flaw] user config, plugin api and application api are mutually exclusive


samuelstroschein:
## Problem 

The observation that we need:

1. a "user config" that must be CRUD operable. 
2. A plugin API that exposes stuff like #1116
3. an inlang instance with "app API" #1122

1 <-> 2 can't (?) be identical to enable simple CRUD ops. Hence, a singular `.js` config that combines "user" and "plugin config" is likely wrong. 

Furthermore:

- .js config is not easily typesafe
- .js config leads to design flaws like #1139 because the full power of code can be used (but is never used)

## Proposal 

Split inlang core APIs into three: 

1. A user-defined config. 
2. A plugin API that defines callbacks. 
3. An application API (an inlang app, see https://github.com/inlang/inlang/issues/1122#issuecomment-1641898910)

```
- core
  - plugin
    - api.ts
    - createPlugin.ts
  - config
    - schema.json
    - parseConfig.ts
  - app
    - api.ts
    - createApp.ts
```

#### 1. A user defined config (JSON)

- enable CRUD changes via apps
- simplify the config (no more env, fs, etc)
- achieve typesafety via https://json-schema.org/ (native support in VSCode)
- increase interop (other apps can simply read the json config)

Rough draft of user json config

This: 

```ts
/**
 * @type { import("@inlang/core/config").DefineConfig }
 */
export async function defineConfig(env) {
	const { default: jsonPlugin } = await env.$import(
		"https://cdn.jsdelivr.net/npm/@inlang/plugin-json@3/dist/index.js",
	)

	const { default: standardLintRules } = await env.$import(
		"https://cdn.jsdelivr.net/npm/@inlang/plugin-standard-lint-rules@3/dist/index.js",
	)

	return {
		referenceLanguage: "en",
		plugins: [
			jsonPlugin({ pathPattern: "source-code/website/lang/{language}.json" }),
			standardLintRules(),
		],
	}
}
```

Could turn into this:

```json

{
	"referenceLanguageTag": "en",
	"plugins": {
		"inlang.standardLintRules": {
			"module": "https://cdn.jsdelivr.net/npm/@inlang/plugin-standard-lint-rules@3/dist/index.js"
		},
		"inlang.plugin-json": {
			"module": "https://cdn.jsdelivr.net/npm/@inlang/plugin-json@3/dist/index.js",
			"options": {
				"pathPattern": "source-code/website/lang/{language}.json"
			}
		}
	}
}

```

[...]

Issue #1139, opened 2023-07-19
linting system won't work for translators


samuelstroschein:
## Problem

- nobody uses the lint settings feature 
- lints are not adjustable by apps (turning on, off, error level, etc.)
- lint rules will likely always be provided by plugins 

The following API is not CRUD operable because it's code and will always rely on developers to change lint settings. Hint: Translators and PMs will hate this design.

https://github.com/inlang/inlang/blob/4ce4c35d35cf11e5bf10a06af09471d20749f022/source-code/core/src/lint/createLintRule.test-d.ts#L30-L45

## Proposal 

- remove lint settings feature 
- set lint level via simple id-level object that can be CRUD by apps 

Concept API for #1140 

```json
{
  "lint": {
    "rules": {
       "inlang.missingMessage": "off"
    }
  }

```

samuelstroschein:
Further observations: the `createLintRule` function serves no benefit because (assumption) everyone will use `createPlugin`. 

- the main reason for `createLintRule` is the option to have scoped state and settings
- `createPlugin` serves the same purpose. 
- a lint rule could be vastly simplified by just saying "write a plugin that exposes a lint rule type"

samuelstroschein:
More observations: If a simple CRUD config to set the level of lint rules exists, than it totally makes sense to bundle standard lint rules in core by default. 

Users can (via apps even!) turn off "annoying" standard lint rules.

NilsJacobsen:
🤔 I'm not sure if that is the biggest pull factor for me to see CRUD as a benefit, but we can keep the door open with the conversion to JSON

samuelstroschein:
will be addressed in #1149

PR #1149, opened 2023-07-20
[core refactor] API spec

samuelstroschein:
## Context

Use this PR to draft the new core API together.

Watch this loom on how to navigate the code base https://www.loom.com/share/aa683611795946fa946ddf80b0e2dae7. 

## What you need to do 

Comments and, if you have a concrete proposal, open a PRs into `api-spec`, see #1152 as an example.

### IMPORTANT

We can always add APIs... please focus on bare minimum surface API that eases the path for future additions and refrain from proposing "nice to have" stuff.

---

#### Closes/succeeds  

- [x] #1144 
- [x] #1143
- [x] #1139 
- [x] #1140 
- [x] #1099 
- [x] #1111 
- [x] #1116 
- [x] #1098

### DIFF
--- a/source-code/core2/src/app/api.ts
+++ b/source-code/core2/src/app/api.ts
@@ -6,8 +6,6 @@ import type { InlangInstanceEnv } from "./env/types.js"
 export type InlangInstance = {
 	config: InlangConfig
 	env: InlangInstanceEnv
-	// TODO: exception handling
-	exceptions: InlangInstanceException[]
 	messages: {
 		query: MessagesQueryApi_1
 	}
@@ -16,12 +14,5 @@ export type InlangInstance = {
 		// see https://github.com/inlang/inlang/issues/1098
 		reports: LintReport[]
 		exceptions: LintException[]
-	}
-}
-
-class InlangInstanceException extends Error {
-	constructor(message: string) {
-		super(message)
-		this.name = "InlangInstanceException"
 	}
 }
--- a/source-code/core2/src/app/api.ts
+++ b/source-code/core2/src/app/api.ts
@@ -1,0 +1,27 @@
+import type { InlangConfig } from "../config/schema.js"
+import type { LintException, LintReport } from "../lint/api.js"
+import type { MessagesQueryApi_1 } from "../messages/query.js"
+import type { InlangAppEnvironment } from "./env/types.js"
+
+export type InlangApp = {
+	config: Readonly<InlangConfig>
+	env: InlangAppEnvironment
+	// TODO: exception handling
+	exceptions: InlangAppException[]
+	messages: {
+		query: MessagesQueryApi_1
+	}
+	lint: {
+		// for now, only simply array that can be improved in the future
+		// see https://github.com/inlang/inlang/issues/1098
+		reports: LintReport[]
+		exceptions: LintException[]
+	}
+}
+
+class InlangAppException extends Error {
+	constructor(message: string) {
+		super(message)
+		this.name = "InlangAppException"
+	}
+}
--- a/source-code/core2/src/app/api.ts
+++ b/source-code/core2/src/app/api.ts
@@ -4,15 +4,18 @@ import type { MessageQueryApi } from "../messages/query.js"
 import type { InlangInstanceEnv } from "./env/types.js"
 
 export type InlangInstance = {
-	config: InlangConfig
+	config: {
+		get: () => InlangConfig
+		set: (config: InlangConfig) => void
+	}
 	env: InlangInstanceEnv
-	messages: {
-		query: MessageQueryApi
+	query: {
+		messages: MessageQueryApi
 	}
 	lint: {
 		// for now, only simply array that can be improved in the future
 		// see https://github.com/inlang/inlang/issues/1098
-		reports: LintReport[]
-		exceptions: LintException[]
+		reports: () => LintReport[]
+		exceptions: () => LintException[]
 	}
 }
--- a/source-code/core2/src/app/api.ts
+++ b/source-code/core2/src/app/api.ts
@@ -4,7 +4,7 @@ import type { MessagesQueryApi_1 } from "../messages/query.js"
 import type { InlangInstanceEnv } from "./env/types.js"
 
 export type InlangInstance = {
-	config: Readonly<InlangConfig>
+	config: InlangConfig
 	env: InlangInstanceEnv
 	// TODO: exception handling
 	exceptions: InlangInstanceException[]
--- a/source-code/core2/src/config/schema.ts
+++ b/source-code/core2/src/config/schema.ts
@@ -9,6 +9,9 @@ export const InlangConfigSchema = z.object({
 	sourceLanguageTag: z.string(),
 	languageTags: z.array(z.string()),
 	plugins: z.array(PluginSchema),
+	lint: z.object({
+		rules: z.record(z.union([z.literal("off"), z.literal("warning"), z.literal("error")])),
+	}),
 })
 
 export type InlangConfig = z.infer<typeof InlangConfigSchema>
--- a/source-code/core2/src/config/schema.ts
+++ b/source-code/core2/src/config/schema.ts
@@ -1,0 +1,14 @@
+import { z } from "zod"
+
+const PluginSchema = z.object({
+	module: z.string(),
+	options: z.record(z.string()),
+})
+
+export const InlangConfigSchema = z.object({
+	sourceLanguageTag: z.string(),
+	languageTags: z.array(z.string()),
+	plugins: z.array(PluginSchema),
+})
+
+export type InlangConfig = z.infer<typeof InlangConfigSchema>
--- a/source-code/core2/src/fullExample.ts
+++ b/source-code/core2/src/fullExample.ts
@@ -69,7 +69,7 @@ const exampleInlangConfig: InlangConfig = {
 // 1. Create the app instance
 //    - env needs be created
 const inlang = createInlang({
-	config,
+	configPath: "/hello/inlang.config.json",
 	env: { fs: undefined as any, import: undefined as any },
 })
 
--- a/source-code/core2/src/fullExample.ts
+++ b/source-code/core2/src/fullExample.ts
@@ -1,0 +1,100 @@
+import { createApp } from "./app/createApp.js"
+import type { InlangConfig } from "./config/schema.js"
+import type {  MessageLintRule } from "./lint/api.js"
+import type { Plugin_Proposal_2 } from "./plugin/api.js"
+
+// --------------------- LINT RULE ---------------------
+
+const missingMessage: MessageLintRule = {
+	id: "inlang.missingMessage",
+	displayName: {
+		en: "Missing message",
+		de: "Fehlende Nachricht",
+	},
+	defaultLevel: "error",
+	type: "Message",
+	message: ({ message, inlang }) => {
+		if (message.languageTag !== inlang.config.sourceLanguageTag) {
+			return
+		}
+		// TODO pain to make reports typesafe...
+		// - MessageLintReport has a lot of duplicate information
+		const reports = []
+		for (const languageTag of inlang.config.languageTags) {
+			const translation = inlang.messages.query.get({ where: { id: message.id, languageTag } })
+			if (!translation) {
+				reports.push({
+					languageTag,
+					messageId: message.id,
+					content: `Message "${message.id}" is missing in language tag "${languageTag}".`,
+				})
+			}
+		}
+		return reports
+	},
+}
+
+// --------------------- PLUGIN ---------------------
+
+export const myPlugin: Plugin_Proposal_2<{ pathPattern: string }> = {
+	meta: {
+		id: "inlang.myPlugin",
+		displayName: "My Plugin",
+	},
+	setup: ({ inlang, options }) => {
+		return {
+			extendLanguageTags: () => {
+				return ["de"]
+			},
+			// OMG really nice API. Just pass an array of lint rule objects <3
+			addLintRules: () => [missingMessage],
+		}
+	},
+}
+
+// --------------------- APP ---------------------
+
+// --- SETUP ---
+
+// 0. Example config on disk
+const exampleInlangConfig: InlangConfig = {
+	sourceLanguageTag: "en",
+	languageTags: ["en"],
+	plugins: [
+		{ module: "https://example.com/myPlugin.js", options: { pathPattern: "src/**/*.{ts,tsx}" } },
+	],
+}
+
+// 1. The JSON config needs to be read manually.  
+//    - an app instance can't know where a config is located
+//    - gives apps more freedom to load different configs if desired (testing, etc.)
+const config = fs.readFile("./inlang.config.json")
+
+// 2. Create the app instance
+//    - env needs be created
+const inlang = createApp({
+	config,
+	env: { fs: undefined as any, import: undefined as any },
+})
+
+// --- CRUD ---
+inlang.messages.query.create({ id: "myMessageId", languageTag: "en", pattern: [{"type": "Text", value:  "Hello World" }]})
+
+// assuming that get is reactive
+const message = inlang.messages.query.get({ where: { id: "myMessageId", languageTag: "en" })
+
+inlang.messages.query.update({ where: { id: "myMessageId", languageTag: "en" }, data: { pattern: [{"type": "Text", value:  "Hello World" }] } })
+
+inlang.messages.query.delete({ where: { id: "myMessageId", languageTag: "en" } })
+
+// --- CONFIG ACCESS ---
+//   - config is static. adjusting the config requires a re-initialization of the app.
+//     IMHO (@samuelstroschein) that okay. 
+
+inlang.config.sourceLanguageTag
+inlang.config.languageTags
+
+// --- LINT ---
+inlang.lint.reports.filter((report) => report.level === "error")
+inlang.lint.exceptions
+
--- a/source-code/core2/src/fullExample.ts
+++ b/source-code/core2/src/fullExample.ts
@@ -63,6 +63,11 @@ const exampleInlangConfig: InlangConfig = {
 	plugins: [
 		{ module: "https://example.com/myPlugin.js", options: { pathPattern: "src/**/*.{ts,tsx}" } },
 	],
+  lint: {
+    rules: {
+      "inlang.missingMessage": "warning"
+    }
+  }
 }
 
 // 1. The JSON config needs to be read manually.  
--- a/source-code/core2/src/fullExample.ts
+++ b/source-code/core2/src/fullExample.ts
@@ -38,12 +38,10 @@ export const myPlugin: Plugin<{ pathPattern: string }> = {
 		displayName: { en: "My Plugin" },
 		description: { en: "My plugin description" },
 		keywords: [],
+		usedApis: ["addLintRules"],
 	},
 	setup: () => {
 		return {
-			extendLanguageTags: () => {
-				return ["de"]
-			},
 			// OMG really nice API. Just pass an array of lint rule objects <3
 			addLintRules: () => [missingMessage],
 		}
--- a/source-code/core2/src/fullExample.ts
+++ b/source-code/core2/src/fullExample.ts
@@ -1,7 +1,8 @@
 import { createInlang } from "./app/createInlang.js"
 import type { InlangConfig } from "./config/schema.js"
 import type { MessageLintRule } from "./lint/api.js"
-import type { Plugin_Proposal_2 } from "./plugin/api.js"
+import { createVariant } from "./messages/utilities.js"
+import type { Plugin } from "./plugin/api.js"
 
 // --------------------- LINT RULE ---------------------
 
@@ -13,16 +14,16 @@ const missingMessage: MessageLintRule = {
 	},
 	defaultLevel: "error",
 	message: ({ message, inlang, report }) => {
-		for (const languageTag of inlang.config.languageTags) {
+		for (const languageTag of inlang.config.get().languageTags) {
 			const translation = message.body[languageTag]
 			if (!translation) {
 				report({
 					languageTag,
 					messageId: message.id,
 					body: {
-						'en':`Message "${message.id}" is missing in language tag "${languageTag}".`,
-						'de':`Nachricht "${message.id}" fehlt für Sprachtag "${languageTag}".`,
-					}
+						en: `Message "${message.id}" is missing in language tag "${languageTag}".`,
+						de: `Nachricht "${message.id}" fehlt für Sprachtag "${languageTag}".`,
+					},
 				})
 			}
 		}
@@ -31,10 +32,12 @@ const missingMessage: MessageLintRule = {
 
 // --------------------- PLUGIN ---------------------
 
-export const myPlugin: Plugin_Proposal_2<{ pathPattern: string }> = {
+export const myPlugin: Plugin<{ pathPattern: string }> = {
 	meta: {
 		id: "inlang.myPlugin",
-		displayName: "My Plugin",
+		displayName: { en: "My Plugin" },
+		description: { en: "My plugin description" },
+		keywords: [],
 	},
 	setup: () => {
 		return {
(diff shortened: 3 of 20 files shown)


source-code/core2/src/app/api.ts
```diff
@@ -0,0 +1,27 @@
+import type { InlangConfig } from "../config/schema.js"
+import type { LintException, LintReport } from "../lint/api.js"
+import type { MessagesQueryApi_1 } from "../messages/query.js"
+import type { InlangAppEnvironment } from "./env/types.js"
+
+export type InlangApp = {
+	config: Readonly<InlangConfig>
+	env: InlangAppEnvironment
+	// TODO: exception handling
```
samuelstroschein, line 9: ## Global exception handling? 

### Problem 

Apps need to deal with exception handling which leads to not so good exception handling :D 

### Proposal 

Have a global exception listener that apps can subscribe to where all exceptions that happen within core are send to? 

**Pros**

- eases app building 
- identical exception handling in core 

**Con**

- "one solution to rule them all" could be bad.

source-code/core2/src/lint/api.ts
```diff
@@ -0,0 +1,46 @@
...
+export type LintRule = {
+	id: `${string}.${string}`
+	displayName: Record<string, string>
+	/**
+	 * The default level of the lint rule.
+	 *
+	 * Can be overwritten by the user config.
+	 */
+	defaultLevel: "error" | "warn"
+}
```
samuelstroschein, line 14: ## How to make linting easier 

### Problem 

Linting is difficult because the current API tries to be agnostic but idk that leads to difficult implementations

source-code/core2/src/messages/query.ts
```diff
@@ -0,0 +1,15 @@
...
+// 1. Query builder?
+// 2. Filtering etc how?
+//
+// inspiration: Prisma CRUD https://www.prisma.io/docs/concepts/components/prisma-client/crud
+export type MessagesQueryApi_1 = {
+	create: (message: Message) => void
+	get: () => Message[]
+	update: () => void
+	delete: () => void
+}
```
samuelstroschein, line 13: ## How to design Query API? 

### Problem 

We need a query API.

(Don't think about reactivity for now, that's a separate issue [#1122](https://github.com/inlang/inlang/issues/1122))

source-code/core2/src/config/schema.ts
```diff
@@ -0,0 +1,17 @@
...
+
+const PluginSchema = z.object({
+	module: z.string(),
+	options: z.record(z.string()),
+})
+
+export const InlangConfigSchema = z.object({
+	sourceLanguageTag: z.string(),
+	languageTags: z.array(z.string()),
+	plugins: z.array(PluginSchema),
```
samuelstroschein, line 11: ## Use key based plugin config schema or array?  

Current type: 

```
{
  plugins: [
    { 
      module: "https://..........."
      options: {

      }
    }
  ]
}
```

### Problem (?): 

Removing plugins manually is a bit cumbersome because one needs to identify the module link. 

### Question 

We could make manual removal of plugins easier. BUT, the plugin key (id) can (will) go out of sync with the returned plugin id of a plugin. 

```
{
  plugins: {
    inlang.plugin-i18next: {
      module: "https://......",
      options: {

      }
    }
  }
}

```

I vote for not worth increasing manual removal of plugins at the risk of out of sync plugin ids.

samuelstroschein:
@inlang/team @openscript homework for tomorrow :)

source-code/core2/src/config/schema.ts
NilsJacobsen: We already changed the naming/ids of plugins pretty often. That will happen in the future, too.
I think links are the better choice. 👍

source-code/core2/src/messages/query.ts
NilsJacobsen: Actually I love this prisma inspired API.

```ts
inlang.messages.query.create({ id: "myMessageId", languageTag: "en", pattern: [{"type": "Text", value:  "Hello World" }]})

const message = inlang.messages.query.get({ where: { id: "myMessageId", languageTag: "en" })

inlang.messages.query.update({ where: { id: "myMessageId", languageTag: "en" }, data: { pattern: [{"type": "Text", value:  "Hello World" }] } })

inlang.messages.query.delete({ where: { id: "myMessageId", languageTag: "en" } })
```

This brings us to the question of what is the return type of get?
To stay with prisma, they let you define specific properties as unique. So you could use `getUnique({ where: { id: "..." } })` to get only one return.  If you would use it in our case this wouldn't work because the combination of languageTag and id is unique. They also have `getFirst()` with returns also one and `getMany()` with returns the array. 

Personally, I do not like that there are this many variations. What do you think?

Do I get all when just using `inlang.messages.query.get()`?

source-code/core2/src/lint/api.ts
NilsJacobsen: The question is will the community use the agnostic linting features more often? Because today, nobody use something other than standard-lint-rules?

Level configuration will become more important with the rise of github actions/ CI. :D

source-code/core2/src/app/api.ts
NilsJacobsen: Makes sense to me 👍 While working on the editor I would have loved to use something like that.

source-code/core2/src/config/schema.ts
felixhaeberle: > We already changed the naming/ids of plugins pretty often

- I do not think that this should be a common practice (how often do npm packages names change?)
- module link & identifier ("inlang.plugin-i18next") are even duplicating themselves
- giving an identifier & a dedicated field for version (for example in options) is also a way to go that we could differentiate behavior with versions in the future

source-code/core2/src/lint/api.ts
felixhaeberle: +1 lints in general & levels could be more sophisticated.
it's currently loved by our users although it's dead simple.
maybe huge potential in here.  

- not only `string`
- could be object of type `Level` where multiple properties can be exposed

source-code/core2/src/config/schema.ts
NilsJacobsen: > I do not think that this should be a common practice anymore

Although it maybe is not a problem for us, in an ecosystem with hundreds of community plugins, this can become an issue.

source-code/core2/src/messages/query.ts
felixhaeberle: There need to be filter options in the crud functions. In my opinion it returns either `Message` or `Message[]` based on that you have chosen:

- only `id` provided -> `Message[]`
- `id` & `languageTag` -> `Message`
- `languageTag` & `data.pattern["type"]: "Text"` -> `Message[]`

There is no need to have this many functions as of now; controlling the returned results can/should be done by apps. 
We can potentially add them later if the apps have an extreme urge.

Do we want those query options and not just a regular filter? -> `{where: { /**/ }}, {include: {/**/}}, {select: {/**/}}, ...` syntax? https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#model-query-options

source-code/core2/src/config/schema.ts
samuelstroschein: > Although it maybe is not a problem for us, in an ecosystem with hundreds of community plugins, this can become an issue.

In addition, transfering plugins would become unfeasible. 

Seems like we agree with "array of objects" with no identifiers because plugins define the ids. 

Vote with 👍 or 👎

source-code/core2/src/lint/api.ts
samuelstroschein: > defaultLevel
> could be object of type Level where multiple properties can be exposed

Why? What is the use case?

source-code/core2/src/messages/query.ts
samuelstroschein: > @felixhaeberle Do we want those query options and not just a regular filter? -> {where: { /**/ }}, {include: {/**/}}, {select: {/**/}}, ... syntax? https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#model-query-options

I don't understand the question. Please elaborate. 

> In my opinion it returns either Message or Message[] based on that you have chosen:

If you establish and enforce that each message with a `id-languageTag` combination can only exists once, we can distinguish if `get` returns a single message or an array. 

BUT if we introduce variants (A/B testing etc), the `get` function would be breaking because the unique combination would become `id-languageTag-variant`. 

Prisma gets around it by having `get` and `getMany`, `create`, `createMany`. Separating the query ops in single and many future proofs the API.

NilsJacobsen: 👍 `get` and `getMany`, `create`, `createMany`. I do not love the variation of queries, but I think those set very distinct expectations and that is a good think. (I don't know if we need createMany, do you think of bulk translations?)

source-code/core2/src/config/schema.ts
felixhaeberle: I can get myself into links, but just wanted to add that with npm, there is also no way to rename an already published package. They resolve it like: https://stackoverflow.com/a/38632439 

with `options.version` & `identifier`, we would follow npm registry behavior until we have our own registry.

source-code/core2/src/app/api.ts
NiklasBuchfink: The pros outweigh the cons; this makes sense and simplifies app building.

source-code/core2/src/lint/api.ts
felixhaeberle: I see a lot of systems going with `severity` like from `1` to `5` in addition to the `level`.
https://about.gitlab.com/handbook/engineering/quality/issue-triage/#severity

but maybe that's not in `defaultLevel` but belongs in `LintRule`.

source-code/core2/src/app/api.ts
felixhaeberle: In order to decide on that, which data structure would be exposed by the `global exception`? If I just get a shitload of data in the exception and can't properly react to it, nobody will introduce them to their app. 😄 

Also, we then need a code pattern for app developers to react on exceptions.

source-code/core2/src/messages/query.ts
samuelstroschein: > (I don't know if we need createMany, do you think of bulk translations?) 

For consistency, would be good to have `createMany` as well then.

source-code/core2/src/config/schema.ts
samuelstroschein: > with options.version & identifier, we would follow npm registry behavior until we have our own registry. 

NPM is too complex. Resolve from a link is vastly simpler. Resolving because we agreed.

source-code/core2/src/plugin/api.ts
```diff
@@ -0,0 +1,86 @@
...
+export type Plugin_Proposal_2<Options extends Record<string, string> = Record<string, never>> = {
+	meta: {
+		id: `${string}.${string}`
+		displayName: string
+	}
+	setup: (args: { options: Options; inlang: InlangApp }) => {
+		extendLanguageTags?: () => LanguageTag[]
+		loadMessages?: () => Message[]
+		saveMessages?: (args: { messages: Message[] }) => void
+		addLintRules?: () => LintRule[]
```
NilsJacobsen, line 61: Wouldn't it be changeLintRules?
Can be plugins be treated the same way? No, because they would replace themselves. 
Do you guys see a solution for that? Maybe out of scope for now. But I see a settings menu in the editor where I could literaly configure the config. Which would be nice 👍

source-code/core2/src/lint/api.ts
samuelstroschein: > but maybe that's not in defaultLevel but belongs in LintRule. 

We can always add APIs... please focus on bare minimum surface API that eases the path for future additions and refrain from proposing "nice to have" stuff.

NiklasBuchfink: Lints should be easy to implement and more used by our users. It's always a big selling point in the first place and the best call to action to improve globalization and use inlang. This should be as easy to create a rule in Grammarly.

Lint rules are used to:
- automate
- communicate between the developer and translator
- define standards for translations

This makes it easier to create and the `displayName` is a good addition 👍 

Quick note: I have the feeling the way we implement linting will be similar to the implementation of collaborative features like flagging translations for a review, as accepted or as blocked by a role, but maybe this shouldn't be mixed up atm.

source-code/core2/src/app/api.ts
ivanhofer: I would delay the exception handling discussion for after the refactor

felixhaeberle: Agree 👍

source-code/core2/src/lint/api.ts
samuelstroschein: @NiklasBuchfink your reply reads like ChatGPT lol. 

The question is not if we have lint rules, the question is how should the API for lint rules look like.

source-code/core2/src/plugin/api.ts
samuelstroschein: The API is runtime context. 

The API DOES NOT extend the config. Hence, you can't remove lint rules. You only add lint rules.

source-code/core2/src/plugin/api.ts
```diff
@@ -0,0 +1,86 @@
...
+ * - arguably not so nice API.
+ */
+export type Plugin_Proposal_2<Options extends Record<string, string> = Record<string, never>> = {
+	meta: {
+		id: `${string}.${string}`
+		displayName: string
+	}
+	setup: (args: { options: Options; inlang: InlangApp }) => {
+		extendLanguageTags?: () => LanguageTag[]
+		loadMessages?: () => Message[]
```
ivanhofer, line 59: I would split lint rules and messages plugins. This API will grow once we have video audio etc. APIs, making it less clear "what do I need to provide, when I want to write a lint rule"

source-code/core2/src/plugin/api.ts
```diff
@@ -0,0 +1,86 @@
...
+		saveMessages?: (args: { messages: Message[] }) => void
```
ivanhofer, line 60: similar reason to above, I would put it into a nested object
```ts
messages : {
	read: () => Messages[]
	write: (args: { messages: Messages[] }) => void
}
```

source-code/core2/src/plugin/api.ts
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * Proposal 2
+ *
+ * Split static meta data from setup function.
+ *
+ * + access to static without crash chance
+ * - arguably not so nice API.
+ */
+export type Plugin_Proposal_2<Options extends Record<string, string> = Record<string, never>> = {
```
ivanhofer, line 52: I vote for proposal 2

source-code/core2/src/plugin/api.ts
samuelstroschein: Distinguishing different sort of plugins seems wrong. 

Let plugins use the plugin API however they want and need. If you only want to add a lint rule, only use `addLintRule`. The concept of having one "entrypoint" to extend inlang, namely creating a plugin, is superior to "you can create a lint rule via a lint rule plugin", "you can create a message plugin via a message plugin", etc. 

No, just "you can create a plugin". "In a plugin you can add lint rules, or define how messages should be loaded and saved". etc.

source-code/core2/src/fullExample.ts
```diff
@@ -0,0 +1,105 @@
+import { createApp } from "./app/createApp.js"
+import type { InlangConfig } from "./config/schema.js"
+import type {  MessageLintRule } from "./lint/api.js"
+import type { Plugin_Proposal_2 } from "./plugin/api.js"
+
+// --------------------- LINT RULE ---------------------
+
+const missingMessage: MessageLintRule = {
+	id: "inlang.missingMessage",
+	displayName: {
```
ivanhofer, line 10: we probably need a "fallback" prop, because "en" won't be present everywhere

samuelstroschein: We should force `en` to be available.

source-code/core2/src/fullExample.ts
```diff
@@ -0,0 +1,105 @@
...
+		en: "Missing message",
+		de: "Fehlende Nachricht",
+	},
+	defaultLevel: "error",
+	type: "Message",
+	message: ({ message, inlang }) => {
```
ivanhofer, line 16: I would keep the API, that we already have currently.
Here you forgot to pass the lint level and the id of the lint rule. Sure this can be altered afterwards, but I would keep the `.report()` API.

source-code/core2/src/fullExample.ts
```diff
@@ -0,0 +1,105 @@
+import { createApp } from "./app/createApp.js"
+import type { InlangConfig } from "./config/schema.js"
+import type {  MessageLintRule } from "./lint/api.js"
+import type { Plugin_Proposal_2 } from "./plugin/api.js"
+
+// --------------------- LINT RULE ---------------------
+
+const missingMessage: MessageLintRule = {
+	id: "inlang.missingMessage",
```
ivanhofer, line 9: should probably also belong into a `meta` object

source-code/core2/src/fullExample.ts
samuelstroschein: The report api makes the implementation vastly more complex. 

Simply returning a LintReport object is much simpler both internally and externally (explaining the system). Hence, I vote against `report`.

source-code/core2/src/plugin/api.ts
samuelstroschein: Idk. If we start to nest the API, we always have to decide "this belongs to messages" -> "no it belongs to ide extension" etc. 

A simple flat callback list takes away the mental overhead. In the future, we could start to bundle APIs. But, in the short run i think we are better of with a flat list.

source-code/core2/src/lint/api.ts
felixhaeberle: my conclusion: we can always add more properties to `LintRule` to make it more flexible. currently, no need.

source-code/core2/src/fullExample.ts
ivanhofer: No, If I only create a rule for my italian company and we only support it and de, why should I add an english text?

source-code/core2/src/fullExample.ts
ivanhofer:
```ts
const missingMessage: MessageLintRule = {
	id: "inlang.missingMessage",
	displayName: {
		en: "Missing message",
		de: "Fehlende Nachricht",
	},
	defaultLevel: "error",
	type: "Message",
	message: ({ message, inlang, report }) => {
		if (message.languageTag !== inlang.config.sourceLanguageTag) {
			return
		}
		// TODO pain to make reports typesafe...
		// - MessageLintReport has a lot of duplicate information
		for (const languageTag of inlang.config.languageTags) {
			const translation = inlang.messages.query.get({ where: { id: message.id, languageTag } })
			if (!translation) {
				report({
					languageTag,
					messageId: message.id,
					content: `Message "${message.id}" is missing in language tag "${languageTag}".`,
				})
			}
		}
	},
}
```

no need to return an array, just call the `report` function.

source-code/core2/src/fullExample.ts
samuelstroschein: > No, If I only create a rule for my italian company and we only support it and de, why should I add an english text? 

Because inlang apps (like the editor) are globalized. 

The fallback is English. If you want to add translations for the display name of your lint rule, go for it and add Italian. Really minor issue. Forcing English to be provided seems vastly superior to a fallback option in case a "company doesn't want to write two words" for the English display name. I will push against the proposal.


CleanShot 2023-07-21 at 12 26 28@2x

ivanhofer: If I write a private plugin (no plan to publish it) ,why should I care about en?
It is not just the display name.. also the lint messages would need to be translated accordingly

source-code/core2/src/plugin/api.ts
ivanhofer: the `inlang` object will also have a nested `messages` object ;)
It would be consitent

source-code/core2/src/fullExample.ts
samuelstroschein: > If I write a private plugin (no plan to publish it) ,why should I care about en?

Because translators in your company might use the editor in English. 

> It is not just the display name.. also the lint messages would need to be translated accordingly

Correct. Forgot to add it. 

TBH it seems very straightforward that lint messages must provide `en` for the sake of providing a good UX across the inlang ecosystem.

source-code/core2/src/fullExample.ts
samuelstroschein: PR is out https://github.com/inlang/inlang/pull/1154#issue-1815593130. 

Closing this discussion in favor of the PR>

source-code/core2/src/messages/query.ts
felixhaeberle: > I don't understand the question. Please elaborate.

@samuelstroschein Elaborated in person

Are we still agreeing with `get` and `getMany`, `create`, `createMany` – or does a more simpler API like `messages.list()` make sense?

NiklasBuchfink: 1. Is an `updateMany` also needed for consistency?
2. Don't we need an `upsert` function?
3. I see the use case where we want to push multiple changes (some are updates, some creates); therefore a `upsertMany` would be interesting.

source-code/core2/src/app/api.ts
felixhaeberle: @samuelstroschein Postpone exception handling discussion and resolve conversation?

source-code/core2/src/plugin/api.ts
NilsJacobsen: What is this rule then doing?

samuelstroschein: It adds lint rules to the inlang instance which can be processed by the instance linter.

source-code/core2/src/fullExample.ts
NilsJacobsen: I would say 98% of our repos are using `en` as their default. However https://github.com/tournesol-app/tournesol/pull/1610 for example have `fr` as a default.

samuelstroschein: This is about lint rules. I am 200% certain that we can force everyone to provide `en` for the sake of being able to globalize inlang apps themselves.

felixhaeberle: I have built a lot of web projects in the past and can confidently say that NOT all webpages want English by default. Especially for the EU market, people often want their language as the default. (Speaking about German, French, Italian, Spanish, Portuguese, ...). 

Proposal: We can derive the required languages for lint translations from `config.languageTags` via TS?

samuelstroschein: GUYS OMG STOP MIXING STUFF. 

THIS IS ABOUT LINT RULES, NOT ABOUT ORGANIZATIONS USING INLANG TO GLOBALIZE THEIR APPS. PR will follow but sorry this discussion is completely derailed.

felixhaeberle: Sorry! I thought my input would be valuable.
GitHub didn't update my feed so I didn't saw your comment: 

> This is about lint rules. I am 200% certain that we can force everyone to provide en for the sake of being able to globalize inlang apps themselves.

source-code/core2/src/messages/query.ts
NilsJacobsen: I think it makes more sense to make one PR and comment on that.

samuelstroschein: @NilsJacobsen give a reason

source-code/core2/src/fullExample.ts
samuelstroschein: PR is up https://github.com/inlang/inlang/pull/1155. Closing this in favor of the PR

source-code/core2/src/messages/query.ts
NilsJacobsen: We gonna have multiple pretty similar starting points that are inspired by prisma. If you want I can hand in mine. Not sure if that is the most effective way of doing it.

source-code/core2/src/app/api.ts
samuelstroschein: Postponed. [7f9b9fc](https://github.com/inlang/inlang/pull/1149/commits/7f9b9fc8a7ce44b281857343111bfe978c7a3e65)

source-code/core2/src/messages/query.ts
samuelstroschein: Okay, then you propose a concrete API in a PR? @felixhaeberle and @NiklasBuchfink you?

NiklasBuchfink: I don't need to propose a full API. Due to different expected returns, I agree with `get` and `getMany`. 

I just wanted to bring to the discussion that `upsert` might still be helpful. If there should be an `updateMany` or `upsertMany` should be discussed if we see use cases for that.

For example, when applying multiple changes, do we need all three `updateMany`,  `createMany`, and `upsertMany`?

NiklasBuchfink (approved):

felixhaeberle (approved):

Issue #1122, opened 2023-07-18
discussion about reactivity architecture 


samuelstroschein:
## Problem

1. how to do caching? 
2. reactivity? messages change, how to propagate to linting, updating the UI, etc. 
3. where to "solve the complexity": plugins, apps, or core? 

## Additional information 

- app state problem driven by persistent layer (repository), and app state
- maybe something like https://tinybase.org/ can help the app state layer but is likely not interop between different apps

NilsJacobsen:
Take that as a thought experiment. It's not about api definitions, it's about functional responsibilities of the ecosystem.
Watch the -> [Loom](https://www.loom.com/share/0cafc240f89241c4bc04aef8914a1ada?sid=ab9d2942-7e38-41d9-a4d3-bdc28520d67e)

<img width="689" alt="image" src="https://github.com/inlang/inlang/assets/58360188/08b0d603-1e93-427c-9ef8-46e00d2905aa">

[Figjam](https://www.figma.com/file/mv2W6vqffN22BARI3hRy2o/organization-structure?type=whiteboard&node-id=0%3A1&t=trnby3GzFDXtHH69-1)

NilsJacobsen:
I always love to hear that the plugins do not get more complexity :D 

If the watcher is in core, do we want that core handles file changes. I thought we made a clear cut from files to only messages in core and app? Or only for apps?

Otherwise I really like the watching being part of core. And query being the central interface for apps. Must admit, that I do not know in detail how to establish reactivity for query without using framework specific technology. 

Only using reactive CRUD would be absolute best case 👍

felixhaeberle:
@NilsJacobsen Thank you for your Loom and what you are thinking of the architectural approaches.

I will conclude based on our discussion:
- reactivity is likely needed for apps (or some kind of)
- query should likely have caching built-in
- there need to be some setter/getter functions
- there is a need to define an updating process that is easy

## Proposal WIP do not comment on it 😂

```ts
// Message Type
type Message = {
  id: string;
  // Define your message properties here
};

// Plugin Type
type Plugin = {
  getMessages: ({ filter }: { filter: Message["id"][] }) => Message[];
  saveMessages: (messages: Message[]) => void;
};

// Messages Type
type Messages = {
  set: (messages: Message[]) => void;
  exec: () => void;
  get: () => Message[];
};

// Core Implementation
const createCore = (plugins: Plugin[], _messages: Messages) => {
  const buildQueryFromPlugins = (): void => {
[...]

PR #1152, opened 2023-07-21
[api spec] load config by createApp

samuelstroschein:
## Problem 

Changing the user config via an app becomes impossible if the `createApp` instance doesn't know how to write changes of the config back to the filesystem. 

## Proposal 

Pass the path of a config to `createApp`; let `createApp` read (and write) the config. 

**Pros**

- apps can very easily perform CRUD ops on the user config
- apps don't need to reinitialize `createApp` (bad UX) 

**Cons**

- [neglactable] testing becomes slightly harder because one must call fs.writeFile("inlang.config.json") before.

### DIFF
--- a/source-code/core2/src/fullExample.ts
+++ b/source-code/core2/src/fullExample.ts
@@ -1,4 +1,4 @@
-import { createApp } from "./app/createApp.js"
+import { createInlang } from "./app/createInlang.js"
 import type { InlangConfig } from "./config/schema.js"
 import type {  MessageLintRule } from "./lint/api.js"
 import type { Plugin_Proposal_2 } from "./plugin/api.js"
@@ -68,8 +68,8 @@ const exampleInlangConfig: InlangConfig = {
 
 // 1. Create the app instance
 //    - env needs be created
-const inlang = createApp({
-	configPath: "/hello/inlang.config.json",
+const inlang = createInlang({
+	config,
 	env: { fs: undefined as any, import: undefined as any },
 })
 
--- a/source-code/core2/src/fullExample.ts
+++ b/source-code/core2/src/fullExample.ts
@@ -12,25 +12,20 @@ const missingMessage: MessageLintRule = {
 		de: "Fehlende Nachricht",
 	},
 	defaultLevel: "error",
-	type: "Message",
-	message: ({ message, inlang }) => {
+	message: ({ message, inlang, report }) => {
 		if (message.languageTag !== inlang.config.sourceLanguageTag) {
 			return
 		}
-		// TODO pain to make reports typesafe...
-		// - MessageLintReport has a lot of duplicate information
-		const reports = []
 		for (const languageTag of inlang.config.languageTags) {
 			const translation = inlang.messages.query.get({ where: { id: message.id, languageTag } })
 			if (!translation) {
-				reports.push({
+				report({
 					languageTag,
 					messageId: message.id,
 					content: `Message "${message.id}" is missing in language tag "${languageTag}".`,
 				})
 			}
 		}
-		return reports
 	},
 }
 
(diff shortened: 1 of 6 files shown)


source-code/core2/src/fullExample.ts
```diff
@@ -70,15 +71,10 @@ const exampleInlangConfig: InlangConfig = {
...
-//    - an app instance can't know where a config is located
-//    - gives apps more freedom to load different configs if desired (testing, etc.)
-const config = fs.readFile("./inlang.config.json")
-
-// 2. Create the app instance
+// 1. Create the app instance
 //    - env needs be created
 const inlang = createApp({
-	config,
+	configPath: "/hello/inlang.config.json",
```
felixhaeberle, line 77: we could encourage people to pass the config path via an env variable, but certainly not must because we do not have sensitive information (yet) in the config (google translate keys)
furthermore, do we want to allow URLs here? (makes read/write hard or impossible)

https://www.prisma.io/docs/reference/database-reference/connection-urls#env

source-code/core2/src/fullExample.ts
```diff
@@ -93,12 +89,13 @@ inlang.messages.query.update({ where: { id: "myMessageId", languageTag: "en" },
...
 
 // --- CONFIG ACCESS ---
-//   - config is static. adjusting the config requires a re-initialization of the app.
-//     IMHO (@samuelstroschein) that okay. 
 
 inlang.config.sourceLanguageTag
 inlang.config.languageTags
 
+// adding a language tag via an app
+inlang.config.languageTags = [...inlang.config.languageTags, "de-AT"]
```
felixhaeberle, line 97: config is mutable then? 😅

NilsJacobsen:
I agree with the proposal 👍

- Even though we ship more files(resource)-specific code to the core. But I guess the tradeoff is fine.

source-code/core2/src/fullExample.ts
samuelstroschein: > we could encourage people to pass the config path via an env variable, but certainly not a must because we do not have sensitive information (yet) in the config (google translate keys)

Yeah... no use case. 

> furthermore, do we want to allow URLs here? (makes read/write hard or impossible)

Prob not. In the future we can add a new prop `config` that allows manual passing of the config. 

Please resolve this conversation if you agree that we shouldn't change anything here.

source-code/core2/src/fullExample.ts
samuelstroschein: I mean everything is mutable in inlang anyways. But you want read/write separation? 

We could have an API like 

```ts
// get a prop
inlang.config.get()

// set a prop
inlang.config.set()

```

But, that really depends on #1122

source-code/core2/src/fullExample.ts
NilsJacobsen: Url to the inlang.config.json will never be sensitive right? 

> https://www.prisma.io/docs/reference/database-reference/connection-urls#env

Haha slowly this really feels like building git as a database :DDD

samuelstroschein: > Url to the inlang.config.json will never be sensitive right? 

No, it won't. I am closing this discussion. 

> Haha slowly this really feels like building git as a database :DDD

Right now, we are talking about building lisa prisma (the thing in memory) :)

ivanhofer (approved):
I aggree on the idea

samuelstroschein:
@NilsJacobsen @felixhaeberle @NiklasBuchfink please approve or reject.

felixhaeberle (approved):

NiklasBuchfink (approved):

NilsJacobsen (approved):
Aprove but need to change to createInlang right?

samuelstroschein:
@NilsJacobsen https://github.com/inlang/inlang/pull/1153

samuelstroschein:
@felixhaeberle please leave merging to the assigned person. Hella confusing otherwise

samuelstroschein:
@felixhaeberle you merged wrong... you overwrote https://github.com/inlang/inlang/pull/1153

samuelstroschein:
@felixhaeberle please do nothing

felixhaeberle:
sorry! I merged too fast here – you should merge 👍

PR #1154, opened 2023-07-21
[api spec] use lint report function

samuelstroschein:
## Context

Follow up of https://github.com/inlang/inlang/pull/1149#discussion_r1270506548. 

## Observations 

The API becomes simpler and more agnostic:

#### 1. No manual "type": "Message" for a lint rule. 

If a message callback is provided, the lint rule is for a message. 

In the future, we can extend lint rules by other callbacks. 

#### 2. The report function makes "reports" a bit easier

No more `const result: LintReport` which lead to a lot of type issues.

### DIFF
--- a/source-code/core2/src/fullExample.ts
+++ b/source-code/core2/src/fullExample.ts
@@ -12,25 +12,20 @@ const missingMessage: MessageLintRule = {
 		de: "Fehlende Nachricht",
 	},
 	defaultLevel: "error",
-	type: "Message",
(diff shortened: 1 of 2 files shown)


felixhaeberle (approved):

NiklasBuchfink (approved):

samuelstroschein:
Merging because @ivanhofer proposed this and @NilsJacobsen will likely approve. (Is blocking a follow up PR of mine).

PR #1155, opened 2023-07-21
[api spec] make lints translatable 

samuelstroschein:
## Problem 

Inlang apps like the IDE-extension and web editor can't show messages in the users language. 

## Proposal 

1. Make lint reports _optionally_ translatable
2. Always force language tag "en" to be provided as fallback. 

https://github.com/inlang/inlang/assets/35429197/bda0b105-ce23-4263-840b-f0be2a9d2f45

### DIFF
--- a/source-code/core2/src/fullExample.ts
+++ b/source-code/core2/src/fullExample.ts
@@ -22,7 +22,10 @@ const missingMessage: MessageLintRule = {
 				report({
 					languageTag,
 					messageId: message.id,
-					content: `Message "${message.id}" is missing in language tag "${languageTag}".`,
+					body: {
+						'en':`Message "${message.id}" is missing in language tag "${languageTag}".`,
+						'de':`Nachricht "${message.id}" fehlt für Sprachtag "${languageTag}".`,
+					}
 				})
 			}
 		}
--- a/source-code/core2/src/lint/api.ts
+++ b/source-code/core2/src/lint/api.ts
@@ -47,4 +47,4 @@ export class LintException extends Error {
  *
  * The language tag `en` is always required.
  */
-type TranslatedStrings = Record<string, string> & { en: string }
+type TranslatedStrings = Record<LanguageTag, string> & { en: string }
--- a/source-code/core2/src/lint/api.ts
+++ b/source-code/core2/src/lint/api.ts
@@ -4,7 +4,7 @@ import type { Message } from "../messages/schema.js"
 
 export type LintRule = {
 	id: `${string}.${string}`
-	displayName: Record<string, string>
+	displayName: TranslatedStrings
 	/**
 	 * The default level of the lint rule.
 	 *
@@ -20,13 +20,13 @@ export type MessageLintRule = LintRule & {
 export type ReportMessageLint = (args: {
 	messageId: Message["id"]
 	languageTag: LanguageTag
-	content: LintReport["content"]
+	body: LintReport["body"]
 }) => MessageLintReport
 
 export type LintReport = {
 	ruleId: LintRule["id"]
 	level: "error" | "warn"
-	content: string
+	body: TranslatedStrings
 }
 
 export type MessageLintReport = LintReport & {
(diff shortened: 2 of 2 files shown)


samuelstroschein:
I will 10000000% confident that this is the right todo

NiklasBuchfink (approved):

NilsJacobsen (approved):
Now we're talking 👍

felixhaeberle:
@samuelstroschein 

> I am 10000000% confident that this is the right to do

Got your point with "en".

But, then you have to call your translator every time you add a new language and ask him what "identicalPattern" in, for example, Chinese stand for, and adjust your lint rule based on that. 

If the "add language" button is clicked in the editor, and someone adds "Portuguese", there will be "identicalPattern" until the dev has the time and add "padrãoIdêntico". 

Or do I still not get it?

samuelstroschein:
> Or do I still not get it?

I think you are still not getting it. 

- Actions in the editor like "add language tag" are irrelevant to lint rules being translateable. 
- translated lint rules allow the editor to show a German user (PM, my grandma, translator, whoever) that "Die Nachricht mit dem Sprachtag 'en' fehlt." 

Stell dir einfach vor Eslint zeigt deutschen usern deutschsprachige warnungen an.

felixhaeberle:
So then we are not connecting the languages a project is translated to and the lint translations at all? We could have, for whatever reason, a project which is in Spanish & English and the lint rule could expose languages for Spanish, English, German, ... and so on. 👍 

Yep, now we are on the same page.

felixhaeberle (approved):

samuelstroschein:
> So then we are not connecting the languages a project is translated to and the lint translations at all?  

Exactly. Those are two separate pair of shoes. 

> standard-lint rules then probably needs translations to all the languages 😄

Yep ^^

ivanhofer (changes requested):
Reporting a lint violation for 100 languages in this way is far from ideal..

What if we use the inlang function (like in the SDK) to let developers translate also plugins? 🤯
Plugins can have it's own `inlang.config.js` function and can be opened and edited in the editor 😎

samuelstroschein:
> Reporting a lint violation for 100 languages in this way is far from ideal..

This is just the API and the simplest I can think of. Make a counter proposal. 

> What if we use the inlang function (like in the SDK) to let developers translate also plugins? 🤯
> Plugins can have it's own inlang.config.js function and can be opened and edited in the editor 😎
 
You can do that with the proposed API _without us forcing you to use the SDK_. 

```ts
report({
  messageId: message.id,
  body: i("missing-message")
  // ...
})
```

ivanhofer:
> Make a counter proposal

Will do on monday morning

ivanhofer:
> * plugin authors can _magically_ (aka edge cases won't work) use `i`

Doesn't inlang want to be the company that makes i18n as easy as possible. In my opinion this also includes not just metaframeworks but also CLI applications, figma plugins, and yes, also inlang plugins ;)

> * plugin writing becomes more complex because you force SDK usage

We don't force it, we suggest it.. People can still publish plugins without translations if they want.

> * SDK becomes more complex because you need an esbuild plugin

We have a template and can integrate it already. So no need for a developer to set something up. It will just work.

> * apps become more complex because you need to communicate to the SDK what language it should return

We have the same problem in every application. And that's why the `inlang.config.json` file exists.

> * plugins become larger because each plugin bundles the SDK

Is 1kb to much? 🙃 I don't know the exact bundle size, but it should be less than that.

samuelstroschein:
> Doesn't inlang want to be the company that makes i18n as easy as possible. 

Yes, but you are overshooting here. 

Requiring a dictionary (JS object) that requires `en` to be defined is the simplest API you can expose. Every developer is free to manually set up a dictionary, or use whatever i18n lib they want. The reality is that nobody will use an i18n lib for this use case. Because an i18n lib is a complete overkill. 

Merging this PR as "3 vs 1 in favor of translated strings"

PR #1157, opened 2023-07-21
[api-spec] query api draft nils

NilsJacobsen:
It's a draft.

I want to keep it as easy as possible:
Because there will be confusion with return types for get, I added `getMany`. The rest stays straightforward.

I also added Message as return type of `get`, `update`, `create` and `delete`. Was inspired by prisma. Helps with framework specific data handling.

Current:

```ts
export type MessagesQueryApi = {
	create: (args: Message) => [Message, Error]
	get: (args: { where: UniqueFilterType }) => [Message, Error]
	getMany: (args: { where: FilterType }) => [Message[], Error]
	update: (args: Message) => [Message, Error]
	updateMany: (args: Message[]) => [Message[], Error]
	upsert: (args: Message) => [Message, Error]
	upsertMany: (args: Message[]) => [Message[], Error]
	delete: (args: { where: UniqueFilterType }) => [Message, Error]
	deleteMany: (args: { where: FilterType }) => [Message[], Error]
}
```

### DIFF
--- a/source-code/core2/src/messages/query.ts
+++ b/source-code/core2/src/messages/query.ts
@@ -1,14 +1,15 @@
 import { createInlang } from "../app/createInlang.js"
+import type { LanguageTag } from "../languageTag.js"
 import type { Message } from "./schema.js"
 
 type UniqueFilterType = {
-	id: string
-	languageTag: string
+	id: Message["id"]
+	languageTag: LanguageTag
 	//variant: string -> in the future
 }
 type FilterType = {
-	id?: string
-	languageTag?: string
+	id?: Message["id"]
+	languageTag?: LanguageTag
 	//variant?: string -> in the future
 }
 
--- a/source-code/core2/src/messages/query.ts
+++ b/source-code/core2/src/messages/query.ts
@@ -1,3 +1,4 @@
+import { createInlang } from "../app/createInlang.js"
 import type { Message } from "./schema.js"
 
 type UniqueFilterType = {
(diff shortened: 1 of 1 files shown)


NiklasBuchfink (changes requested):
We should extend it with an `upsert` function, as it's commonly used when applying changes. We don't want to check if a message with the id already exists every time we iterate over changes, then use update or create.

source-code/core2/src/messages/query.ts
```diff
@@ -1,15 +1,31 @@
...
 // inspiration: Prisma CRUD https://www.prisma.io/docs/concepts/components/prisma-client/crud
 export type MessagesQueryApi_1 = {
-	create: (message: Message) => void
-	get: () => Message[]
-	update: () => void
-	delete: () => void
+	create: (message: Message) => Message
+	get: (where: UniqueFilterType) => Message
+	getMany: (where: FilterType) => Message[]
+	update: (where: UniqueFilterType, data: Message["pattern"]) => Message
```
ivanhofer, line 19: update should also be possible with a single payload like this:
```ts
const message = inlang.messages.query.update({ id: "myMessageId", languageTag: "en", pattern: [{"type": "Text", value: "Hello World" }] })
```

NilsJacobsen: Should both be possible? I feel like we should decide on one.

source-code/core2/src/messages/query.ts
```diff
@@ -1,15 +1,78 @@
...
 // inspiration: Prisma CRUD https://www.prisma.io/docs/concepts/components/prisma-client/crud
 export type MessagesQueryApi_1 = {
-	create: (message: Message) => void
-	get: () => Message[]
-	update: () => void
-	delete: () => void
+	create: (args: Message) => Message
+	get: (args: GetType) => Message
+	getMany: (args: GetManyType) => Message[]
+	update: (args: UpdateType) => Message
```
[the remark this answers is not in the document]
ivanhofer: yep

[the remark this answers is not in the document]
NilsJacobsen: Do we then even need the UniqueFilter, because it needs to be in a message? Makes the API more simple.

source-code/core2/src/messages/query.ts
```diff
@@ -1,15 +1,58 @@
...
-	get: () => Message[]
-	update: () => void
-	delete: () => void
+	create: (args: Message) => Message
+	get: (args: { where: UniqueFilterType }) => [Message, Error]
+	getMany: (args: { where: FilterType }) => [Message[], Error]
+	update: (args: Message) => Message
+	updateMany: (args: Message[]) => [Message[], Error]
+	upsert: (args: Message) => [Message, Error]
+	upsertMany: (args: Message[]) => [Message[], Error]
```
NilsJacobsen, line 23: From call with @NiklasBuchfink this will be a super convenient function for the editor. Just put the local changes (array of messages) in the function.

samuelstroschein: Yes, agree!

samuelstroschein (changes requested):

source-code/core2/src/messages/query.ts
```diff
@@ -28,8 +27,8 @@ const inlang = createInlang({
 	env: { fs: undefined as any, import: undefined as any },
 })
 
-// example usage
-const message1 = inlang.messages.query.create({
+// create
+const message1 = inlang.messages.query.upsert({
```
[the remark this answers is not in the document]
NilsJacobsen: Hmm I think, when we want to keep create we need update, too?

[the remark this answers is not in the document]
samuelstroschein: > Hmm I think, when we want to keep create we need update, too?

Yes, why would you want to remove update?

source-code/core2/src/messages/query.ts
```diff
@@ -14,7 +14,6 @@ type FilterType = {
 
 // inspiration: Prisma CRUD https://www.prisma.io/docs/concepts/components/prisma-client/crud
 export type MessagesQueryApi_1 = {
-	create: (args: Message) => Message
 	get: (args: { where: UniqueFilterType }) => [Message, Error]
```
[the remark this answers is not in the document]
samuelstroschein: Ah nvm. Result with generic error should be okayish.

NiklasBuchfink (approved):

source-code/core2/src/messages/query.ts
```diff
@@ -1,15 +1,54 @@
...
+})
+
+// create
+const message1 = inlang.messages.query.upsert({
+	id: "myMessageId",
+	languageTag: "en",
+	pattern: [{ type: "Text", value: "Hello World" }],
+})
+
+const message2 = inlang.messages.query.get({ where: { id: "myMessageId", languageTag: "en" } })
```
ivanhofer, line 38: can we get rid of `.query`?
`inlang.messages.get(/* ... */)`

NilsJacobsen: We discussed having a query2 in the future, which can be easily added to this setup to introduce reactivity or caching.

samuelstroschein: I vote for `inlang.messages.query.*`

The proposal would add 10 functions to the messages object. If we decide to add query unrelated props, the auto complete experience will be degraded because query functions mix with other properties. 

FYI we can easily move `messages.query` props to `messages` if its so painful. We can't easily move `messages` functions to `messages.query` without having an endless list of deprecated props on autocompletion.

felixhaeberle: Fair enough 👍

samuelstroschein: After https://github.com/inlang/inlang/issues/1159, I am 100% convinces `messages.query.*` is the way to go to avoid breaking changes. Changing the query API because new requirements emerge can break the entire ecosystem.

NilsJacobsen:
@felixhaeberle can you sum up the points where you see optimization from your API, or are you convinced with this?

felixhaeberle:
@NilsJacobsen We will probably come to a point where we realize that we don't need that many functions to handle data, but for now, I'm willing to **agree with what the majority is thinking** about this. As these are only inlang-apps-facing API's at the moment, we can change them at a later point anyway.

felixhaeberle (approved):

samuelstroschein:
But... the problem with #1159 remains. The query API depends on #1159

NiklasBuchfink:
There are use cases for `createMany`:
- automatic machine translate
- automatic message extraction

samuelstroschein:
> @felixhaeberle too many apis
> @NiklasBuchfink  There are use cases for createMany:

I agree with @felixhaeberle. We don't need `createMany` because we can simply iterate: 

_From my not yet proposed query API https://github.com/inlang/inlang/blob/fbc9ae989250e9d8cef914dfcf660073b1130704/source-code/core2/src/messages/query.ts_

```ts
for (const languageTag of ["de", "en", "fr", "es"]) {
	inlang.messages.query.delete({ where: { id: "myMessageId", languageTag } })
}
```

samuelstroschein:
closing in favor of https://github.com/inlang/inlang/pull/1171

PR #1158, opened 2023-07-21
[api-spec] query api draft felix

felixhaeberle:
hi folks, this is my idea of a dead simple API.

this PR focuses on the return type to always be an array of messages (empty, one message, multiple messages) and to have the least amount of functions we need.

I also see the value in having more / distinct functions, as it's currently discussed in other PR's but this idea really wants to focus on the basics (we can also make the API bigger afterward).

I skipped the error/exception handling for now, but we need that in this approach because queries could fail if not executed correctly with the right properties. -> ivan corrected me https://github.com/inlang/inlang/pull/1158#issuecomment-1645586360

### DIFF
--- a/source-code/core2/src/messages/query.ts
+++ b/source-code/core2/src/messages/query.ts
@@ -1,47 +1,51 @@
+import { createInlang } from "../app/createInlang.js"
 import type { Message } from "./schema.js"
-const inlang: any = {}
-
 /* --------------------- TYPES --------------------- */
 
-type UniqueFilterType = {
+type FilterType = {
 	id: string
 	languageTag: string
 	// more properties in the future
 }
 
 type QueryOptions = {
-	where?: Partial<UniqueFilterType>[]
+	where: Partial<FilterType>[]
 }
 
-export type MessagesQuery = {
-	get: (args: QueryOptions) => Message[]
+export type MessagesQueryApi_1 = {
+	get: (args?: QueryOptions) => Message[]
 	update: (data: Message[]) => Message[]
 }
 
 /* --------------------- CRUD --------------------- */
 
+const inlang = createInlang({
+	configPath: "/hello/inlang.config.json",
+	env: { fs: undefined as any, import: undefined as any },
+})
+
 // - create
 const message = inlang.messages.update([
 	{ id: "myMessageId", languageTag: "en", pattern: [{ type: "Text", value: "Hello World" }] },
-]) as ReturnType<MessagesQuery["update"]>
+])
 
 // - get
 const messageOrMessages = inlang.messages.get({
-	where: { id: "myMessageId", languageTag: "en" },
-}) as ReturnType<MessagesQuery["get"]>
+	where: [{ id: "myMessageId", languageTag: "en" }],
+})
 
 // - update
 const messageOrMessagesUpdated = inlang.messages.update([
 	{ id: "myMessageId", languageTag: "en", pattern: [{ type: "Text", value: "Hello World" }] },
-]) as ReturnType<MessagesQuery["update"]>
+])
 
 // - delete
-const messageDeleted = inlang.messages.update({
-	where: { id: "myMessageId", languageTag: "en", pattern: undefined },
-}) as ReturnType<MessagesQuery["update"]>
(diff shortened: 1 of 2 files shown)


ivanhofer:
> I skipped the error/exception handling for now, but we need that in this approach because queries could fail if not executed correctly with the right properties.

I think qeuries should not fail.. they could return an empty array if the input is invalid?

source-code/core2/src/messages/query.ts
```diff
@@ -1,15 +1,48 @@
...
+const messageOrMessages = inlang.messages.get({
+	where: { id: "myMessageId", languageTag: "en" },
+}) as ReturnType<MessagesQuery["get"]>
+
+// - update
+const messageOrMessagesUpdated = inlang.messages.update([
+	{ id: "myMessageId", languageTag: "en", pattern: [{ type: "Text", value: "Hello World" }] },
+]) as ReturnType<MessagesQuery["update"]>
+
+// - delete
```
NiklasBuchfink, line 39: I don't think that update to an undefined pattern is the same as a dedicated delete. This leads to confusion.

source-code/core2/src/messages/query.ts
```diff
@@ -1,15 +1,48 @@
...
-// query.
+export type MessagesQuery = {
+	get: (args: QueryOptions) => Message[]
+	update: (data: Message[]) => Message[]
+	list: () => Message[]
+}
+
+/* --------------------- CRUD --------------------- */
+
+// - create
```
NiklasBuchfink, line 24: So the `update` function is basically an `upsert`, right?

source-code/core2/src/messages/query.ts
felixhaeberle: I think having to decide and flesh out the cases between 5+ crud functions in apps are the bigger burden as passing `undefined` to one function and let core handle the rest.

samuelstroschein:
@felixhaeberle the types are all `any`. I thought of a PR as implementing the API for the query. 

We don't know whether your proposal works from a type perspective.

felixhaeberle:
@samuelstroschein fixed the types 👍

source-code/core2/src/messages/query.ts
```diff
@@ -1,15 +1,51 @@
...
+/* --------------------- TYPES --------------------- */
+
+type FilterType = {
+	id: string
+	languageTag: string
+	// more properties in the future
+}
+
+type QueryOptions = {
+	where: Partial<FilterType>[]
```
ivanhofer, line 12: Why an array?
Why a nested object with a single prop `where`?

samuelstroschein:
I think @NilsJacobsen proposal is better :P 

Close this one?

felixhaeberle:
@samuelstroschein No I think it's not, please argument.

samuelstroschein:
> @samuelstroschein No I think it's not, please argument.

Your API is simpler but will eventually break. 

1. Give CRUD ops. Not only `get` and `update` where `update` behaves like an `upsert`. Wtf?
2. Get should return a single message. Hence `get` and `getMany` is better. 

3. The benefit of your API, that you can query multiple messages at once can easily be ported to @NilsJacobsen API: 

```ts

inlang.messages.get({ where: { id: ["myMessage", "anotherMessage"] })
inlang.messages.delete({ where: { id: ["myMessage", "anotherMessage"] }

```

I'm on a call with Nils rn and we are both not sure on how to compare those approaches.

source-code/core2/src/messages/query.ts
ivanhofer: is this a use-case?

felixhaeberle:
> Give CRUD ops. Not only get and update where update behaves like an upsert. Wtf?

Fair enough, should have been `get` & `set` 

> Get should return a single message. Hence get and getMany is better.

The goal of this PR was to have a return type of `Message[]` all the time, you are right, there could be situations where I only want the object – without wrapped in an array. 

> The benefit of your API, that you can query multiple messages at once can easily be ported to @NilsJacobsen API:

Yep 👍

samuelstroschein:
Closing in favor of https://github.com/inlang/inlang/pull/1171

PR #1163, opened 2023-07-23
[api spec] make plugin.meta json serializable for future proofing

samuelstroschein:
See https://github.com/inlang/inlang/pull/1163#issuecomment-1646879127

### DIFF
--- a/source-code/core2/src/plugin/api.ts
+++ b/source-code/core2/src/plugin/api.ts
@@ -58,30 +58,3 @@ export const examplePlugin: Plugin<PluginOptions> = {
 		console.log(messages)
 	},
 }
-
-export const examplePlugin2NotWorking: Plugin<PluginOptions> = {
-	id: "inlang.plugin-i18next",
-	displayName: "i18next",
-	description: "i18next plugin for inlang",
-	keywords: ["i18next", "react", "nextjs"],
-	setup: ({ options, inlang }) => {
-		//! The following line will crash any static access to plugin functions
-		if (options.pathPattern === undefined) {
-			throw Error("Path pattern is undefined")
-		}
-		return {
-			extendLanguageTags: () => {
-				return ["en-US"]
-			},
-			loadMessages: () => {
-				for (const languageTag of inlang.config.languageTags) {
-					console.log(languageTag + options.pathPattern)
-				}
-				return []
-			},
-			saveMessages: ({ messages }) => {
-				console.log(messages)
-			},
-		}
-	},
-}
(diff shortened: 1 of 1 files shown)


samuelstroschein:
BTW The code beneath becomes a one-liner if we led plugins directly provide keywords etc. 

https://github.com/inlang/inlang/blob/3816d9859e7a1c5997be3b3dd217132495aec2c5/source-code/plugin-registry/src/registry.ts#L3-L11

```ts
export const plugins = [
  "https://cdn.jsdelivr.net/npm/@inlang/plugin-i18next/dist/index.js",
]
```

The plugin registry can import the module and directly access the metadata. 

```ts
const plugin = await import("https://cdn.jsdelivr.net/npm/@inlang/plugin-i18next/dist/index.js")

plugin.id
plugin.displayName
plugin.description
plugin.keywords

const definedApis = Obejct.keys(plugin.filter((key) => typeof plugin[key] === "function")
```

felixhaeberle:
My main concern is that separation of concerns with plugin "informational data" and plugin logic should be preserved and atm, we are only exposing logic out of the plugin what currently "hides" this problem — although we already have a plugin registry which duplicates?/populates information.

VSCode for example rely's on object properties to be set in `package.json`. This is nice because it's just JSON, can be universally be used and follows (js) ecosystem standards. Having this data in **js** which had to be initialized gives a future burden of having to initialize/parse the `js` to read from it.

**TLDR**: Why inlang.config.json now in json format but **leave informational data for plugins in js**, we should have learned from the config flaw that this doesn't hold for long. 

> But: I could be wrong on that, just a concern which came through my mind while reading this, as you mentioned, when used in js env, this is no problem because we can use `import`. In the future, inlang could/should provide plugin data to systems not having the ability to execute `js`?

samuelstroschein:
@felixhaeberle FYI Having plugin metadata in the JS object is off-topic because that's already decided. Marking all three comments as off-topic to not derail this proposal which is about static access to callbacks. 

> VSCode for example rely's on object properties to be set in package.json. This is nice because it's just JSON, can be universally be used and follows (js) ecosystem standards. Having this data in js which had to be initialized gives a future burden of having to initialize/parse the js to read from it.

But that's the beauty! No package.json/manifest required. From a dev perspective, no difference! We only need it for the plugin registry btw.

```ts

// import the first plugin for example purposes
const {default: plugin} = await import(inlang.config.plugins[0].module) 

plugin.id
plugin.description
plugin.loadMessages()
```

The alternative is requiring developers to write a plugin manifest that we _somehow_ need to resolve. 

```ts

// import the first plugin for example purposes
const { default: plugin } = await import(inlang.config.plugins[0].module) 

// making up a metadata resolving API here but put illustrated. We add unnecessary concepts.
const metadata = inlang.config.plugins[0].module.replace(".js", ".metadata.json")

metadata.id
metadata.description
plugin.loadMessages()
```

samuelstroschein:
Update after a call with @felixhaeberle: 

- we might want to have a separate plugin manifest _in the future_
- thus, `meta` must be json serializable
- hence, proposal 2 is the way to go aka nothing changed from the current API design except that `usedApis` is added.

PR #1164, opened 2023-07-24
[lisa] initial client api spec

samuelstroschein:
## Changes

Minimal API for a lisa client. 

Watch this [loom](https://www.loom.com/share/925a7dff15ac4f4282a10a8b321dd371?sid=7b13f91f-600b-4b58-9039-eebeae59bddb). 

- we can change the API often as long as we don't have external devs using lisa 
- reactivity for the client is likely the same issue inlang has with #1122 
- server API not drafted because the client is the API (the server must do whatever to please the client API)

#### Related discussions

- https://github.com/inlang/inlang/issues/915
- [git SDK requirements RFC](https://github.com/inlang/inlang/blob/main/rfcs/git-sdk-requirements/RFC.md)

#### Follow up tasks

- [ ] we should rename lisa to something where domains are available

CleanShot 2023-07-25 at 09 46 09@2x


samuelstroschein:
@inlang/editor you guys worked with lisa. Is the API spec in https://github.com/inlang/inlang/pull/1164/files#diff-93e23aaa45c2b051b15409307981ee12b16db3453964d7c5e8897e8214a71d5c sound?

NiklasBuchfink (approved):
LGTM, very straight forward 👍

NilsJacobsen:
Hey, do you think you could pop that graph in here? It'd be super helpful to have a visual to go along with everything. @samuelstroschein 

👍 I think the API looks super simple, we should try to keep it that way. :D

NilsJacobsen (approved):

felixhaeberle (approved):
@samuelstroschein This is definitely only a fraction of what we can foresee as of now, but yes, talking about those very limited set of functions: they provide the most basic functionalities.

PR #1171, opened 2023-07-24
[api spec] query v3

samuelstroschein:
## Context

**Pros**

- uniqueness via message id because language tag and (future) variants are nested
- fast language tag retrieval via `message.body['de'].pattern`
- extendable with variants in the future 

**Cons**

- more complicated (but required) type 

## Additional information 

**- we agreed on a call to go with this API, start implementing and see how far we get**
- `Pattern` must be a unique AST node because a `Pattern` can nest other patterns for pluralization

### DIFF
--- a/source-code/core2/src/fullExample.ts
+++ b/source-code/core2/src/fullExample.ts
@@ -1,6 +1,6 @@
 import { createInlang } from "./app/createInlang.js"
 import type { InlangConfig } from "./config/schema.js"
-import type {  MessageLintRule } from "./lint/api.js"
+import type { MessageLintRule } from "./lint/api.js"
 import type { Plugin_Proposal_2 } from "./plugin/api.js"
 
 // --------------------- LINT RULE ---------------------
@@ -13,11 +13,8 @@ const missingMessage: MessageLintRule = {
 	},
 	defaultLevel: "error",
 	message: ({ message, inlang, report }) => {
-		if (message.languageTag !== inlang.config.sourceLanguageTag) {
-			return
-		}
 		for (const languageTag of inlang.config.languageTags) {
-			const translation = inlang.messages.query.get({ where: { id: message.id, languageTag } })
+			const translation = message.body[languageTag]
 			if (!translation) {
 				report({
 					languageTag,
@@ -36,7 +33,7 @@ export const myPlugin: Plugin_Proposal_2<{ pathPattern: string }> = {
 		id: "inlang.myPlugin",
 		displayName: "My Plugin",
 	},
-	setup: ({ inlang, options }) => {
+	setup: () => {
 		return {
 			extendLanguageTags: () => {
 				return ["de"]
@@ -59,11 +56,11 @@ const exampleInlangConfig: InlangConfig = {
 	plugins: [
 		{ module: "https://example.com/myPlugin.js", options: { pathPattern: "src/**/*.{ts,tsx}" } },
 	],
-  lint: {
-    rules: {
-      "inlang.missingMessage": "warning"
-    }
-  }
+	lint: {
+		rules: {
+			"inlang.missingMessage": "warning",
+		},
+	},
 }
 
 // 1. Create the app instance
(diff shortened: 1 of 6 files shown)


NiklasBuchfink:
@samuelstroschein What was the reason for omitting the `upsert` function?

samuelstroschein:
@NiklasBuchfink re-added upsert

NiklasBuchfink (approved):

source-code/core2/src/fullExample.ts
```diff
@@ -74,24 +71,42 @@ const inlang = createInlang({
 })
 
 // --- CRUD ---
-inlang.messages.query.create({ id: "myMessageId", languageTag: "en", pattern: [{"type": "Text", value:  "Hello World" }]})
+inlang.messages.query.create({
```
ivanhofer, line 74: is it just me or does the word `query` look odd when doing a mutation?

source-code/core2/src/fullExample.ts
```diff
@@ -74,24 +71,42 @@ const inlang = createInlang({
...
+	data: {
```
ivanhofer, line 75: why the `data` prop?

source-code/core2/tsconfig.json
```diff
@@ -5,7 +5,7 @@
 		// DOM because of fetch. TODO: extract fetch from this package and
 		// let fetch be provided by the environment.
 		"lib": ["ESNext", "DOM"],
-		"types": [],
+		"types": ["node"],
```
ivanhofer, line 8: 🤔

samuelstroschein: To use the `node:test` runnner. Instead of vitest.

source-code/core2/src/fullExample.ts
NilsJacobsen: inspired by https://www.prisma.io/docs/concepts/components/prisma-client/crud#update

source-code/core2/src/messages/query.ts
```diff
@@ -1,15 +1,41 @@
...
-//
-// inspiration: Prisma CRUD https://www.prisma.io/docs/concepts/components/prisma-client/crud
-export type MessagesQueryApi_1 = {
-	create: (message: Message) => void
-	get: () => Message[]
-	update: () => void
-	delete: () => void
+export type MessageQueryApi = {
+	create: (args: { data: Message }) => void
+	get: (args: { where: { id: Message["id"] } }) => Message | undefined
```
NilsJacobsen, line 5: How can we access all messages? Are we in the future adding filters to the get query to return, for example, only one variant or only one language? To achieve something similar from outside, we need at least access to all the messages.

source-code/core2/src/messages/query.ts
```diff
@@ -1,15 +1,41 @@
...
+	update: (args: { where: { id: Message["id"] }; data: Partial<Message> }) => void
+	upsert: (args: { where: { id: Message["id"] }; data: Message }) => void
```
NilsJacobsen, line 7: We decided not to do `many` because we can loop from outside, right?

source-code/core2/src/messages/query.test.ts
```diff
@@ -0,0 +1,77 @@
...
+
+		assert.equal(message, mockMessage)
+	})
+})
+
+describe("get", () => {
+	const query = createQuery([])
+
+	it("should return undefined if a message does not exist", () => {
+		const message = query.get({ where: { id: "none-existent-message" } })
```
NilsJacobsen, line 25: Nice to get a unique message finally.

source-code/core2/tsconfig.json
ivanhofer: Is it also possible to just include `node:test`?

samuelstroschein: Nope :(

The linting rules of don't use "node" are still applying though.

source-code/core2/src/messages/query.ts
NilsJacobsen: We gonna discuss an `includedMessageIds`function in a follow-up PR for accessing all the messages.

NilsJacobsen (approved):

samuelstroschein:
@felixhaeberle approve? then we can start refactoring after the reactivity lib is choosen

felixhaeberle (approved):

source-code/core2/src/fullExample.ts
felixhaeberle: @samuelstroschein ?

`inlang.message.query.get()`
`inlang.message.mutation.create()`

PR #1172, opened 2023-07-24
remove `where`

felixhaeberle:
hi folks,

this is a PR which removes `where`because I think this is not needed anymore.

do you agree? :+1: / :-1:

### DIFF
--- a/source-code/core2/src/fullExample.ts
+++ b/source-code/core2/src/fullExample.ts
@@ -81,7 +81,7 @@ inlang.messages.query.create({
 })
 
 // assuming that get is reactive
-const message = inlang.messages.query.get({ where: { id: "myMessageId" } })!
+const message = inlang.messages.query.get({ id: "myMessageId" })!
 
 message.body["en-US"] = {
 	pattern: {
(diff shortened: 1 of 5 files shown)


source-code/core2/src/fullExample.ts
```diff
@@ -96,11 +96,11 @@ message.body["en-US"] = {
 }
 
 inlang.messages.query.update({
-	where: { id: "myMessageId" },
+	id: "myMessageId",
 	data: message,
```
ivanhofer, line 100: why do we need a separate `data` prop?

ivanhofer (approved):

source-code/core2/src/fullExample.ts
samuelstroschein: mixing `id` with `data` would likely become confusing, don't you think?

samuelstroschein (approved):

source-code/core2/src/fullExample.ts
ivanhofer:
```ts
inlang.messages.query.update({
	id: "myMessageId",
	body: [/* */],
})
```
?

samuelstroschein (commented):
The `where` clause exists to update the query API over time. 

```ts
const messages = query({ where: { body: { pattern: {  includes: "hello-world" })
```

Hence, accepting this proposal makes current query requirements simpler, potential future requirements difficult. I go with the majority vote.

felixhaeberle:
This could be a simple filter in the app? Most apps will need to filter in some sort anyway.

NilsJacobsen:
I feel like removing these filter options just makes it harder in the future. The tradeoff is small.

```ts
const messages = query({ where: { body: { pattern: {  includes: "hello-world" })
```

felixhaeberle:
I'll close this. Leaving it in gives the possibility for further advanced use cases.

PR #1174, opened 2023-07-25
stricter lint rules

### LINKED ISSUE #1168: stricter lint rules
### Discussed in https://github.com/inlang/inlang/discussions/1162

<div type='discussions-op-text'>

<sup>Originally posted by **ivanhofer** July 22, 2023</sup>
I want to share a few issues with the current code base:

- it should not be possible to build an appliaction if there are `console.log`s in the code ([see issue](https://github.com/inlang/inlang/issues/1160))
  `console.log` are mosten used for debugging and often forgotten to remove. If there really is the case where we want to log something we can use `console.info` for that.
 - enable formatting in CI. I sometimes format the code and I always need to revert the canges from packages I don't touch as I only care about my packages. Often it involves code I had written a few days or weeks ago for another PR so I have changes in my PR that should not be there. We once had a job that failed when the code was not formatted well, but we removed it. We should add a new CI job that formats the code and commits the canges to the branch instead of failing.
 - someone always forgets to run `npm i`. It happens to often that I branch of from main, run `npm i` and have changes in `package_lock.json`. We should enforce that `package_lock.json` must be up2date.</div>

ivanhofer:
closes #1168

 - `console.log` is now a lint error
 - the `lint` CI actions will now
    - run `npm i` and modifies the `package_lock.json` file if not up2date
    - auto-fix lint issues
    - format the whole code base
    - commit all changes

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -28,6 +28,38 @@
 		"unicorn/prefer-node-protocol": "error",
 		"@typescript-eslint/ban-ts-comment": "off",
 		"@typescript-eslint/no-explicit-any": "off",
+		"no-console": [
+			"error",
+			{
+				// allow everything except `console.log`
+				"allow": [
+					"debug",
+					"error",
+					"info",
+					"warn",
+					"dir",
+					"dirxml",
+					"table",
+					"trace",
+					"group",
+					"groupCollapsed",
+					"groupEnd",
+					"clear",
+					"count",
+					"countReset",
+					"assert",
+					"profile",
+					"profileEnd",
+					"time",
+					"timeLog",
+					"timeEnd",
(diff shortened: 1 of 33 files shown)


openscript:
I've had the situation, that I formatted all the code with Visual Studio Code, but still failing the linting and format checks. The reason for this is, that Visual Studio Code is directly using Prettier as a formatter, whereas `npm run lint` is using ESLint and through it Prettier, which leads to other formatting rules.

Not setting any `"editor.defaultFormatter": "esbenp.prettier-vscode"` inside `settings.json`, but `"eslint.format.enable": true,` should fix the problem.

ivanhofer:
@openscript thanks, I have updated the settigns

PR #1179, opened 2023-07-26
fix: tableOfContents not found

floriandwt:
resolving the docs error in production

### DIFF
--- a/documentation/tableOfContents.ts
+++ b/documentation/tableOfContents.ts
@@ -6,63 +6,80 @@ import { RequiredFrontmatter } from "@inlang/website/markdown"
 export const FrontmatterSchema = RequiredFrontmatter
 
 /**
- * Function to dynamically import a markdown file given its path.
- */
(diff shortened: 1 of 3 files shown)


documentation/tableOfContents.ts
```diff
@@ -1,69 +1,85 @@
...
-			}),
-		)
-
-		for (const content of contents) {
-			tableOfContents.push({ category, content })
-		}
-	}
-
-	return tableOfContents
+export const tableOfContents: Record<string, { raw: string[]; import: string[] }> = {
```
NilsJacobsen, line 11: I feel like for better maintainability, we could structure it like:

```ts
tableOldRecords = {
   "getting started": [
      {
         raw: "",
         import: (await import("")) 
      },
      {
         raw: "",
         import: (await import("")) 
      }
   ]
}
```

NilsJacobsen (changes requested):
See comment

PR #1186, opened 2023-07-28
fix: exclude meta tags from not being loaded

floriandwt:
fixed the issue that loading the locales prevented the meta tags from being loaded directly, which is necessary to let search engines read our SEO content.

Todo:
- Fix error in /editor page that appears because of the change in Root


source-code/website/src/pages/Layout.tsx
```diff
@@ -45,16 +46,18 @@ export function Layout(props: { children: JSXElement }) {
 
 export const LandingPageLayout = (props: { children: JSXElement; landingpage?: boolean }) => {
 	return (
-		<div class="flex flex-col min-h-screen">
-			<Header landingpage={props.landingpage} />
-			{/* the outer div is growing to occupy the entire height and thereby
+		<Show when={localesLoaded()}>
```
samuelstroschein, line 49: But... now nothing is rendered when the languages didn't load?

NilsJacobsen (approved):

samuelstroschein (approved):

PR #1187, opened 2023-07-28
 Implement remote-objectStoreFs to improve cloning in large repos

araknast:
This code implements `mappedObjectStore`, a filesystem-like interface to the git object store, which removes the need to checkout files onto the host filesystem.

This code also implements a `createRemoteMappedObjectStore` function, which augments a `mappedObjectStore` interface,  allowing for objects to be lazily fetched from a remote repository as needed, circumventing the need for a git clone. 

Finally, there is `objectStoreFs`, which operates on a `mappedObjectStore` interface and provides a nodeish filesystem api.

I've gone ahead and implemented a remote`mappedObjectStore` interface in the editor in this branch. This is  is currently a bit slower than the regular `mappedObjectStore` due to certain inefficiencies in `core`, mainly lack of parallelization. Once the refactor of `core` is finished this should produce significant performance improvements in the editor, specifically in cloning.


samuelstroschein:
This PR removes the need for checkout(s) correct? Not, lazy fetching of files as lazy fetching requires a server that we do not have yet.

araknast:
> This PR removes the need for checkout(s) correct? Not, lazy fetching of files as lazy fetching requires a server that we do not have yet.

This PR does both. `objectStoreFs` which removes the need for checkout, and `createRemoteMappedObjectStore` which allows `objectStoreFs` to lazy fetch files.

PR #1192, opened 2023-07-31
feat: getLocale function as helper

floriandwt:
added:
- getLocale() function as a reusable function that gets 2 arguments from the specific page

@NilsJacobsen unsure about the newly created folder, there might be a more elegant solution for the function or I've overseen a suitable folder for that. :D Therefore creating it as draft PR.


samuelstroschein:
stale/close? @floriandwt

floriandwt:
Closing, I will implement the helper function after the refactoring is done.

PR #1193, opened 2023-08-01
style(website): optimize the format under Chinese translation

normal-coder:
## What does this Pull Request do

Optimized the format of `source-code/website/lang/zh.json` (basically adjusting spaces), making website browsing and reading more comfortable.

## Why do this

In the Chinese context, the best practice of content layout follows the following principles:

- When Chinese and English are mixed (such as with technical terms or specific English terms), spaces need to be added between Chinese and English
- Spaces need to be added between Chinese characters and numbers
- Spaces need to be added between numbers and units
- No spaces between full-width punctuation and other characters

Based on the above principles, the Chinese translation of Website has been improved.

Thanks: This improvement is done with the help of the VSCode plugin [AutoCorrect](https://github.com/huacnlee/autocorrect)

### DIFF
--- a/source-code/website/lang/zh.json
+++ b/source-code/website/lang/zh.json
@@ -5,17 +5,17 @@
 			"title": "软件基础设施",
 			"cta": "开始使用",
 			"description": {
-				"inlangs": "inlang的",
+				"inlangs": "inlang 的",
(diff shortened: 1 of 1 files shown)


samuelstroschein:
This is great! Thank you @normal-coder

PR #1211, opened 2023-08-04
Translate README.md into README_ZH-CN.md in Chinese

normal-coder:
### What does this Pull Request do

- Translate `README.md` into `README_ZH-CN.md` in Chinese
- Optimized the Chinese expression of Website, including but not limited to:
  - Replace `您` with `你` (reduce the polite language and draw the user closer)
  - Representation of improved properties
- Corrected the expression of special terms (`CLI` should be expressed as a `command-line tool` instead of a `command-line interface` in Chinese)


samuelstroschein:
Thanks for the contribution @normal-coder !

PR #1218, opened 2023-08-08
update vite-plugin-ssr and update import path

lucidNTR:
update vite plugin ssr to get rid of warnings and as preparation for the depracation of RenderErrorPage


samuelstroschein (approved):
Yeah rly tiny :D looks good

PR #1219, opened 2023-08-08
Lisa client

lucidNTR:
First version of lisa client as minimal wrapper for github api and isomorphic git

### DIFF
--- a/project-lisa/source-code/client/src/api.ts
+++ b/project-lisa/source-code/client/src/api.ts
@@ -17,10 +17,10 @@ import { wrap, transformRemote } from "./helpers.js"
 export const raw = _git
 
 /**
-  * The http client for the git raw api. (Used for legacy migrations)
-  *
-  * Note: The http client is web-based. Node version 18 is
-  * required for the http client to work in node environments.
+	* The http client for the git raw api. (Used for legacy migrations)
+	*
+	* Note: The http client is web-based. Node version 18 is
+	* required for the http client to work in node environments.
 */
 export const http = _http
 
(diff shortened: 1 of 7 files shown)


project-lisa/source-code/client/src/github.ts
```diff
@@ -0,0 +1,27 @@
+import { Octokit } from "octokit"
+
+/**
+ * The path used for proxy purposes.
+ *
+ * The server will proxy all requests to this path to the github API.
+ */
+export const PATH = "/github-proxy/"
```
samuelstroschein, line 8: Not needed anymore. The github proxy is also just lisa/server.

project-lisa/source-code/client/src/helpers.ts
```diff
@@ -0,0 +1,45 @@
+import type { NodeishFilesystem } from "@inlang-git/fs"
+
+/**
+* Wraps a nodeishFs implementation with a js proxy for detailed logging, debugging and transparently replacing the file access behaviour.
+* advantage of using this approach is that the underlying fs can also be swapped to something like lightingfs seamlessly.
+*/
+export const wrap = (targetRoot: NodeishFilesystem, _module: string, fn?: (args: { prop: keyof typeof targetRoot, execute: () => any }) => any ): NodeishFilesystem => {
```
samuelstroschein, line 7: can we find a better name than `wrap`?

wrap doesn't express why the fs is wrapped. `debug` maybe?

project-lisa/source-code/client/src/api.ts
```diff
@@ -1,24 +1,261 @@
...
+		corsProxy,
+		url: normalizedUrl,
+		singleBranch: true,
+		depth: 1,
+		noTags: true,
+	}).finally(() => {
+		pending = undefined
+	}).catch((err: any) => {
+		console.error('error cloning the repository', err)
+	})
```
samuelstroschein, line 85: Why don't you await the promise? 

If this is preparation for the lazy fetching, the correct solution is likely to ditch `raw.clone` entirely for a custom "clone" functionality that only fetches the minimum metadata required. Calls to `await fs.readFile()` fetch underlying contents of files lazily.

project-lisa/source-code/client/src/api.ts
```diff
@@ -1,24 +1,261 @@
...
+		 *
+		 * @returns The git origin url or undefined if it could not be found.
+		 */
+		async listRemotes () {
+			try {
+				const wrappedFS = wrap(rawFs, 'listRemotes', delayedAction)
+
+				const remotes = await raw.listRemotes({
+					fs: wrappedFS,
+					dir: await raw.findRoot({
```
samuelstroschein, line 110: Seems like we can extract `findRoot` to the scope of a repo. The concept of changing working directories does not exist in lisa/client.

project-lisa/source-code/client/src/api.ts
```diff
@@ -1,24 +1,261 @@
...
+			try {
+				response = await github.request(
+					"GET /repos/{owner}/{repo}/collaborators/{username}",
+					{
+						owner,
+						repo: repoName,
+						username,
+					},
+				)
+			} catch (_err) { /* throws on non collaborator access */ }
```
samuelstroschein, line 205: Needs to be extracted in the future and pushed into `createGitHubAdapter()` or a similar pattern. (We don't want lock-in to a specific git host. Users are using Azure, Gitlab, Bitbucket, etc.)

project-lisa/source-code/client/src/api.ts
```diff
@@ -1,24 +1,261 @@
...
+			return transformRemote(result)
+		},
+
+		async getCurrentBranch () {
+			// TODO: make stateless?, migrate to getMainBranch
+			return await raw.currentBranch({
+				fs: wrap(rawFs, 'getMainBranch', delayedAction),
+				dir: "/",
+			}) || undefined
+		},
```
samuelstroschein, line 237: What is `getMainBranch` for?

project-lisa/source-code/client/src/api.ts
```diff
@@ -1,24 +1,261 @@
...
-	status: () => any
+	fs: NodeishFilesystem
+	commit: (args: { dir: string, author: any, message: string}) => Promise<Awaited<ReturnType<typeof raw.commit>> | undefined>
+	push: (args: { dir: string }) => Promise<Awaited<ReturnType<typeof raw.push>> | undefined>
+	pull: (args: { dir: string, author: any, fastForward: boolean, singleBranch: true }) => any
+	add: (args: { dir: string, filepath: string }) =>  Promise<Awaited<ReturnType<typeof raw.add>>>
+	listRemotes: () => Promise<Awaited<ReturnType<typeof raw.listRemotes>> | undefined>
+	log: (args: { dir: string, since: any}) =>  Promise<Awaited<ReturnType<typeof raw.log>>>
+	statusMatrix: (args: { dir:string, filter: any }) => Promise<Awaited<ReturnType<typeof raw.statusMatrix>>>
+	status: (args: { dir: string, filepath: string }) => Promise<Awaited<ReturnType<typeof raw.status>>>
```
samuelstroschein, line 36: We can remove `dir` from every command. 

The `dir` is always (for now at least) the root of the repo.

samuelstroschein: On that note, `load` should take a `workingDirectory` argument in the future once we support the opening of local repositories. 

Wait a sec, does `open` make more sense than `load` as entry point for repositories? 

- we don't "load" the repo as in "downloading" it once lazy fetching is implemented
- "opening" a local repository makes more sense than "loading" a local repository

```ts
const repository = open("my-repo.com", options)
```

project-lisa/source-code/client/src/api.ts
```diff
@@ -1,24 +1,261 @@
...
+			fullName: string
+		} | undefined
+	}>
+
+	// TODO: implement these before publishing api, but not used in badge or editor
+	// currentBranch: () => unknown
+	// changeBranch: () => unknown
+}
+
+export function load ({ url, fs, corsProxy }: { url: string, fs?: NodeishFilesystem, path?: string, corsProxy?: string, auth?: unknown }): Repository {
```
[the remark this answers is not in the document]
samuelstroschein: Question, what looks nicer? 

```ts
const repository = await load("link.to/repository", { hostAdapter: createGitHubAdapter(), fs: createNodeishFs() })
```

or 

```ts
const repository = await load({ url: "link.to/repository", hostAdapter: createGitHubAdapter(), fs: createNodeishFs() })
```

project-lisa/source-code/client/src/api.ts
samuelstroschein: No action required. hardcoded is fine for now

project-lisa/source-code/client/src/api.ts
samuelstroschein: No need to call findRoot for every listRemotes call. There is always only one root, the root of the repo.

project-lisa/source-code/client/src/api.ts
```diff
@@ -1,24 +1,261 @@
...
+
+	// parse url in the format of github.com/inlang/example and split it to host, owner and repo
+	const [host, owner, repoName] = [...url.split("/")]
+
+	// check if all 3 parts are present, if not, return an error
+	if (!host || !owner || !repoName) {
+		throw new Error(
+			`Invalid url format for '${url}' for cloning repository, please use the format of github.com/inlang/example.`,
+		)
+	}
```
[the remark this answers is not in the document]
samuelstroschein: No action required. Just want you to be aware of being host agnostic. Hit resolve once read.

project-lisa/source-code/client/src/github.ts
samuelstroschein: Reverting my remark: Using `load('", { corsProxy: "link.to/proxy"})` is the API that will be used. No manual export is required anymore.

project-lisa/source-code/client/src/api.ts
```diff
@@ -1,24 +1,242 @@
...
+		} | undefined
+	}>
+
+	// TODO: implement these before publishing api, but not used in badge or editor
+	// currentBranch: () => unknown
+	// changeBranch: () => unknown
+}
+
+export function open (url: string, args?: { nodeishFs?: NodeishFilesystem, workingDirectory?: string, corsProxy?: string, auth?: unknown }): Repository {
+	const rawFs = args?.nodeishFs ? args.nodeishFs : createNodeishMemoryFs()
```
samuelstroschein, line 39: I would not create a memory fs under the hood. let users explicitly choose the fs they want to use. (also makes `args` not optional.)

samuelstroschein (changes requested):
Looks good after making args non optional in `load` and letting users explicitly pass nodeishFs

PR #1227, opened 2023-08-09
chore: add e2e testing to ide-extension

openscript:


samuelstroschein:
@openscript i would prefer if you can spent the time updating the ide extension to the new inlang API. 

Merge this, then update the ide extension to the new inlang API in `armageddon`?

PR #1230, opened 2023-08-09
Badge lisa client

lucidNTR:
migrate the badge service to use the new lisa client api


source-code/badge/src/badge.ts
```diff
@@ -28,21 +26,20 @@ export const badge = async (url: string) => {
...
 	}
 
-	// initialize a new file system on each request to prevent cross request pollution
-	const fs = createNodeishMemoryFs()
-	await cloneRespository(url, fs)
+	// initialize a lisa repo instance on each request to prevent cross request pollution
+	const repo = await open(url, { nodeishFs: createNodeishMemoryFs() })
 
 	// Set up the environment functions
 	const env: InlangEnvironment = {
```
samuelstroschein, line 33: InlangEenvironment shouldn't be needed anymore. But, we can fix that in a follow up commit.

samuelstroschein (approved):

PR #1238, opened 2023-08-11
CLI: Fix sdk-js-plugin getting updated to "0"

WarningImHack3r:
Pretty self-explanatory, the `npx @inlang/cli config update` command upgrades `sdk-js-plugin` to `"0"` because it only returns the major version.
Now this plugin is excluded from this mechanism and gets the full version returned.

I couldn't test that because I don't have access to multiple dependencies required to run it, but I don't see why it wouldn't work.

Also, I don't know if it's the right branch to PR to, nor if it should be added as a test case in a `.test.ts` file.


ivanhofer:
@samuelstroschein I think we should exclude the `sdk-js-plugin` from the auto-update command as the version gets managed by the installed version of the `@inlang/sdk-js` npm package.

samuelstroschein:
The command has been remove in the latest releases of inlang.

PR #1250, opened 2023-08-15
Fix reactivity for create reactive query

lucidNTR:


ivanhofer:
Awaiting `query.create` seems to make the test successfull even if `query.create` is not an async function.. This does not look like a reliable solution to me 😅

lucidNTR:
fixed with patching solid, the source of solution that ivan found was that solid reactivemap only works in the browser but disables reactivity in the server import

PR #1264, opened 2023-08-18
implement debounced saving of changes applied with to reactive query

janfjohannes:


samuelstroschein (commented):
beautifully simple (if it works). keep in mind that `const newMessages` is not only new messages but all messages. future optimization could have some diffing (if required)

samuelstroschein (approved):

source-code/core/app/src/createInlang.ts
```diff
@@ -136,6 +137,23 @@ export const createInlang = async (args: {
 
 		const query = createReactiveQuery(() => messages()!)
 
+		let debouncedSave = (_newMessages) => {
+			// Skip initial call for setup
```
ivanhofer, line 141: I would move the comment near `atBegin`

source-code/core/app/src/createInlang.ts
```diff
@@ -136,6 +137,23 @@ export const createInlang = async (args: {
...
+			debouncedSave = debounce(500, async (newMessages) => {
+				// console.log('saving changes to messages')
+				try {
+					await resolvedModules()?.data.plugins.data.saveMessages({ messages: newMessages})
```
ivanhofer, line 145: You can use `!` instead of `?` here because we know that this cannot be `undefined` at this point in time.

source-code/core/app/src/createInlang.ts
```diff
@@ -136,6 +137,23 @@ export const createInlang = async (args: {
...
+				} catch (err) {
+					console.error("Error in saving messages ", err)
```
ivanhofer, line 147: Should we add this to the list of errors an app can show?
Please also use a dedicated class for the error.

source-code/core/app/src/createInlang.ts
```diff
@@ -8,6 +8,7 @@ import { ConfigPathNotFoundError, ConfigSyntaxError, InvalidConfigError } from "
 import { LintRuleThrowedError, LintReport, lintMessages } from "@inlang/lint"
 import { createRoot, createSignal, createEffect } from "./solid.js"
 import { createReactiveQuery } from "./createReactiveQuery.js"
+import { debounce } from "throttle-debounce"
```
ivanhofer, line 11: You don't have this package installed in `@inlang/app`

ivanhofer (approved):
Looks good!

source-code/core/app/src/createInlang.ts
samuelstroschein: Oh yes, @janfjohannes add it to the error array. We should ban console.* logs in @inlang/app.

source-code/core/app/src/createInlang.ts
janfjohannes: @ivanhofer   but it does not explain atBegin, but why i reassign to debouncedSave in the first call...

ivanhofer: Ah, now I see what you do here 😅
Nice trick, but it really is intransparent when not deeply thinking about it.. Can we find another solution?

source-code/core/app/src/errors.ts
```diff
@@ -18,3 +18,17 @@ export class ConfigPathNotFoundError extends Error {
 		this.name = "ConfigPathNotFoundError"
 	}
 }
+
+export class PluginSaveError extends Error {
+	constructor(message: string, options: ErrorOptions) {
+		super(message, options)
+		this.name = "PluginSaveError"
+	}
+}
```
samuelstroschein, line 27: I vote for more being more explicit like `PluginLoadMessagesError`, `PluginSaveMessagesError`

source-code/core/app/src/createInlang.ts
janfjohannes: ok what about now? its also not pretty until js has pipe operator but as simple as it gets.

samuelstroschein:
@janfjohannes can you add a test? should be straightforward to add

source-code/core/app/src/createInlang.ts
ivanhofer: way better!

samuelstroschein (approved):

source-code/core/app/src/createInlang.test.ts
```diff
@@ -7,6 +7,8 @@ import type { Message, Plugin } from "@inlang/plugin"
 import type { LintRule } from "@inlang/lint"
 import type { ImportFunction, InlangModule } from "@inlang/module"
 import { ConfigPathNotFoundError, ConfigSyntaxError, InvalidConfigError } from "./errors.js"
+import pluginJson from "../../../plugins/json/dist/index.js"
+import pluginLint from "../../../plugins/standard-lint-rules/dist/index.js"
```
samuelstroschein, line 11: Remove in favor of a mock plugin with a simple `saveMessages` function. 

- dependency on plugin json will lead to side effects

samuelstroschein (changes requested):

samuelstroschein (approved):
meeeerge

PR #1286, opened 2023-08-24
feat: inlang.com/install app

floriandwt:
work in progress


floriandwt:
@NiklasBuchfink did you provide modules in the query / from the marketplace? The registry has to be filled with relative paths to the modules for testing atm

NiklasBuchfink (approved):
I tested adding a plugin and the process was quite straightforward. Well done 👍

source-code/website/src/pages/install/InstallationProvider.tsx
```diff
@@ -0,0 +1,261 @@
...
+	setStep: (step: Step) => void
+	optIn: Record<string, any>
+	children: JSXElement
+}) {
+	const [localStorage, setLocalStorage] = useLocalStorage() ?? []
+	const user = localStorage?.user
+
+	/**
+	 * This function checks for common errors before repo initialization (to be more performant) and sets the step accordingly.
+	 */
```
samuelstroschein, line 27: @floriandwt if this functions checks for common errors **before** a repo initialization, why is this reactive und not a `validateRepo` function that is called once?

source-code/website/src/pages/marketplace/@id/index.page.tsx
```diff
@@ -0,0 +1,193 @@
...
+	/* Automatically sets the fitting item e.g. when changing bundle */
+	createEffect(() => {
+		setItem(
+			marketplaceItems.find(
+				(item) =>
+					item.meta.displayName.en?.toLowerCase().replaceAll(" ", "-") ===
+					props.markdown?.frontmatter?.title?.toLowerCase().replaceAll(" ", "-"),
+			),
+		)
+	})
```
samuelstroschein, line 36: @floriandwt use derived state. You don't need an effect for derived state. If you find yourself calling `set*` in an effect, you are likely using the effect wrong. 

Please change the code to used derived state

```ts
const item = () => marketplaceItems.find((item) => ...)
```

samuelstroschein (commented):
Looks OK. Two minor remarks.

Off-topic: We should extract install in the future to a separate package to avoid building an gigantic monolith.

source-code/website/src/pages/install/InstallationProvider.tsx
floriandwt: Shouldn't run once because the user opts in to the process and the app has to proceed based on the input

floriandwt:
@samuelstroschein I've edited the wording based on https://github.com/inlang/inlang/issues/1298 and implemented the feedback, am I free to merge?

> We should extract install in the future to a separate package to avoid building an gigantic monolith.

Definitely!

samuelstroschein (approved):
@floriandwt merge!

PR #1377, opened 2023-09-19
fix: wording in install logic

floriandwt:
Better wording for the input of the repository URL as proposed by @felixhaeberle


felixhaeberle (approved):
<3

PR #1385, opened 2023-09-20
feat: [plugin] isolate detect formatting

NilsJacobsen:
- added detect-formatting module
- added unit tests
- implemented it in json and i18next plugin


NilsJacobsen:

inlang/source-code/detect-formatting/package.json
```diff
@@ -0,0 +1,30 @@
+{
+	"name": "@inlang/detect-formatting",
```
samuelstroschein, line 2: @NilsJacobsen can we name it @inlang/detect-json-formatting for more explicitness also when people search for the package?

NilsJacobsen: Sure, you are right that feels more self-explaining.

PR #1398, opened 2023-09-21
add: project error log to machine translate cmd

NiklasBuchfink:
@felixhaeberle Feel free to merge.

**Changes:**
- add error logging in the machine translate command

**Suggestion:**
With almost all CLI commands, we have to handle errors at some point because otherwise, the actions will not work as expected. These should always be checked before executing the actual command.

### DIFF
--- a/.changeset/dry-hornets-boil.md
+++ b/.changeset/dry-hornets-boil.md
@@ -1,0 +1,5 @@
+---
+"@inlang/cli": minor
+---
+
+add better error handling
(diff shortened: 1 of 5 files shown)


felixhaeberle:
On a second thought, this error handling should be for all commands and therefore on a more granular (higher) level in the CLI

felixhaeberle:
@NiklasBuchfink I'll take over. 👍

felixhaeberle:
@NiklasBuchfink Improved error handling, merging now 👍

PR #1407, opened 2023-09-25
feat: install json formatting detection module

floriandwt:


floriandwt:
@NiklasBuchfink @NilsJacobsen The formatting did change when I tested it. Did I use the module wrong?
The keys of an array are still becoming ordered below each other instead of in a row, see here: https://github.com/floriandwt/inlang-ide-next-demo/commit/b877eb3f905602f0cb124b581cf40457954821d9

PR #1410, opened 2023-09-26
improve tailwind css with RTL compatible classes

felixhaeberle:
@inlang/team this PR aims to improve our current usage of tailwind css classes ("left" and "right") and replaces them with the "start" & "end" logical counterparts (called logical properties).

This change allows RTL users to have our UI displayed correctly & is no further maintenance burden for us. Please use whenever possible the correct "start" & "end" alternatives & ulimately ban "right" & "left" from your CSS box model thinking.

Thanksyy <3

More on logical properties & why we should use them: 
https://www.bram.us/2019/09/19/css-logical-properties-and-values-the-next-step-of-css-evolution/

### DIFF
--- a/inlang/source-code/end-to-end-tests/inlang-nextjs/app/page.tsx
+++ b/inlang/source-code/end-to-end-tests/inlang-nextjs/app/page.tsx
@@ -4,11 +4,11 @@ export default function Home() {
 	return (
 		<main className="flex min-h-screen flex-col items-center justify-between p-24">
 			<div className="z-10 w-full max-w-5xl items-center justify-between font-mono text-sm lg:flex">
-				<p className="fixed left-0 top-0 flex w-full justify-center border-b border-gray-300 bg-gradient-to-b from-zinc-200 pb-6 pt-8 backdrop-blur-2xl dark:border-neutral-800 dark:bg-zinc-800/30 dark:from-inherit lg:static lg:w-auto  lg:rounded-xl lg:border lg:bg-gray-200 lg:p-4 lg:dark:bg-zinc-800/30">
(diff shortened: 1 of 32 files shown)


NiklasBuchfink:
@felixhaeberle I have two thoughts:

1. Does it help to read the English RTL as we haven't translated or extract the messages for our apps in any RTL language yet? I asked my flatmate from Israel what he would expect.

2. The best use case at the moment would be to display and write the patterns of RTL languages from right to left. But we would need to parse them and we get the same problems as we have with the detection of a correct language tag.

felixhaeberle:
@NiklasBuchfink I don't understand both of your points, can you elaborate?

1. English is never RTL. Why would somebody display English in RTL?
2. The text display is the same, just flipped.

This PR is only about replacing the legacy CSS classes with more modern ones which align with the goal of our app to be served in multiple languages in the future.

NilsJacobsen:
I feel like the maintenance burden is too significant for basically no user asking for this. I quickly checked other big companies with many employees to see if they support it. The answer is yes, but they have a lot of layout problems. If they can't maintain it, we shouldn't try to accomplish it with a handful of people.

image
image
image

felixhaeberle:
What is holding us back from just using the modern CSS tailwind alternatives to the standard ones? 

We already decided not to go "full RTL support" but to make our components more RTL friendly ootb with using modern CSS alternative. 

I don't see a maintenance burden in using a modern CSS standard which is interchangeable to the current one, but with great benefits.

samuelstroschein:
> I don't see a maintenance burden in using a modern CSS standard which is interchangeable to the current one, but with great benefits. 

Because you need to a) disallow `m{l,t,r,b}`, `p{l,t,r,b}` and b) explain developers that a) can't be used anymore

a) is a one time effort
b) is a continuous effort

a and b combined don't achieve c, your stated goal of RTL support.

felixhaeberle:
Gotcha. Still not changing my mind on it but if team agrees on not going for it, we skip it for now. 👍

PR #1412, opened 2023-09-28
1044 refactor add a single point of truth for commands in ide extension

openscript:
Happy how easy it is to mock `vscode` with vitest.

```typescript
vi.mock("vscode",  () => {
	return {
		commands: vi.fn(),
		EventEmitter: vi.fn(),
	}
})
```


openscript:
recheck

felixhaeberle:
@openscript Will review this on Wednesday, today is public holidays in Germany.

PR #1414, opened 2023-09-29
feat: match object to array

NilsJacobsen:
See https://github.com/inlang/monorepo/issues/1403


NilsJacobsen:
I tested to open the editor and the ide extension with the starter project. It seems to work like before.

PR #1422, opened 2023-10-03
[refactor] json formatting detection

samuelstroschein:
The API is "stark verbesserungswürdig".

### DIFF
--- a/inlang/source-code/detect-json-formatting/README.md
+++ b/inlang/source-code/detect-json-formatting/README.md
@@ -1,1 +1,2 @@
 # @inlang/detect-json-formatting
+


samuelstroschein:
Initial reason for the refactor: 

I need a `const stringify = detectFormatting(json)` function. The API of however initializes a function with no file and then exposes `values` which are never used in the entire monorepo (?!). 

CleanShot 2023-10-03 at 16 15 30@2x

Further investigation revealed that the detect formatting function is parsing the JSON to detect nested keys, and based of the detection yields different stringified results. Very bad. The detect formatting function is essentially useless for non i18next or plugin json usage which may or may not rely on nested keys.

samuelstroschein:
Setting defaults in the detectJsonFormatting function seems contradictive too. If no formatting is detected, the JSON has no formatting/didn't exist before. No need to detect formatting then.

samuelstroschein:
Aha. You exposed values and enabled passing no file to the function to ease your testing. @NilsJacobsen don't introduce APIs that severely worsen usage but simplify testing.  

CleanShot 2023-10-03 at 16 44 35@2x

inlang/source-code/detect-json-formatting/src/detectJsonFormatting.test.ts
```diff
@@ -1,109 +1,27 @@
...
-	})
+it("should detect spacing", () => {
+	// test all possible spacings
+	for (const spacing of [1, 2, 3, 4, 6, 8, "\t"]) {
+		const withSpacing = `{\n${
+			spacing === "\t" ? "\t" : " ".repeat(spacing as number)
+		}"test": "test"\n}`
+		const serialize = detectJsonFormatting(withSpacing)
+		expect(serialize(JSON.parse(withSpacing))).toBe(withSpacing)
+	}
```
samuelstroschein, line 12: @NilsJacobsen using a generic test to go through all spacing possibilities probably saved 2-3 other tests. the test coverage is still 100%

samuelstroschein:
@NilsJacobsen 

1. do you understand my reasoning? 
2. how can i improve the feedback i give?

PR #1429, opened 2023-10-04
Move auth to lix

janfjohannes:


janfjohannes:
so much fun when a pr can suddenly remove so much :)

samuelstroschein (approved):
lovely to see a code reduction!

lix/source-code/client/src/browser-auth.ts
```diff
@@ -0,0 +1,98 @@
...
+	window.open(authUrl, "_blank")
+}
+export function logout() {
+	return fetch(`${publicEnv.PUBLIC_GIT_PROXY_BASE_URL}/services/auth/sign-out`, {
+		method: "POST",
+		credentials: "include",
+	})
+}
+
+interface Email {
```
felixhaeberle, line 27: @janfjohannes does this belong in here (also not exported) or on a more global level? could imagine that this could probably be used by multiple modules.

inlang/source-code/website/src/services/auth/src/pages/oauth-callback.page.tsx
```diff
@@ -14,8 +14,7 @@ import { useLocalStorage } from "#src/services/local-storage/index.js"
  */
 export function Page() {
 	// ! Extremely important to wrap the get user function
```
felixhaeberle, line 16: @janfjohannes why?

lix/source-code/client/src/openRepository.ts
```diff
@@ -24,56 +24,80 @@ export async function openRepository(
...
+				ghArgs[1].credentials = "include"
 
 				// @ts-ignore
 				return fetch(...ghArgs)
 			},
 		},
 	})
 
-	const normalizedUrl = `https://${host}/${owner}/${repoName}`
+	// TODO: clean way to use local repo allready in the fs
```
felixhaeberle, line 99: @janfjohannes typo 🤓

inlang/source-code/website/src/services/auth/src/pages/oauth-callback.page.tsx
```diff
@@ -1,7 +1,7 @@
 import { createEffect, createResource, Match, Switch } from "solid-js"
 import MaterialSymbolsCheckCircleRounded from "~icons/material-symbols/check-circle-rounded"
 import MaterialSymbolsArrowBackRounded from "~icons/material-symbols/arrow-back-rounded"
-import { getUserInfo } from "../implementation.telefunc.js"
+import { browserAuth } from "@lix-js/client"
```
felixhaeberle, line 4: @janfjohannes NICE! 😍

felixhaeberle (commented):
@janfjohannes So nice seeing things move into the right places & have code reduction. Great work! 

Do we also have tests for the browser auth? *I don't see a `browser-auth.test.ts`.*

lix/source-code/client/src/browser-auth.ts
samuelstroschein: On that note, the rest of the codebase uses `type`. Probably wise to choose `type` instead of `interface`.

lix/source-code/client/src/api.test.ts
```diff
@@ -26,7 +26,7 @@ describe("main workflow", () => {
 
 	it("opens a repo url without error and without blocking io", async () => {
 		// fix normalization of .git
-		repository = await openRepository("github.com/inlang/example", {
+		repository = await openRepository("https://github.com/inlang/example", {
```
samuelstroschein, line 29: How is a github repository opened without a lix server in between for CORS? 

If the openRepository implicitly routes via a cors proxy, I vote to make that explicit. 

```diff
-repository = await openRepository("https://github.com/inlang/example")
+repository = await openRepository("https://lix.dev/git/github.com/inlang/example")
```

lix/source-code/client/src/openRepository.ts
```diff
@@ -24,56 +24,80 @@ export async function openRepository(
...
 
-	// Git cors proxy and github cors proxy have to be on the same server and will be unified further
-	if (gitProxyUrl && gitProxyUrl?.startsWith("http")) {
-		const { origin } = new URL(gitProxyUrl)
-		gitHubProxyBaseUrl = origin
+		// check if all 3 parts are present, if not, return an error
+		if (!namespace || !host || !owner || !repoName) {
+			throw new Error(
+				`Invalid url format for '${url}' for cloning repository, please use the format of https://lix.inlang.com/git/github.com/inlang/monorepo.`
+			)
```
samuelstroschein, line 78: Should we write a function `isValidLixUri` that we can use to a) verify URIs and b) write tests for? 

I am referring to URI here because lix will be able to open local repositories where the path is not a URL but an URI.

samuelstroschein (changes requested):
Wups my previous review was too quick.

lix/source-code/client/src/api.test.ts
janfjohannes: @samuelstroschein this would be without any lix server or proxy, just using the lix client eg. in nodejs.   we have to decide if we want to support this usecase in the future when we seperate more from git/github, but for now this is how tests are run and works.

inlang/source-code/website/src/services/auth/src/pages/oauth-callback.page.tsx
janfjohannes: @felixhaeberle  good catch that was left over fromthe removed telefunc impl.

lix/source-code/client/src/browser-auth.ts
janfjohannes: @felixhaeberle   @samuelstroschein  i changed from interface to type but i dont want to expose it as this is typing according to github logic and i would prefer to only expose and reuse types more if we created them specifically for lix, tooling can still use it with return type inference.

janfjohannes:
> @janfjohannes So nice seeing things move into the right places & have code reduction. Great work!
> 
> Do we also have tests for the browser auth? _I don't see a `browser-auth.test.ts`._


browserAuth is just the moved logic of the editor github login to lix. as this requires cookies and redirects we could only test this with real integration tests but as we are probably moving to a github appp and own lix auth i think its currenlty not worth testing this, it was also not tested in the editor before

lix/source-code/client/src/helpers.test.ts
```diff
@@ -0,0 +1,53 @@
...
+
+		expect(parseResult).toStrictEqual({
+			protocol: "https:",
+			lixHost: "",
+			namespace: "",
+			repoHost: "github.com",
+			owner: "inlang",
+			repoName: "monorepo",
+		})
+	})
```
samuelstroschein, line 18: that tests contradicts the last test "should throw an misformed uri" where opening a github link is not possible.

samuelstroschein (changes requested):

lix/source-code/client/src/helpers.test.ts
janfjohannes: no that test checks if missing repo names are catched. i added explicit error test names to avoid this confusion.

felixhaeberle:
> browserAuth is just the moved logic of the editor github login to lix. as this requires cookies and redirects we could only test this with real integration tests but as we are probably moving to a github appp and own lix auth i think its currenlty not worth testing this, it was also not tested in the editor before

@janfjohannes Got it! You're right about not testing this atm, but at a later stage. 
Let's keep this in mind tho.

felixhaeberle (approved):

PR #1431, opened 2023-10-05
chore: refactor event emitters

openscript:


felixhaeberle (approved):
+1 lgtm

felixhaeberle:
@openscript We don't do patches anymore. We switched to ComVer instead of SemVer. See reasoning behind https://inlang.com/documentation/comver 

cc @samuelstroschein I see us communicating this a lot in the future 😄

samuelstroschein:
Both work. I wouldn't be too strict about SemVer vs ComVer.

felixhaeberle:
This is confusing 😄

samuelstroschein:
@felixhaeberle the alternative is to open a PR to changesets which enables a `noPatchVersions` config and doesn't prompt the developer if something is a patch

openscript:
> @felixhaeberle the alternative is to open a PR to changesets which enables a `noPatchVersions` config and doesn't prompt the developer if something is a patch

We probably both have been lurking around in the changesets config options, to see if this is possible to configure already.

openscript:
> @felixhaeberle the alternative is to open a PR to changesets which enables a `noPatchVersions` config and doesn't prompt the developer if something is a patch

The api is probably cleaner to have a `versioningScheme` option or so..

felixhaeberle:
@samuelstroschein created an issue over at changeset https://github.com/changesets/changesets/issues/1241

PR #1433, opened 2023-10-05
[improve] sdk error messages

samuelstroschein:
## Why

- i was constantly getting error messages that are hard to understand

- using `error.cause` to display additional information is unintuitive. most apps won't display those addiitonal cause errors (ran into this with paraglide)

- opening the definition of an error did not include the error message which contributed to "hard to understand the reason behind this error" problem

## Changes

- [x] messages are now part of the errors itself for fast lookup and separation of error content and error logic

- [x] error causes are now included in the error message


samuelstroschein:
@felixhaeberle @NiklasBuchfink @NilsJacobsen no in-depth review requires. i want you to be aware of this change and object if you deem this to be wrong. 

if you don't deem this change to be wrong, approve the pr.

felixhaeberle (commented):
👍 lgtm

NiklasBuchfink (approved):

PR #1445, opened 2023-10-12
1424 ide extension should work with http proxys

openscript:
welcome on board :cn:


felixhaeberle:
> I usually use https://github.com/goproxy/goproxy to create a proxy on my machine. Let's say the proxy runs on port 8080. Then I would set the environment variable https_proxy to https://127.0.0.1:8080 without having the proxy running. The tests should fail now. Then I switch on the proxy. The tests should succeed. 
Ah.. and be aware of that a docker container doesn't see the host system as localhost.
With goproxy it can look like proxy tcp -p ":8080" -T tcp -P "127.0.0.1:443"
https://proxyman.io/ would also work but it's quite expensive :)
or https://mitmproxy.org/

The tests are passing in both cases 😅

openscript:
Are you running the unit test (where the setting is mocked)? or inside the extension host?

felixhaeberle:
@openscript yes, I'm running the unit tests for the ide extension. can you take this testing task over and verify it works — I have trouble reproducing the "bad test" behavior. 

alternative: implement tests which checks not only good, but also bad behavior?

openscript:
Of course I've tested if it works, but now I also added a failing unit test.

felixhaeberle (approved):

felixhaeberle:
👍 lgtm

PR #1446, opened 2023-10-12
force absolute path for `settingsFilePath` to intercept `nodeishFs` with absolute paths

felixhaeberle:
> Why is this change needed?
> 
> To resolve absolute paths from the modules in the "project.inlang.json" properly, we need the "settingsFilePath" in "loadProject" to be absolute. This ensures that, based on the directory, we only have absolute paths, which are much easier to handle than relative paths.
> 
> We first noticed this issue with the paraglide-js CLI , which compiles messages. It can receive a relative "../../project.inlang.json" file as input but then resolves the files from its own CLI root instead of the root of the "project.inlang.json" directory – which is wrong.
> 
> We therefore have to make the paths absolute based on the "project.inlang.json" base directory to properly resolve files, which we do with intercepting `nodishFs` with `createNodeishFsWithAbsolutePaths` in `loadProject()`.
> 

In this PR:

- rewrote & tested `normalizePath` to be more powerful
- created & tested `createNodeishFsWithAbsolutePaths`
- integrated `createNodeishFsWithAbsolutePaths` in `loadProject()` to intercept `args.nodeishFs`
- rewrote all apps & tests to align with absolute paths

> minimal side task: added e2e testing for badge


samuelstroschein (approved):

felixhaeberle:
@janfjohannes Can you also review?

janfjohannes (approved):

Issue #1453, opened 2023-10-13
[improvement] validate that `sourceLanguageTag` is contained in `languageTags`


samuelstroschein:
## Problem 

The inlang ecosystem expects the `sourceLanguageTag` to be contained in `languageTags`. The expectation is not validated, though, because JSON schema can't be used to express the expectation. 

## Proposal 

Hardcode an error in the SDK to increase the ecosystem experience (EX).

```ts
// loadProject

if (project.languageTags.includes(project.sourceLanguageTag) === false){
   project.errors.push(new ProjectLanguageTagsDoesNotContainSourceError())
}
```

## Additional information 

I was about to hardcode this validation in paraglide-js. But, having the validation in the SDK and thereby increase the ecosystem experience (EX) is better.

samuelstroschein:
@felixhaeberle can you do this?

felixhaeberle:
@samuelstroschein yes!

PR #1467, opened 2023-10-15
add check `languageTags` for `sourceLanguageTag`

felixhaeberle:
This PR aims to verify that  `sourceLanguageTag` is in `languageTags. 

If not, it should throw an error.

### DIFF
--- a/inlang/source-code/sdk/src/loadProject.ts
+++ b/inlang/source-code/sdk/src/loadProject.ts
@@ -6,7 +6,7 @@ import type {
 	Subscribable,
 } from "./api.js"
 import { type ImportFunction, resolveModules } from "./resolve-modules/index.js"
-import { TypeCompiler } from "@sinclair/typebox/compiler"
+import { TypeCompiler, ValueErrorType } from "@sinclair/typebox/compiler"
 import {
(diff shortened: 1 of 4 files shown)


felixhaeberle:
fixes #1453

inlang/source-code/sdk/src/loadProject.ts
```diff
@@ -278,6 +278,24 @@ const parseSettings = (settings: unknown) => {
 			})
 		}
 	}
+
+	const { sourceLanguageTag, languageTags } = settings as ProjectSettings
```
samuelstroschein, line 282: why are you destructuring and then casting to an incorrect (!) type?

Just use `settings.sourceLanguageTag` and `setings.languageTags` directly?

samuelstroschein (changes requested):

inlang/source-code/sdk/src/loadProject.ts
felixhaeberle: I did not know why settings are of type `unknown` – or is it true after we have checked for false with `.Check()` ?

samuelstroschein (approved):

Issue #1456, opened 2023-10-13
[bug] loadProject doesn't resolve modules


samuelstroschein:
## Problem 

The implementation is buggy. 

The regex is wrong...

<img width="902" alt="CleanShot 2023-10-13 at 19 09 03@2x" src="https://github.com/inlang/monorepo/assets/35429197/110d73ae-90eb-40f5-9458-6b849c88377b">

<img width="990" alt="CleanShot 2023-10-13 at 19 07 21@2x" src="https://github.com/inlang/monorepo/assets/35429197/791d4f0e-1351-403c-b7d7-97767bde3e77">

## Proposal 

@felixhaeberle I'll change the code to 

1. don't use basepath anymore. nodeishfs should resolve from the settings path. that makes nodeish fs unit testable. 
2. remove that regex (i knew the regex will be problematic the moment i saw it)

cc @martin-lysk this might have been your problem too

samuelstroschein:
@felixhaeberle the error should not have been thrown in `createNodesifh...`. instead `loadProject` for transparency. nothing in SDK should throw to catch errors more effectively. validating input parameters is an exception

samuelstroschein:
actually, i decided to throw at both points although the loadProject pre-ceeds. and, i also decided that we should give a proper error

samuelstroschein:
fixed by [936cfa4](https://github.com/inlang/monorepo/commit/936cfa40138c31ff36686e19b83a9743569d0bf4)

felixhaeberle:
Can you add a backslash example to the tests for Windows? Or is the path already normalized at this point? Because you proceed with .split("/") then and there, it's probably not correct.

Also "bathPath" is probably not what you want 🛁😄

samuelstroschein:
> Sorry if there was a bug, I verified that all apps were working and adjusted all tests which were passing but you are correct the tests where not optimal.

The code only worked for the limited test you wrote and failed in production. 

> Can you add a backslash example to the tests for Windows? Or is the path already normalized at this point? Because you proceed with .split("/") then and there, it's probably not correct.

Good point. @inlang/lix should force UNIX paths and automatically transform windows paths.

felixhaeberle:
@samuelstroschein Where in production did it fail? I tested the editor, cli, ide extension, badge – all manually, but never saw it failing.

samuelstroschein:
@felixhaeberle paraglide-js

samuelstroschein:
[...]

PR #1468, opened 2023-10-15
#1456 add windows tests & fix tests

felixhaeberle:
#1456

### DIFF
--- a/inlang/source-code/sdk/src/createNodeishFsWithAbsolutePaths.test.ts
+++ b/inlang/source-code/sdk/src/createNodeishFsWithAbsolutePaths.test.ts
@@ -21,6 +21,14 @@ it("intercepts paths correctly for readFile", async () => {
 		["../folder/file.txt", `/Users/samuel/Documents/paraglide/folder/file.txt`],
 		["../../file.txt", `/Users/samuel/Documents/file.txt`],
 		["../../../file.txt", `/Users/samuel/file.txt`],
+		["..\\file.txt", `/Users/samuel/Documents/paraglide/file.txt`],
+		["..\\folder\\file.txt", `/Users/samuel/Documents/paraglide/folder/file.txt`],
+		["..\\..\\file.txt", `/Users/samuel/Documents/file.txt`],
+		["..\\..\\..\\file.txt", `/Users/samuel/file.txt`],
+		[
+			"\\Users\\samuel\\Documents\\paraglide\\file.txt",
+			`/Users/samuel/Documents/paraglide/file.txt`,
+		],
 	]
 
 	const mockNodeishFs = {
--- a/inlang/source-code/sdk/src/createNodeishFsWithAbsolutePaths.ts
+++ b/inlang/source-code/sdk/src/createNodeishFsWithAbsolutePaths.ts
@@ -19,14 +19,14 @@ export const createNodeishFsWithAbsolutePaths = (args: {
 
 	// get the base path of the settings file by
 	// removing the file name from the path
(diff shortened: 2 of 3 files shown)


felixhaeberle:
fixes #1456

inlang/source-code/sdk/src/createNodeishFsWithAbsolutePaths.test.ts
```diff
@@ -21,6 +21,14 @@ it("intercepts paths correctly for readFile", async () => {
 		["../folder/file.txt", `/Users/samuel/Documents/paraglide/folder/file.txt`],
 		["../../file.txt", `/Users/samuel/Documents/file.txt`],
 		["../../../file.txt", `/Users/samuel/file.txt`],
+		["..\\file.txt", `/Users/samuel/Documents/paraglide/file.txt`],
+		["..\\folder\\file.txt", `/Users/samuel/Documents/paraglide/folder/file.txt`],
+		["..\\..\\file.txt", `/Users/samuel/Documents/file.txt`],
+		["..\\..\\..\\file.txt", `/Users/samuel/file.txt`],
```
samuelstroschein, line 27: I don't think windows "go up to parent directory" paths use `\\..\..`. An initial search supports my impression https://stackoverflow.com/questions/5050525/what-are-and-in-windows-directory#:~:text=Use%20two%20consecutive%20periods%20(..,For%20more%20information%2C%20see%20Paths.

inlang/source-code/sdk/src/createNodeishFsWithAbsolutePaths.ts
```diff
@@ -19,14 +19,14 @@ export const createNodeishFsWithAbsolutePaths = (args: {
 
 	// get the base path of the settings file by
 	// removing the file name from the path
-	const bathPath = args.settingsFilePath.split("/").slice(0, -1).join("/")
+	const basePath = args.settingsFilePath.split(/[/\\]/).slice(0, -1).join("/")
```
samuelstroschein, line 22: why are you doing this? The path should be normalized at this point. And a windows path doesn't even seem to use `\\`

samuelstroschein (changes requested):

inlang/source-code/sdk/src/createNodeishFsWithAbsolutePaths.test.ts
felixhaeberle: It's just a test which should properly resolve & I don't cover Windows standard behavior here. It can be human error & therefore I want to guarantee it's resolving correctly even with backslashs.

inlang/source-code/sdk/src/createNodeishFsWithAbsolutePaths.ts
felixhaeberle: it's the same expression like here https://github.com/inlang/monorepo/blob/029d78dc877dbcb3273dd617a709a06df467dc9d/inlang/source-code/sdk/src/createNodeishFsWithAbsolutePaths.ts & no, it's not normalized at this point.

inlang/source-code/sdk/src/createNodeishFsWithAbsolutePaths.test.ts
felixhaeberle: fixed with normalizing path before.

inlang/source-code/sdk/src/createNodeishFsWithAbsolutePaths.ts
felixhaeberle: fixed with normalizing path before.

samuelstroschein (approved):

PR #1471, opened 2023-10-16
Make inlang the marketplace for globalization

NiklasBuchfink:
The goal of this PR is to make inlang the marketplace for globalization. We want a new starting point to iterate in the next few days. Therefore, this will be a stripped-down version that only includes some of the features we have in mind.

- [x] remove old components accept card layout
- [x] change routing
- [x] landing page (feature & welcome card)
- [x] design and add welcome card graphic
- [x] search in header
- [x] category nav
- [x] subcategory nav
- [x] lix section and badge / marker / prefix

---
After this PR is merged:
- adjust cards / wider layout
- optimize for conversion
- new search algorithm
- more powerful markdown pages
- new content
- separate `/editor`
- use lit everywhere

cc @floriandwt


NilsJacobsen:
Add the marketplace header section. (no graphic) @NiklasBuchfink 

image

NilsJacobsen:
image

Alright guys. @inlang/team feel free to review this thing.

The marketplace should work like this for now. Don't be too nitty-gritty with some detailed styling feedback. This is a subject that will be improved. CI is green; we tested the features. We tried to find every bug, but I am sure someone will find something so... This is the structured review process that we talked about :D

Before we merge, let's update the registry with some additional keywords. So we do not get empty pages for the categories even though they share the same modules... this will happen in a different PR, directly on main, because it does not affect the live version.

samuelstroschein (approved):
Here and there "style discussions" But, le'ts merge this and improve iteratively!

PR #1497, opened 2023-10-19
Branch support

janfjohannes:
(depends on new auth branch, merge after that)

implements a simple version of branch support that re-opens the repo on a different branch when switching. This is not super optimized but uses the same code path for opening on a branch with a specific url/link and switching the branch, so is simpler to test and low complexity. this can be optimized so that only the diff between branches are fetched after we implement lazy loading.


felixhaeberle:
@janfjohannes Wuhu this finally gets reality 🤩 so many people have asked for this. 🎉

samuelstroschein (approved):

PR #1507, opened 2023-10-20
Added validate Modulte Settings 

jannesblobel:
This PR adds the function of validating the module settings. It allows us to throw an error if the users try to use incorrect Module settings


samuelstroschein:
I was justa bout to review :D

inlang/source-code/sdk/src/resolve-modules/resolveModules.ts
```diff
@@ -13,12 +14,13 @@ import { createImport } from "./import.js"
...
 import { resolvePlugins } from "./plugins/resolvePlugins.js"
 import { TypeCompiler } from "@sinclair/typebox/compiler"
+import { validatedModuleSettings } from "./validatedModuleSettings.js"
 
 const ModuleCompiler = TypeCompiler.Compile(InlangModule)
 
 export const resolveModules: ResolveModuleFunction = async (args) => {
 	const _import = args._import ?? createImport({ readFile: args.nodeishFs.readFile })
-	const moduleErrors: Array<ModuleError> = []
+	const moduleErrors: Array<any> = []
```
samuelstroschein, line 23: Stop. Why `any` as a type. The Error should be `ModuleError`

samuelstroschein (changes requested):
You have not added a test in `resolveModules` that checks if `ModuleSettingsAreInvalidError` is returned. Please do so.

samuelstroschein (approved):

PR #1533, opened 2023-10-30
Add note about Vite' Pre-bundling to build-a-global-svelte-app blog post

sunxyw:


samuelstroschein:
@sunxyw `noExternal` is also required to force vite to bundle messages, correct?

inlang/guides/build-a-global-svelte-app/build-a-global-svelte-app.md
```diff
@@ -56,6 +56,30 @@ npm run build
...
+
+Or just in local environment (`serve` or `dev` command):
+```js
+// vite.config.js/ts
+export default defineConfig(({ command }) => ({
+  optimizeDeps: {
+    exclude: command === 'serve' ? ['@inlang/paraglide-js'] : []
+  },
+}));
+```
```
samuelstroschein, line 81: I don't think this is necessary. 

The vite docs mention that optimizeDeps only happens in dev mode. "Dependency pre-bundling only applies in development mode".

sunxyw: you are right! I'll remove this part.

sunxyw:
> @sunxyw `noExternal` is also required to force vite to bundle messages, correct?

@samuelstroschein I am not sure.

I'm not an expert on vite and don't quite understand what `noExternal` does.
After reading the description in the documentation, this seems to be just an optimization for dependencies that provide both ESM and CommonJS. It should not affect the loading of `messages.js`?

I haven't included `noExternal` in my project but it seems to work.

I'll happily include it in the article if you think it's necessary.

samuelstroschein (approved):
Seems good. Thanks for the contribution!

NilsJacobsen (approved):
Looks good to me.

PR #1542, opened 2023-10-30
add `--project` flag – fixes [#1527]

### LINKED ISSUE #1527: No monorepo support for cli lint
## Problem 

I can not use the --config anymore. The `package.json` and the `project.inlang.config` need to be on the same dir right now.

## Proposal
Add a --project option as we have it in the compile command of paraglide-js.

@inlang/cli

felixhaeberle:
fixes #1527

### DIFF
--- a/inlang/source-code/cli/src/commands/lint/index.ts
+++ b/inlang/source-code/cli/src/commands/lint/index.ts
@@ -3,11 +3,12 @@ import Table from "cli-table3"
 import { getInlangProject } from "../../utilities/getInlangProject.js"
 import { log } from "../../utilities/log.js"
 import type { InlangProject, MessageLintReport } from "@inlang/sdk"
+import { projectOption } from "../../utilities/globalFlags.js"
 
 export const lint = new Command()
 	.command("lint")
 	.description("Commands for linting translations.")
-	.requiredOption("--project <path>", "Path to the project settings file.", "./project.inlang.json")
+	.requiredOption(projectOption.flags, projectOption.description, projectOption.defaultValue)
 	.option("--no-fail", "Disable throwing an error if linting fails.", false)
 	.action(async (args: { project: string }) => {
 		const project = await getInlangProject({ projectPath: args.project })
--- a/inlang/source-code/cli/src/commands/lint/index.ts
+++ b/inlang/source-code/cli/src/commands/lint/index.ts
@@ -7,9 +7,10 @@ import type { InlangProject, MessageLintReport } from "@inlang/sdk"
 export const lint = new Command()
 	.command("lint")
 	.description("Commands for linting translations.")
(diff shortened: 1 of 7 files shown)


inlang/source-code/cli/src/commands/machine/translate.ts
```diff
@@ -8,9 +8,10 @@ import prompts from "prompts"
 
 export const translate = new Command()
 	.command("translate")
+	.requiredOption("--project <path>", "Path to the project settings file.", "./project.inlang.json")
```
samuelstroschein, line 11: what about a variable consumed by commands who need it to ensure consistency without relying on global options? 

```ts
export const projectOption = ["--project <path>", "Path to the project settings file", "./project.inlang.json"]
```

inlang/source-code/cli/src/utilities/getInlangProject.ts
```diff
@@ -1,14 +1,14 @@
...
 import { telemetry } from "../services/telemetry/implementation.js"
+import { normalizePath } from "@lix-js/fs"
 
 /**
  * Gets the inlang project and exists if the project contains errors.
  */
-export async function getInlangProject(): Promise<InlangProject> {
+export async function getInlangProject(args: { projectPath: string }): Promise<InlangProject> {
 	const baseDirectory = process.cwd()
-	const settingsFilePath = resolve(baseDirectory, "project.inlang.json")
```
samuelstroschein, line 11: why is resolve not required anymore?

inlang/source-code/cli/src/commands/machine/translate.ts
felixhaeberle: done ✅

inlang/source-code/cli/src/utilities/getInlangProject.ts
felixhaeberle: removed it for normalizedPath, but let's keep resolve where we can.

inlang/source-code/cli/src/commands/lint/index.ts
```diff
@@ -3,11 +3,12 @@ import Table from "cli-table3"
 import { getInlangProject } from "../../utilities/getInlangProject.js"
 import { log } from "../../utilities/log.js"
 import type { InlangProject, MessageLintReport } from "@inlang/sdk"
+import { projectOption } from "../../utilities/globalFlags.js"
 
 export const lint = new Command()
 	.command("lint")
 	.description("Commands for linting translations.")
-	.requiredOption("--project <path>", "Path to the project settings file.", "./project.inlang.json")
+	.requiredOption(projectOption.flags, projectOption.description, projectOption.defaultValue)
```
samuelstroschein, line 11: I think you can spread the project option

felixhaeberle: no somehow this raised type issues thus typed array.

Issue #1550, opened 2023-10-31
solution for missing sitemap


kiemflo:
## Problem 
A lot of our pages are not listed on google and the search console bot can't detect them without a sitemap see here:
<img width="952" alt="Screenshot 2023-10-31 at 3 36 33 PM" src="https://github.com/inlang/monorepo/assets/92092993/38d0eaa9-7c1f-421b-8943-35d2ecdb7f32">

## Proposal
- search for a solution to dynamically generate a sitemap with every build step so we get more visitors via SEO

cc @inlang/website

floiem:
Possible solution: https://github.com/jbaubree/vite-plugin-pages-sitemap

NiklasBuchfink:
@floiem this will be a problem:

> You must use [vite-plugin-pages](https://github.com/hannoeru/vite-plugin-pages/) routes generation to use this plugin. If you are not working with vite-plugin-pages, i recommend to use [vite-plugin-sitemap](https://github.com/jbaubree/vite-plugin-sitemap).

But to leverage the vite ecosystem for that seems to be the right choice. What about this library:
https://github.com/jbaubree/vite-plugin-sitemap

floiem:
@NiklasBuchfink I've looked at all, but it seems to make more sense to just write an own script we can adjust freely without any additional dependencies: https://github.com/inlang/monorepo/pull/1563

I got this idea from looking into the solidjs website repo. They also have their own script for making the sitemap.

NiklasBuchfink:
Top job 👍  Maybe you can adjust it to keep track of "inlang/editor" after the merge as well.

PR #1563, opened 2023-11-01
feat: implement dynamic sitemap generation

kiemflo:
Based on #1550 

TLDR: Google couldn't find marketplace items because there was no sitemap + a sitemap is very useful anyway.

This PR introduces a sitemap generation script executable by the website as a build step, similar to the solution `solidJS` has: https://github.com/solidjs/solid-site/blob/main/scripts/generateSitemap.js

### What pages to track?

We have dynamic content -> hard to generate sitemaps for and there is no good, well maintained solution for it.

Therefore, I have introduced a simple array we can update when we introduce new pages. For all pages that are important atm, I have implemented separate methods to track all pages, specifically for dynamic pages. It is the best way to maintain the website's SEO imo.

### How to generate?
from the repo root execute: `npm run build:sitemap --w=@inlang/website`

Screenshot 2023-11-01 at 3 50 17 PM


samuelstroschein:
@floiem why don;t you include the sitemap building in the build script itself? An extra build step has the risk of becoming unused because ppl use `npm run build` alone

floiem:
> @floiem why don;t you include the sitemap building in the build script itself? An extra build step has the risk of becoming unused because ppl use `npm run build` alone

@samuelstroschein because we don't always need to update the sitemap, just when new pages are getting added or new marketplace items, that's why :D

samuelstroschein:
> @floiem because we don't always need to update the sitemap, just when new pages are getting added or new marketplace items, that's why :D

So the script needs be manually invoked? If so, what system is in place to inform people that the script needs to be run? 

It seems simpler to just run the script on every build and thereby ensure that the sitemap is never out of date.

Issue #1534, opened 2023-10-30
filesystem watch feature


janfjohannes:

janfjohannes:
@samuelstroschein

samuelstroschein:
Thanks @janfjohannes. 

We need the `fs.watch` API in lix nodeish fs. 

- ASAP priority to unblock paraglide-js and #1519 
- only the API and in-memory implementation is needed, no cross env watching like web editor watches for changes on local filesystem

Let me know if I need to elaborate further. Once nodeish fs supports `fs.watch`, @NilsJacobsen can go ahead and implement #1519

PR #1572, opened 2023-11-02
initial version of filesystem watcher [closese #1534]

janfjohannes:


samuelstroschein (approved):
Looked into the type mismatch. Can't figure it out either. The test suite passes. Hence, approve. 

Looking forward to a sane lix fs API in the future

samuelstroschein:
@janfjohannes we shouldn't merge this without the type check. else we can't pass node:fs/promises anymore. could `any` be of help here?

janfjohannes:
@samuelstroschein @NilsJacobsen @felixhaeberle i fixed the flaky test and reenabled the type test except for the watch function, this should be a good enough  baseline without degrading our type checks to be merged.

PR #1573, opened 2023-11-02
Change return type of `extractMessageOptions`

felixhaeberle:
### Problem

The m function matcher should also be able to manipulate the id, and not only the replacement. This is fixing the bug of the ide extension in the function matcher who does a transformation of the id like for the message format plugin.

### Solution

Returning an object `{ messageId: string, messageReplacement: string }` instead of `string`

### DIFF
--- a/.changeset/thick-eels-hang.md
+++ b/.changeset/thick-eels-hang.md
@@ -1,0 +1,9 @@
+---
+"@inlang/plugin": minor
+"@inlang/plugin-m-function-matcher": minor
+"@inlang/plugin-t-function-matcher": minor
+"@inlang/plugin-i18next": minor
+"vs-code-extension": minor
+---
+
+change return type of extractMessageOptions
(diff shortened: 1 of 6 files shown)


samuelstroschein:
> The m function matcher should also be able to manipulate the id, and not only the replacement. This is fixing the bug of the ide extension in the function matcher who does a transformation of the id like for the message format plugin.

I don't understand the requirement from this paragraph nor the code. 

Can you elaborate, probably with an example?

samuelstroschein (approved):
Understood the requirement after the call.

felixhaeberle:
The message id which gets inserted by the user in the process of message extract should also be transformable by the plugin because it's required for example from inlang message format which transforms the input to a JS readable function & therefore the message id & the message function should be the same.

we discussed a validation step which only makes the process of extracting strings longer & we can't properly know which format to validate because multiple formats could be correct. this should be handled by a lint after all & we should just guarantee that the id is valid JS function syntax.

PR #1576, opened 2023-11-02
remove reload prompt

felixhaeberle:
remove reload prompt & instead reload the project silently.


felixhaeberle:
Tested it by myself.

PR #1610, opened 2023-11-08
Sdk watcher

NilsJacobsen:
- adds a watcher to fs
- If something gets loaded, paths establish watcher and when something changes it gets loaded again.


NilsJacobsen:
This is blocked by a paraglide-js type error. 

example:
image

Other than that it is ready to be merged when PR is green.

NilsJacobsen:
@samuelstroschein do you know about the type erros of paraglide? Can we fix that somehow?

samuelstroschein:
@NilsJacobsen yes. I will fix the error.

samuelstroschein:
@NilsJacobsen merge from main. `npm run ci` runs through

NilsJacobsen:
We can still not merge. We have now an error `not implemented` in the paraglide-js cli tests. Probably a fs topic. @janfjohannes  @samuelstroschein @felixhaeberle can you resolve this? I need to work on the website to hot-fix stuff until Saturday.

NilsJacobsen:
Update:

The error get's thrown by memfs. Which doesn't have a watch API.
@samuelstroschein why didn't you use our own implementation?

samuelstroschein:
@NilsJacobsen can you replace the memfs with nodeishfs in pararglide's tests? be aware that you need to work around checking if a directory exists now. it is horrible

NilsJacobsen:
Gonna change it to lix fs now.

janfjohannes:
@samuelstroschein  we should not use node:fs/memfs anywhere, there is fs.stat to check if a file exists, its not the nicest api but commonly used and implemented in our memoryfs. still think its a bit too early to reaimplement the fs api completely, the hard part will be that we will still need to deal with the nodejs crap as we will need to translate our new nice api to nodejs in the end anyways so the amount of work this creates will be immense next to changing everything that uses the nodeishFS api. the situation would be different if someone finds a well adapted and up to date project similar to https://github.com/jimmywarting/native-file-system-adapter

samuelstroschein:
>  the amount of work this creates will be immense next to changing everything that uses the nodeishFS API 

becomes ever larger the longer we wait. 

let's wait a bit longer but if we can't get our fs problems under control, we should think about an fs API that lix provides. nothing crazy, look how simple the vscode fs API is https://code.visualstudio.com/api/references/vscode-api#FileSystem. 

off-topic: the filesystem access API seems unsuited for Lix. the API is designed for handling files that are outside of the browser, sacrificing usability for "tranfitional "read this file" apps" as lix apps will mostly be. interop with the types that file system access provides should work and is what bun does too.

```ts
// redundant in lix apps
const [filehandle] = await window.showOpenFilePicker();
// get file does not take a path -> will not work for lix apps
const file = await fileHandle.getFile();

// lix fs API 
const file = await fs.getFile("/messages.json")
```

NilsJacobsen:
@samuelstroschein it is green now. I can get the empty function that you proposed

lix/source-code/fs/src/NodeishFilesystemApi.ts
```diff
@@ -26,6 +26,7 @@ export type NodeishFilesystem = {
 		options?: {
 			recursive?: boolean
 			signal?: AbortSignal
+			persistent?: false
```
samuelstroschein, line 29: this is not nodeish API spec. the spec is default persistent to true right? If so, please remove this.

janfjohannes: @samuelstroschein  we only support persistent: false .  this is supposed to not be the default value but the only valid value, happy for a better type proposal.

janfjohannes:
@samuelstroschein   regarding "filesystem access API" i did not mean specifically the api just something like the project i linked, a standardized fs api that runs in browsers and other runtime with pluggable backend impl. actually the vscode one is allready on my radar as a possible candidate,

lix/source-code/fs/src/NodeishFilesystemApi.ts
NilsJacobsen: If we remove this, the builds are stuck. @samuelstroschein.

NilsJacobsen:
We merged it so Felix could work on the plugin. If this is still an issue we'll change it.

PR #1624, opened 2023-11-10
human readable message format

samuelstroschein:


samuelstroschein:
the tests dont pass because the registry is outdated. i hate it. will merge to publish the new version to the registry and fix on main

PR #1650, opened 2023-11-17
Fix dev setup

janfjohannes:


janfjohannes:
@samuelstroschein  i think this is a really good compromise to keep most of your ideal external contributor flow, but also be not too error prone for inlang team members :)

janfjohannes:
it seems CI is also relying on the public dev environment on doppler in some parts instead of using the CI environment everywhere, which is a good catch of the new setup.

inlang/source-code/env-variables/build.ts
```diff
@@ -17,16 +16,11 @@ config({ path: rootEnvFilePath, override: false })
 const { error: errors } = validateEnvVariables({ forProduction: false })
 
 if (errors) {
-	console.error("💡 Some env variables are not defined. Fetching public env variables remotely...")
+	console.error(
+		"❗️❗️ Some env variables are not defined or do not pass validation. If you are an external developer, please run `npm run fetchPublicEnv` to get the latest version, if you are part of the inlang team, please check your doppler setup."
```
samuelstroschein, line 20: `npm run fetchPublicEnv` will break in the monorep setup, assuming that developers start from the root of the repo. 

`npm run fetchPublicEnv -w @inlang/env-variables` will likely

samuelstroschein (approved):

inlang/source-code/env-variables/build.ts
janfjohannes: @samuelstroschein  i added the command to the root of the repo too:
"fetchPublicEnv": "npx nx run @inlang/env-variables:fetchPublicEnv",

PR #1659, opened 2023-11-20
Replace `npm` workspaces with `pnpm` workspaces

### LINKED ISSUE #1526: fix npm workspaces + dependency hell – switch to lerna / rush ?
## Problem

In the past, multiple people had problems with our current approach & that dependencies are not correctly updated.

The main problem is with:

```json
"@inlang/json-types": "*",
"@inlang/translatable": "*",
"@inlang/message-lint-rule": "*",
"@inlang/module": "*",
"@inlang/language-tag": "*",
"@inlang/message": "*",
"@inlang/plugin": "*",
```

which doesn't resolve correctly when updated.

## Solution

This issue aims to find a way of tackling this issue and find a new way with another tooling like pnpm + lerna or rush.

> A new approach has to be investigated & tested.

cc @samuelstroschein 

## Related 

- [x] https://github.com/inlang/internal/issues/176

LorisSigrist:
Closes #1526 

## What 
`npm workspaces` has been causing lot's of problems with ghost dependencies, partial installs and so on, so it's being replaced. 

This PR introduces `pnpm` as the primary package manager and updates package scripts & CI to match. This will be significantly more reliable. 

When publishing packages that have workspace-dependencies, `pnpm` will pin the version of the dependency when publishing. `npm` would simply leave a wildcard, causing all sorts of odd behaviour.

As a nice benefit, this also significantly improves install times because of `pnpm`'s global package cache, so each package-version only gets downloaded once. On my machines this brings the install time for the entire Monorepo down to 12 seconds. 

## What you'll need
Your system needs to have `pnpm >= 8` installed.

## What changes
You'll use `pnpm install` instead of `npm install`. That's about it.


LorisSigrist:

LorisSigrist:
The editor & website currently aren't building properly. 
This is because they use paraglide and the paraglide package contains /examples, which are different packages. Pnpm doesn't handle this properly, causing an infinite recursion. We'll probably need to move the examples.

felixhaeberle:
@LorisSigrist so package nesting within packages is not supported by paraglide? aren't they on the same level in the paraglide folder?

samuelstroschein:
Urgh so annoying. Let me review the PR in approx 50 min. 

I assume moving the examples should be OK. Will verify it in approx 50 min.

samuelstroschein (changes requested):
### 1. I vote to remove `ci:publish` and `ci:version` from the scripts. 

#### ci:version

Breaks the versioning github action which automatically versions packages when merging a change PR (see e.g. https://github.com/inlang/monorepo/pull/1648). Hence, no need for a version script anyways. 

#### ci:publish

1. Changesets is unmaintained. 
I would not depend on changeset to version things in an extra script. We are already discussing to move to rush stack (if changeset keeps misfunctioning after switching to npm).

2. Changeset is opinionated. 
Among things, the publish command creates git tags. Nice idea but extremely annoying in a large monorepo. The result are thousands of git tags that nobody looks at anyways. 

### 2. moving the examples

Yes, we can move the examples to `paraglide-js-examples/*`

samuelstroschein:
Oh btw the .npmrc file can be deleted https://github.com/inlang/monorepo/blob/main/.npmrc

LorisSigrist:
Moving the example out of paraglide has seemingly solved the circular dependecy issue, but now results in a new error. 

```
[vite:css] [postcss] EMFILE: too many open files, open '/.../monorepo/inlang/source-code/editor/node_modules/@inlang/paraglide-js/node_modules/memfs/lib/internal/errors.js'
```

This happens when building either the editor, or the website. All other builds are fine. Investigating...

LorisSigrist:
The error happening inside `paraglide-js` in the above message seems to be a coincidence. It can happen in other packages too.

LorisSigrist:
It's very weird that vite would even scan the paraglide folder. Paraglide is not used as a dependecy anywhere.
Some research suggests that tailwind or postcss are trying to scan all files in the monorepo, causing us to hit the 1024 file handle limit. Since the scan order is deterministic it always crashes at the same spot. 

However, I can't explain why this happens when pnpm is used, but no when npm is used. Maybe the symlinks in node_modules trip it up? I don't know yet.

LorisSigrist:
I found the issue.
It's this line in the editor/website `tailwind.config.js`

```
content: [path.join(__dirname, "./**/*.{js,ts,jsx,tsx}")],
```

This does not exclude node_modules. With npm this isn't a problem since the `node_modules` folder includes a copy of the dependencies, meaning it's a neatly organised tree. With pnpm however, `node_modules` contains symlinks, and in our case, symlinks to somewhere else in the monorepo. Therefore tailwind's crawler get's lost and just opens all the files.

Changing it to 
```
content: [path.join(__dirname, "./src/**/*.{js,ts,jsx,tsx}")],
```
fixes it.

LorisSigrist:
Question @samuelstroschein 
If we aren't going to use changesets anymore, what should be the trigger for publishing packages?
Should the CI simply publish any packages that had their version bumped?

samuelstroschein:
> Should the CI simply publish any packages that had their version bumped?

That's how it works at the moment.

felixhaeberle:
@LorisSigrist Have you searched through the docs (globally search the repo) if we have any mentions of npm left? I see you already updated the CONTRIBUTING.md – nice! ✨

.github/workflows/publish-npm-packages.yml
```diff
@@ -41,163 +49,12 @@ jobs:
           token: ${{ secrets.NPM_PUBLISH_TOKEN }}
           package: inlang/source-code/cli/package.json
 
-      - name: "@inlang/json-types"
-        uses: JS-DevTools/npm-publish@v2
```
samuelstroschein, line 45: Does an action like this exist for pnpm? 

- action checks if version is already published

felixhaeberle:
> Changesets is unmaintained.

I would consider it feature complete. Last change happened 4 months ago. Even pnpm use it themselves.

> Changeset is opinionated.

I would not put it like that.

> That's how it works at the moment.

What @LorisSigrist might want to say: with changesets it's managed & we get auto changelog.

LorisSigrist:
> @LorisSigrist Have you searched through the docs (globally search the repo) if we have any mentions of npm left? I see you already updated the CONTRIBUTING.md – nice! ✨

Just found the last occurance of `npm`. Should be good now

.github/workflows/publish-npm-packages.yml
samuelstroschein: Perfect!

samuelstroschein:
> @felixhaeberle  I would consider it feature complete. 

It is buggy. No issues are being closed anymore. 

Regardless, @LorisSigrist found a solution https://github.com/inlang/monorepo/pull/1659#discussion_r1400755030

samuelstroschein (approved):
#

felixhaeberle:
@samuelstroschein Yep we already talked about built-in pnpm publish 👍 I think we are good to go with this.

felixhaeberle (approved):

PR #1571, opened 2023-11-02
create paraglide js vite plugin

felixhaeberle:
fixes #1518 

cc @samuelstroschein do you know how we can fix the ts server not recognizing the change?

PR #1669, opened 2023-11-21
Adds Unplugin Plugins around ParaglideJS

### LINKED ISSUE #1518: paraglide vite plugin
- [ ] sets ssr no external for paraglide-js https://github.com/inlang/monorepo/issues/1485#issuecomment-1779918933
- [ ] includes the compile step (removes the need to include the compile step in package.json)
- [ ] if messages change, the compile step is re-executed [blocked by #1519]
- [ ] exclude paraglide in `optimizeDeps`, see https://github.com/inlang/monorepo/pull/1533, to get HMR working

**Pseudocode**

```ts
import { paraglideJsVitePlugin } from "@inlang/paraglide-js"

export default defineConfig({
	plugins: [
            paraglideJsVitePlugin({
              name: "frontend",
              project: "./project.inlang.json"
            })
	]
})


```

LorisSigrist:
Continuation of #1571 
Closes #1518 

This PR adds several bundler plugins for ParaglideJS. These plugins automatically call `paraglide compile` at the appropriate times (startup, message change), removing the need for the scripts in `package.json`.

- `@inlang/paraglide-js-adapter-vite`
- `@inlang/paraglide-js-adapter-rollup`
- `@inlang/paraglide-js-adapter-webpack`
- `@inlang/paraglide-js-adapter-unplugin` Is directly exposed for use in more exotic setups, such as raw esbuild or rspack.

Thanks to Unplugin, we can get all these Bundler plugins from a single implementation.


inlang/source-code/paraglide/paraglide-js-adapter-webpack/src/index.ts
```diff
@@ -0,0 +1,2 @@
+import { paraglide as unpluginParaglide } from "@inlang/paraglide-js-adapter-unplugin"
+export default unpluginParaglide.webpack
```
samuelstroschein, line 2: HAHAH I love it

inlang/source-code/paraglide/paraglide-js-adapter-unplugin/src/index.ts
```diff
@@ -0,0 +1,48 @@
...
+export const paraglide = createUnplugin((config: UserConfig) => {
+	const options = {
+		silent: false,
+		onInit: true,
+		...config,
+	}
+
+	const execute = () => {
+		exec(
+			`paraglide-js compile --project ${options.project} --outdir ${options.outdir}`,
```
samuelstroschein, line 23: hmm interesting. i would have called the `compile` directly and avoided the CLI. 

Are you certain that calling the binary will work in every environment? If not, calling `compile` directly seems like a safer option

felixhaeberle: I implemented it like that without further thinking about it. Calling compile directly makes probably more sense.

LorisSigrist:
To avoid the case that paraglide has never been run and a build is triggered, I now synchronously run a compile on buildStart. This seems to fix it.

The `paraglide-js-adapter-vite` now has a test that involves building an entire app & that passes. 
I believe it makes sense to test in the vite-adapter, since it's the most feature complete. If that one works, the other ones will work aswel.

samuelstroschein:
> I believe it makes sense to test in the vite-adapter, since it's the most feature complete. If that one works, the other ones will work aswel.

Why not the unplugin adapter given that all other adapters are build on that one?

LorisSigrist:
The unplugin plugin can't be used in a bundler directly, it only provides you with all the other plugins. 

Eg:
```ts
const vitePlugin = unpluginPlugin.vite;
const rollupPlugin = unpluginPlugin.rollup;
const esbuildPlugin = unpluginPlugin.esbuild;
// ...
```

LorisSigrist:
Since we have to go though one of those plugins anyway, I figured vite would be the best one

samuelstroschein:
Understood!

Can you check if you get HMR reload working/avoiding the throw?

inlang/source-code/paraglide/paraglide-js-adapter-unplugin/src/index.ts
samuelstroschein: @LorisSigrist I recommend directly using the compile function from paraglide. What do you think?

LorisSigrist:
@samuelstroschein If we use the compile function directly, we end up with duplicate file-handling logic in the plugin & the cli. Also, it currently isn't exported from `@inlang/paraglide-js`

PR #1674, opened 2023-11-21
Add support for forcing languages via an options argument

### LINKED ISSUE #1567: add options `{ languageTag: languageTag }` to each message for out-of-component tree contexts
Required for server environments/invocations outside the rendering tree. 

```ts
// signature
m.name(params, options)
```

```ts
// example
m.hello_world({ name: "Samuel" }, { languageTag: "en" })
```

LorisSigrist:
Closes #1567 

This PR expands Paraglide's compiler output to include a second `options` parameter on the message index functions. This can be used to force a specific language regardless of the `languageTag()`.

```ts
setLanguageTag("en");
m.hello({ name: "Loris" }, { languageTag: "de" }) // "Hallo Loris"
```

This parameter is only present on the index-message functions, not the language specific ones. (would be pretty redundant there too).

If a message does not take any parameters you can pass either `{}` or `undefined` as a first parameter. 

`options.languageTag` is exactly typed with the available `langaugeTags`.


.changeset/perfect-scissors-invent.md
```diff
@@ -0,0 +1,5 @@
+---
+"@inlang/paraglide-js": patch
+---
+
+feat: Add options argument to messageIndex function, to allow forcing a locale
```
samuelstroschein, line 5: Can you write the changelog manually? 

Changeset doesn't work with paraglide at the moment because of the pre-release stuff. Besides, this is a minor bumb not a patch under SemVer. Good news, we decided to get rid of patches altogether and just have `MAJOR-MINOR` where `MAJOR` is a breaking change/needs user attention. MINOR is an update/new feature/fix/whatever that does not require user attention.

samuelstroschein (changes requested):
I don't see tests whether the implementation works in `compile.test.ts`. Please add tests :)

inlang/source-code/paraglide/paraglide-js/src/compiler/compile.test.ts
```diff
@@ -221,6 +221,33 @@ describe("e2e", async () => {
...
+	test("should return the correct message if a languageTag is set in the message options", async () => {
+		const { m, runtime } = await import(
+			`data:application/javascript;base64,${Buffer.from(
+				compiledBundle.output[0].code,
+				"utf8"
+			).toString("base64")}`
+		)
+
+		// set the language tag to de to make sure that the message options override the runtime language tag
+		runtime.setLanguageTag("de")
```
samuelstroschein, line 234: Change the test to (more is not needed)

```ts
runtime.setLanguageTag("de")
expect(m.onlyText()).toBe("Eine einfache Nachricht.")
expect(m.onlyText(undefined, { languageTag: "en" })).toBe("A simple message.")
```

- ensure that the same message still works without the language tag option
- verifies the language tag option

.changeset/perfect-scissors-invent.md
LorisSigrist: Ok. I removed the changeset and added a section to the changelog myself. Does this mean I also have to bump the package version manually?

I also added a test in `compile.test.ts`

LorisSigrist:
> @LorisSigrist are we confident about the options API/can we come up with a better solution?
> 
> The `onlyText(undefined, { languageTag: "en"})` looks odd but maybe that "oddness" is OK

I agree that it looks a little odd. Alternatively you can also pass `{}` as the first argument (already works). 
I don't think we can make this less awkward without some breaking API changes

LorisSigrist:
I changed the test like you specified

.changeset/perfect-scissors-invent.md
samuelstroschein: @LorisSigrist yes. Only I can publish atm. If you write the release notes and merge, I can bump the version and publish manually.

inlang/source-code/paraglide/paraglide-js/CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # @inlang/paraglide-js
 
+## 1.0.0-prerelease.10
+Add an optional options argument to message function, to allow forcing a locale, regardless of the current language tag.
+
+```ts
+import * as m from "./paraglide/messages"
+const msg = m.hello({name: "John"}, { locale: "de" })
```
samuelstroschein, line 8: locale or languageTag ;)

inlang/source-code/paraglide/paraglide-js/src/compiler/compile.test.ts
```diff
@@ -370,6 +393,18 @@ test("typesafety", async () => {
 
     // a message without params shouldn't require params
     m.onlyText() satisfies string
+
+
+	// --------- MESSAGE OPTIONS ---------
+	// the languageTag option should be optional
+	m.onlyText({}, {}) satisfies string
```
samuelstroschein, line 400: Good test! :)

inlang/source-code/paraglide/paraglide-js/src/compiler/compile.test.ts
```diff
@@ -370,6 +393,18 @@ test("typesafety", async () => {
...
+
+	// the languageTag option should be allowed
+	m.onlyText({}, { languageTag: "en" }) satisfies string
+
+	// the languageTag option must be a valid language tag
+	// @ts-expect-error - invalid language tag
+	m.onlyText({}, { languageTag: "---" })
```
samuelstroschein, line 407: Also good test! This is the quality we need/want :)

samuelstroschein (changes requested):
Minor typo in the changelog. Can be merged afterwards! :)

samuelstroschein (approved):

PR #1679, opened 2023-11-21
Add Telemetry to the ParaglideJS CLI

LorisSigrist:
This PR adds telemetry to the paraglide CLI. The runtime remains unaffected. 
It tracks executions of `paraglide-js init` and `paragrlide-js compile`.

Thanks to @jannesblobel for helping me get this working!

### DIFF
--- a/inlang/source-code/paraglide/paraglide-js/package.json
+++ b/inlang/source-code/paraglide/paraglide-js/package.json
@@ -40,6 +40,7 @@
 	},
 	"dependencies": {
 		"@inlang/detect-json-formatting": "1.0.0",
+		"@inlang/telemetry": "*",
 		"@inlang/sdk": "*",
 		"commander": "11.1.0",
 		"consola": "3.2.3",
--- a/inlang/source-code/paraglide/paraglide-js/package.json
+++ b/inlang/source-code/paraglide/paraglide-js/package.json
@@ -40,12 +40,14 @@
 	},
 	"dependencies": {
 		"@inlang/detect-json-formatting": "1.0.0",
-		"@inlang/telemetry": "*",
 		"@inlang/sdk": "*",
+		"@inlang/telemetry": "*",
+		"@inlang/env-variables": "*",
 		"commander": "11.1.0",
 		"consola": "3.2.3",
 		"dedent": "1.5.1",
-		"json5": "2.2.3"
+		"json5": "2.2.3",
+		"posthog-node": "3.1.3"
 	},
 	"devDependencies": {
 		"@rollup/plugin-terser": "0.4.3",
@@ -53,9 +55,9 @@
 		"@ts-morph/bootstrap": "0.20.0",
 		"@types/minimist": "1.2.3",
 		"@vitest/coverage-v8": "0.34.3",
-		"typescript": "5.2.2",
 		"memfs": "4.6.0",
 		"rollup": "3.29.1",
+		"typescript": "5.2.2",
 		"vitest": "0.34.3"
 	}
 }
--- a/inlang/source-code/paraglide/paraglide-js/src/cli/commands/compile.ts
+++ b/inlang/source-code/paraglide/paraglide-js/src/cli/commands/compile.ts
@@ -1,9 +1,11 @@
 import { loadProject, type InlangProject } from "@inlang/sdk"
+import { parseOrigin, telemetryNode } from "@inlang/telemetry"
 import consola from "consola"
 import { compile } from "../../compiler/compile.js"
 import fs from "node:fs/promises"
 import { resolve } from "node:path"
 import { Command } from "commander"
+import { getGitRemotes } from "../../utils/git.js"
 
 export const compileCommand = new Command()
 	.name("compile")
@@ -13,15 +15,38 @@ export const compileCommand = new Command()
 	.action(async (options: { project: string; outdir: string }) => {
 		consola.info(`Compiling inlang project at "${options.project}".`)
 
-		const path = resolve(process.cwd(), options.project)
-
+		const settingsFilePath = resolve(process.cwd(), options.project)
 		const project = exitIfErrors(
 			await loadProject({
-				settingsFilePath: path,
-				//@ts-ignore
+				settingsFilePath,
 				nodeishFs: fs,
+				_capture(id, props) {
+					telemetryNode.capture({
+						event: id,
+						properties: props,
+						distinctId: "unknown",
+					})
+				},
 			})
 		)
+
+		//For Telemetry
+		const gitOrigin = parseOrigin({
+			remotes: await getGitRemotes({ nodeishFs: fs, filepath: settingsFilePath }),
+		})
+
+		telemetryNode.capture({
+			event: "Paraglide compile",
+			groups: { repository: gitOrigin },
+			distinctId: "unknown",
+		})
+		telemetryNode.groupIdentify({
+			groupType: "repository",
+			groupKey: gitOrigin,
+			properties: {
+				name: gitOrigin,
+			},
+		})
 
 		const output = compile({
 			messages: project.query.messages.getAll(),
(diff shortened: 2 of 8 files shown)


jannesblobel (approved):

inlang/source-code/paraglide/paraglide-js/src/services/telemetry/events.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Typesafe telemetry events.
+ *
+ * Exists to avoid typos/always set the correct event name and properties.
+ */
+export type TelemetryEvents = "Paraglide compile executed" | "Paraglide init executed"
```
jannesblobel, line 6: Could you change it to Paraglide-js...

inlang/source-code/paraglide/paraglide-js/src/cli/commands/init.ts
```diff
@@ -22,6 +23,7 @@ export const initCommand = new Command()
 	.summary("Initializes inlang Paraglide-JS.")
 	.action(async () => {
 		consola.box("Welcome to inlang Paraglide-JS 🪂")
+		telemetry.capture({ event: "Paraglide init executed" })
```
jannesblobel, line 26: same here Paraglide-js init

inlang/source-code/paraglide/paraglide-js/src/cli/commands/compile.ts
```diff
@@ -13,16 +14,25 @@ export const compileCommand = new Command()
...
+						// @ts-ignore the event types
+						event: id,
+						properties: props,
+					})
+				},
 			})
 		)
 
+		telemetry.capture({
+			event: "Paraglide compile executed",
```
jannesblobel, line 33: same here Paraglide-js compile ...

jannesblobel (changes requested):

jannesblobel (approved):

inlang/source-code/paraglide/paraglide-js/package.json
```diff
@@ -41,20 +41,23 @@
 	"dependencies": {
 		"@inlang/detect-json-formatting": "1.0.0",
 		"@inlang/sdk": "*",
+		"@inlang/telemetry": "*",
+		"@inlang/env-variables": "*",
```
samuelstroschein, line 45: 1. @inlang/env-variables is a private package. you can't use it in dependencies
2. @inlang/telemetry is also a private package

I recommend removing both dependencies and hardcode the endpoint + public access key to push events. that should be sufficient.

inlang/source-code/paraglide/paraglide-js/src/cli/commands/compile.ts
```diff
@@ -13,16 +14,25 @@ export const compileCommand = new Command()
...
+						// @ts-ignore the event types
+						event: id,
+						properties: props,
+					})
+				},
 			})
 		)
 
+		telemetry.capture({
+			event: "Paraglide-js compile executed",
```
samuelstroschein, line 33: We use all caps for the app. 

```ts
"PARAGLIDE-JS compile executed"
```

samuelstroschein: If you add the capture in the CLI command, you will have to re-implement the the capture in every code that calls the `compile` function (vite plugin, etc.)

A better solution might be to call `PARAGLIDE-JS compile executed` in the compile function itself.

inlang/source-code/paraglide/paraglide-js/src/cli/commands/init.ts
```diff
@@ -22,6 +23,7 @@ export const initCommand = new Command()
 	.summary("Initializes inlang Paraglide-JS.")
 	.action(async () => {
 		consola.box("Welcome to inlang Paraglide-JS 🪂")
+		telemetry.capture({ event: "Paraglide-js init executed" })
```
samuelstroschein, line 26: `PARAGLIDE-JS init executed`

package.json
```diff
@@ -46,5 +46,8 @@
 		"nx-cloud": "16.4.0",
 		"prettier": "2.8.3",
 		"typescript": "5.2.2"
+	},
+	"dependencies": {
+		"isomorphic-git": "1.25.0"
```
samuelstroschein, line 51: What happened here?

samuelstroschein (changes requested):
Besides the feedback beneath, I am wondering if we should bundle the CLI. 

- all dependencies except for @inlang/sdk are for the CLI
- bundling the CLI enables usage of private packages like @inlang/env-variables

@LorisSigrist what do you think?

inlang/source-code/paraglide/paraglide-js/package.json
samuelstroschein: > Just to check, are you aware that @inlang/cli has those same dependencies?

Yes. The CLI is bundled. Thus, the issue does not apply there (we could also bundle the paraglide cli)

LorisSigrist: I believe bundling is the way to go here. If we hardcode the token there are some serious drawbacks:
- No separation between dev & production tokens
- Almost guaranteed to be forgotten if the posthog token ever changes

package.json
LorisSigrist: Whoops, must have run an install command in the wrong folder. That's not intentional

inlang/source-code/paraglide/paraglide-js/package.json
samuelstroschein: Do you want to add bundling in this PR? 

Btw, please only the CLI. Don't bundle the "library".

jbergstroem:
Perhaps it could benefit from an opt-out?

inlang/source-code/paraglide/paraglide-js/package.json
```diff
@@ -44,17 +44,24 @@
 		"commander": "11.1.0",
 		"consola": "3.2.3",
 		"dedent": "1.5.1",
-		"json5": "2.2.3"
+		"isomorphic-git": "1.24.5",
+		"json5": "2.2.3",
+		"posthog-node": "3.1.3"
```
samuelstroschein, line 49: You can move all those dependencies to devDependencies if you bundle the CLI

inlang/source-code/paraglide/paraglide-js/src/cli/commands/init.ts
```diff
@@ -22,6 +23,7 @@ export const initCommand = new Command()
 	.summary("Initializes inlang Paraglide-JS.")
 	.action(async () => {
 		consola.box("Welcome to inlang Paraglide-JS 🪂")
+		telemetry.capture({ event: "PARAGLIDE-JS init executed" })
 
 		await checkIfUncommittedChanges()
```
samuelstroschein, line 28: I would recommend multiple events to have an insight if the init command worked: 

```
1. PARAGLIDE-JS init started
2. ... we can add various events here to build a funnel and see where people drop off
3. PARAGLIDE-JS init finished

samuelstroschein (changes requested):

inlang/source-code/paraglide/paraglide-js/src/cli/commands/init.ts
LorisSigrist: Will do!

jannesblobel:
@jbergstroem 
> Perhaps it could benefit from an opt-out?

This is definitely planned for the future. We are currently implementing this to get a better understanding of whether the commands work, whether Paraglide works, and whether people are using it.

samuelstroschein (approved):

samuelstroschein:
@LorisSigrist can we merge this? 

cc @jannesblobel

AdamTmHun:
Will there be a way to disable telemetry?

samuelstroschein:
@AdamTmHun yes, see https://github.com/inlang/monorepo/pull/1679#issuecomment-1825782380.

PR #1698, opened 2023-11-22
Clean up .gitignore

openscript:
Main reason for touching it:
 - Inside the devcontainers the `.pnpm-store` lives in the projects workspace root, which must not be pushed to git

The majority of the rules come from gitignore.io and at the top is now a section for inlang specific rules.


samuelstroschein (approved):

LorisSigrist (approved):
While this does include a lot of stuff we'll likely never need to deal with, it's better safe than sorry and probably won't break stuff. 
LGTM

PR #1706, opened 2023-11-22
bail on errors to stop swallowing error messages on critical nx commands

janfjohannes:
please check if this solves the issue and does not break a workflow that requires nx to keep running on errros, but i cannot think of a situation where that is required.


openscript (changes requested):
The reason is not more visible, but it seems like it's now the only failure.

[Kooha-2023-11-22-20-19-32.webm](https://github.com/inlang/monorepo/assets/1105080/56b07618-1ffe-4476-a921-e8d204a620ff)

I would also add the `npm run fetchPublicEnv` step to `CONTRIBUTING.md`  and change it to pnpm.

samuelstroschein (approved):

openscript (approved):

PR #1708, opened 2023-11-22
chore: move examples back into respective folders

samuelstroschein:
I found it "hard" to find the examples for paraglide and the adapters. 

Moving the examples back into an /example folder increases discoverability and does seem to work. 

If CI/CD is green, I will merge this. 

**Changes**

- moved examples into nested structure again to increase discoverability
- updated NX


samuelstroschein:
@LorisSigrist hm don't know what issues you faced with examples being in subpackages.

LorisSigrist:
During the pnpm migration I was dealing with an infinite crawler loop when building the website / editor. Following the stacktrace seemed to suggest that the module resolution got stuck in an infinite loop when resolving paraglide (which contained the examples). Moving the examples changed the error message which seemed like progress.

This was before I found out that Tailwind was the culprit.

PR #1713, opened 2023-11-23
[Paraglide] Fix Eslint not ignoring the output

### LINKED ISSUE #1712: [bug] Paraglide ESLint ignore does not work
### Problem

Paraglide's output should be ignored by developer tooling, like `git`, `eslint` etc. 
To do this we add some ignore files to the output, however this does not work for `eslint`. Eslint only considers `.eslintignore` files at the _root_ of a project, not in a subfolder like we're doing. 

We haven't noticed this before because the output would always pass linting, but #1674 necessitates some occasional unused variables. 

## Possible Solutions
- Add Eslintignore comments inside paraglide's build output
- Tell users to manually add `./src/paraglide` to their `.eslintignore`

### Expected behavior

Paraglide's output should not get linted.

### Reproduction

Go to `./inlang/source-code/editor`
Run build, then lint

### Other information

_No response_

LorisSigrist:
Closes #1712 

This PR changes how we tell eslint to ignore the paraglide output directory.
- Removes the `.eslintignore` since it doesn't work outside the current working directory (usually root) ([Source](https://eslint.org/docs/latest/use/configure/ignore#the-eslintignore-file))
- Adds `/* eslint-disable */` comments to each JS file in the output

### DIFF
--- a/inlang/source-code/paraglide/paraglide-js/src/compiler/compile.test.ts
+++ b/inlang/source-code/paraglide/paraglide-js/src/compiler/compile.test.ts
@@ -22,11 +22,14 @@ test("the files should include a prettierignore file", async () => {
 	expect(output).toHaveProperty(".prettierignore")
 	expect(output[".prettierignore"]).toContain("*")
 })
-// ignore eslint stuff
-test("files should include an eslint ignore", async () => {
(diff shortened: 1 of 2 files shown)


samuelstroschein (approved):
Good change!

PR #1715, opened 2023-11-23
Improve ParaglideJS Documentation

### LINKED ISSUE #1689: Improve Paraglide docs
from https://github.com/inlang/monorepo/issues/1667

## Problem 
Feedback from @KraXen72:
"i am not sure if this is the relevant issue for it, but it took me an hour or two to figure the following things out:

- npx @inlang/paraglide-js@latest init is apparently either supposed to install @inlang/paraglide-js to your project (even though npx does not imply installation usually, or you are supposed to instlal it manually (not mentioned anywhere). i ran the command with the pnpm package manager, pnpm dlx @inlang/paraglide-js@latest init and while it did set up the inlang project sucessfully, it did not add the dependency.
also, is @inlang/paraglide-js supposed to be instlled as a devDependency or normal one?
- it took me very long to figure out the import from ./paraglide-js/messages is intentional (it is also wrong)
what worked for me was importing form (whatever relative path to src is)/paraglide/messages (not paraglide-js)
no where it is mentioned that inlang compilation creates a folder called paraglide in the source folder upon build
i think the init script should automatically add an alias (vite or sveltekit alias, idk about other frameworks tho) to src/paraglide as something like $paraglide or @paraglide, and you should mention that this is an implicity created, invisible folder
- in the svelte summit talk, you guys mentioned that there are several translation formats supported, but only json works so far - but i see yaml is being tracked too, so that's good. would be nice to see more formats down the road"

@inlang/martketplace-infrastructure @inlang/paraglide-js

LorisSigrist:
Closes #1689

We got some feedback about mistakes and omissions in the ParaglideJS documentation. This PR aims to implement that feedback.

- It corrects the imports from `./paraglide-js/messages` to `./paraglide/messages`
- Explicitly informs users about the `src/paraglide` folder that gets created when compiling
- Suggests users should add an alias pointing to the `src/paraglide` folder if using a bundler
- Elaborates on the functions provided in the runtime.

Our end-goal is to abstract away the usage of the paraglide runtime as much as possible using adapters. However, until we have a bunch of adapters ready to go, educating users in it's use is required.

### DIFF
--- a/inlang/source-code/paraglide/paraglide-js/README.md
+++ b/inlang/source-code/paraglide/paraglide-js/README.md
@@ -93,7 +93,7 @@ Paraglide JS provides five exports in `./paraglide/runtime.js`:
 
 | Variable | Description |
 | --- | --- |
-| `sourceLanguageTag` | The default language tag of the project |
+| `sourceLanguageTag` | The source language tag of the project |
 | `availableLanguageTags` | All language tags of the current project |
 | `languageTag()` | Returns the language tag of the current user |
 | `setLanguageTag()` | Sets the language tag of the current user |
--- a/inlang/source-code/paraglide/paraglide-js/README.md
+++ b/inlang/source-code/paraglide/paraglide-js/README.md
@@ -2,7 +2,7 @@
 
 [<img src="https://cdn.loom.com/sessions/thumbnails/a8365ec4fa2c4f6bbbf4370cf22dd7f6-with-play.gif" width="100%" /> Watch the pre-release demo of Paraglide JS](https://www.youtube.com/watch?v=-YES3CCAG90)
 
-Attention: The following features are missing and will be added in the upcoming weeks: 
+Attention: The following features are missing and will be added in the upcoming weeks:
 
 - [ ] Support for pluralization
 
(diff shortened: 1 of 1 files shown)


NilsJacobsen (approved):
Skimming through it, it looks good.

KraXen72:
it's a big improvement. here is my feedback:

the '3 lines of code setup' tagline is a bit misleading
- the config is not 3 lines
- the npx command is just one line
- cli commands aren't generally considered code? idk
i'd just go with easy setup wizard/whatever other frameworks call it, like quickstart

the recommended import path in vite does not use a glob, like you mentioned. you would just set resolve.alias.$paraglide to 'src/paraglide' (atleast  that's how i understand it and set it up) although, they recommend absolute paths? it's a bit confusing.
take a look at https://vitejs.dev/config/shared-options#resolve-alias
or some examples, and i think a very simple 'diff' markdown codeblock adding the alias under this section would help

the usage and architecture sections are good, it taught me how to make parameters in translations

there is also a type export from runtime containing the current languages, so it would be atleast 6 exports.

instead of a manually written api reference, i'd put this description info into jsdoc of the runtime, and use some of the many jsdoc => documentation generators to have a nice, auto-generated api reference, that you would link to. (one of the tools is https://paka.dev/)

LorisSigrist:
Thanks! I'll take a look

inlang/source-code/paraglide/paraglide-js/README.md
```diff
@@ -79,23 +86,62 @@ m.loginHeader({ name: "Samuel" }) // Hello Samuel, please login to continue.
...
+m.loginHeader({ name: "Samuel" }) // Hallo Samuel, bitte melde dich an, um fortzufahren.
 ```
-Paraglide JS exports four variables and functions via "@inlang/paraglide-js":
+
+Paraglide JS provides five exports in `./paraglide/runtime.js`:
 
 | Variable | Description |
 | --- | --- |
-| `sourceLanguageTag` | The source language tag of the project |
+| `sourceLanguageTag` | The default language tag of the project |
```
samuelstroschein, line 96: why default? it's the source language tag

LorisSigrist: I changed it back

PR #1716, opened 2023-11-23
[Breaking] Make `--project` a required argument for `paraglide-js compile`

### LINKED ISSUE #1508: `paraglide-js compile` remove default path
## Problem 

@NilsJacobsen was unaware that the project location can be specified with --project. 

CleanShot 2023-10-23 at 12 05 47@2x

LorisSigrist:
Closes #1508 

It wasn't obvious that you could provide the location of the project file to `paraglide-js compile`, so it was decided to make it a required argument.

This PR:
- Makes `--project` a required argument for `paraglide-js compile`
- Updates all our build commands to include the `--project` flag

### DIFF
--- a/inlang/source-code/paraglide/paraglide-js-adapter-sveltekit/example/package.json
+++ b/inlang/source-code/paraglide/paraglide-js-adapter-sveltekit/example/package.json
@@ -4,8 +4,8 @@
 	"private": true,
 	"scripts": {
 		"_dev": "vite dev",
-		"build": "paraglide-js compile && vite build",
-		"test": "paraglide-js compile && vite build",
+		"build": "paraglide-js compile --project ./project.inlang.json && vite build",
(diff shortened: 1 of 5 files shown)


LorisSigrist:
The `paraglide-js init` command already explicitly specified the `--project` in the build script, but if that was manually removed this would be a breaking change

felixhaeberle:
Yep, more explicit is better 👍

samuelstroschein:
In favor of this change which I requested a few days ago :D https://github.com/inlang/monorepo/issues/1508#issuecomment-1819284099

@LorisSigrist in case you are not using inbox zero to stay ahead of notifications, and don't forget to reply to notifications, I recommend this loom https://www.loom.com/share/252cc168ca174bf7aef6d4c57036050d

samuelstroschein (approved):

PR #1718, opened 2023-11-23
rename id's of syntax matcher

felixhaeberle:
this PR solves the mystery "Expected union value" error. This error has to be improved sometime in the future. Probably there is also coming something from upstream: https://github.com/sinclairzx81/typebox/issues/640


NilsJacobsen (approved):
Great, thanks for the investigation

Issue #1570, opened 2023-11-01
prompts the user for the tech stack (js, sveltekit, nextjs) which forwards to the guide on how to get startedd


samuelstroschein:

felixhaeberle:
@samuelstroschein @LorisSigrist what's the status here?

LorisSigrist:
I built a framework detection that can be used to implement this (#1676). But until we have some adapters, it doesn't make sense to merge it.

samuelstroschein:
> I built a framework detection that can be used to implement this

I highly highly highly recommend to just prompt the user: 

```
What framework are you using? 

- NextJS
- SvelteKit
- ...
```

Anything auto-detecting is a recipe to go wrong :D Especially for something so trivial. When a user selects the framework, I would just forward to the adapter site on inlang.com

LorisSigrist:
That makes sense, I'll change it to prompting once we have some adapters

LorisSigrist:
Now that we have some adapters I would build this to make onboarding easier.

Something we might want to consider is **not** adding the `paraglide-js compile` commands to the `build` script if people choose a stack where a build-system plugin is available.

PR #1721, opened 2023-11-23
`paraglide-js init` detect tech-stack and suggest adapter

LorisSigrist:
As described in #1570 

**Don't merge until #1669 is merged**

This PR extends the `paraglide-js init` command.
After completing it's usually init process, it will try and detect which tech-stack is being used by reading `package.json` and suggest and adapter.


LorisSigrist:
In #1570 it was decided that prompting the user for which tool they are using makes more sense.

PR #1733, opened 2023-11-24
implementation of inlang/manage

flornkm:
manage.inlang.com will be the new place for users to
- install modules
- uninstall modules (coming soon)
- update modules (coming soon)
- … and more in the future!


inlang/source-code/server/src/main.ts
```diff
@@ -87,6 +89,17 @@ app.use(
...
+	"/manage",
+	createProxyMiddleware({
+		target: manageAdress,
+		changeOrigin: true,
+		headers: {
+			Connection: "keep-alive",
+		},
+	})
+)
+
```
samuelstroschein, line 102: if we want to host manage on `manage.inlang.com`, we don't need this proxy server

flornkm: @samuelstroschein but manage.localhost:3000 isn't working right? It's just for development, in production this isn't used.

samuelstroschein: @flornkm correct. you need to open the correct localhost port directly. i don't find resources what better subdomain (manage.inlang.com) or subpath (inlang.com/manage). somewhat ambiguous decision here. 

**arguments for subdomain**

- better UX because users can type `manage*` and get autocomplete
- separation of the teams (no reliance on a common proxy server) 

**arguments for subpath**

- easier to deploy (?) no other domain handling etc

flornkm: Got it, I just thought it would be a better DX if we don't have to redirect to another local port.

Regarding the decision, let's go with subdomain. Additional arguments:

- subpaths on inlang.com are almost entirely SEO optimized, `inlang/manage` should be not indexed. So a subdomain is a better separation in my opinion.

flornkm:
> @flornkm sneak peak: how does inlang manage look like? 🤩

Will update tomorrow probably, have some problems with passing objects in Lit (you can't do it like in JSX 🙂)

flornkm:
@felixhaeberle a lot of bugfixing with jan today for the logic, therefore hadn't time to work much more on the UI today. Here's just a small preview as we can focus on the core of the manage app: handle what's in your project

PS: There is a "Cancel" button below the loading indicator. I think it is good to stop the process if you decide to not install a certain module anymore


https://github.com/inlang/monorepo/assets/92092993/fbc5b8af-0b2f-4643-b82d-608c8348c6ee

felixhaeberle:
why cancel? we currently have not that proud'ish loading times so I might not add a cancel which could be clicked if it does simply take too long for the user 😄

flornkm:
@samuelstroschein that was the first (an easiest) thing I've implemented :D

flornkm:
LOOM: https://www.loom.com/share/bd842e3d50b0413ea6c1076ddf8697fe

NilsJacobsen (approved):

Issue #1705, opened 2023-11-22
re-enable nx cloud


samuelstroschein:
## Problem 

We disabled nx cloud because nx cloud cached env variables from dev to production. 

## Proposal

Let nx cloud invalidate the cache when env variables change 

## Additional information

- https://github.com/inlang/monorepo/pull/1325
- https://github.com/inlang/monorepo/issues/1693

felixhaeberle:
@LorisSigrist Do you want to take over in conjunction with fixing the GitHub Actions?

LorisSigrist:
I can do that.
That means enable nx cloud & make sure that it get's invalidated when environment variables change

Issue #1693, opened 2023-11-22
[refactor] Improve Github Actions efficiency


LorisSigrist:
## Problem 
Currently we do a lot of work multiple times for each action. 
For example: `ci` get's called in three separate actions. All of them have a reason to do so.
- `continuous integration` Calls CI (obviously)
- `versioning` Calls CI to not create changesets for broken
- `npm publish` Calls CI to not publish broken packages. 

## Proposal
Reorganize the Actions so that `versioning` and `npm publish` are dependent on the `continuous integration` action succeeding, so that they can trust that they are working with a working version without having to recheck.

This would roughly cut the CI work in half.

samuelstroschein:
We just need to enable NX cloud again, and the issue is resolved. I will put this one my todo list.

samuelstroschein:
Good proposal. 

In addition, we should re-enable NX cloud https://github.com/inlang/monorepo/issues/1705

PR #1748, opened 2023-11-26
[CI] Re-enable Nx cloud

LorisSigrist:
#1705 

This PR re-enables Nx cloud caching. 
We avoid issues with changing Env variables by including them when NX calculates the hash for the cache, so changes in our Env variables will cause the cache to be invalidated. 

The token included in `nx.json` is read-only, meaning that it can't update the cache. A read-write token is made injected to our developers & CI pipeline via doppler. This keeps random people from messing with our cache. 

(It took me way too long to figure out that I needed to update `nx-cloud`. The error message is very well hidden)

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -1,22 +1,85 @@
 {
-	"tasksRunnerOptions": {
-		"default": {
-			"runner": "nx/tasks-runners/default",
-			"options": {
-				"cacheableOperations": ["build", "test", "lint", "format"],
-				"accessToken": "ODlkYjAwMjYtMGMyMS00YTI3LWIzNTAtN2RhZjgwYzAwZGJifHJlYWQtd3JpdGU="
-			}
-		}
-	},
+	"$schema": "./node_modules/nx/schemas/nx-schema.json",
 	"targetDefaults": {
 		"build": {
-			"dependsOn": ["^build"]
+			"dependsOn": [
+				"^build"
+			],
+			"inputs": [
+				"default",
+				"^default",
+				{
+					"env": "ALGOLIA_ADMIN"
+				},
+				{
+					"env": "ALGOLIA_APPLICATION"
+				},
+				{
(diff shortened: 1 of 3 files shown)


nx.json
```diff
@@ -1,22 +1,85 @@
...
+				},
+				{
+					"env": "PUBLIC_GIT_PROXY_BASE_URL"
+				},
+				{
+					"env": "PUBLIC_LIX_GITHUB_APP_CLIENT_ID"
+				},
+				{
+					"env": "LIX_GITHUB_APP_CLIENT_SECRET"
+				},
```
samuelstroschein, line 31: There is no better way to invalidate the cache? The hardcoded env variables will eventually run out of sync and cause regression bugs that are hard to debug

LorisSigrist: Actually, I did just have a better Idea. 
We could use a runtime input & use the output of `doppler secrets` as an input for the hash. Will try that tomorrow.
https://nx.dev/recipes/running-tasks/customizing-inputs#include-runtime-information-in-the-cache-hash

samuelstroschein: Let's try that!

LorisSigrist:
@samuelstroschein 
This PR now works as advertised and as you can see from the Execution times, the execution-time savings are huge. 

Before I merge this I wanted to confirm that you are aware of NX Cloud's billing plans and the cost associated with enabling NX cloud. 

https://nx.app/pricing

It might be worth delaying merging this until #1693 is implemented, since the actions on the main branch could potentially cause 3 CI pipelines to be billed, instead of just one. What do you think?

samuelstroschein:
@LorisSigrist good that you raised awareness for the pricing. 

The pricing is OK. The speed we gain outweighs the cost. Besides, we are on some open source plan and have a discount i think. 

MERGE! :)

samuelstroschein (approved):

PR #1752, opened 2023-11-27
add quote stripping to extract message of vs code extension

felixhaeberle:
This PR aims to add quote stripping to extract message of vs code extension.


samuelstroschein:
@felixhaeberle, what should i review? 

This seems like a straightforward fix that does not touch the SDK. As long as it is not touching the SDK or some fundamental API changes, putting me in the review is likely not required/I am wondering what I should review.

PR #1753, opened 2023-11-27
add resolve string escape for inline preview

felixhaeberle:
This PR aims to add resolve string escape for inline preview.


samuelstroschein:
See https://github.com/inlang/monorepo/pull/1752#issuecomment-1827892738. 

I don't think a review from myself is required :)

Issue #1772, opened 2023-11-28
Reactivity/State management in the SDK improvement collection


samuelstroschein:
## Problem 

We need to find a better suited state (reactive) management solution for the SDK. 

## Context

The SDK currently uses (Solid) Signals for state management/reactivity. We (knew we would) face numerous issues that pimple up now to the extent that we are inclined to take action. 

## Requirements

- [ ] simple(r) way to prevent/coordinate multiple executions
-- #1740, 
-- https://github.com/inlang/monorepo/issues/1769
- [ ] lazy execution/evaluation for lints
-- https://github.com/inlang/monorepo/issues/1413
- [ ] avoid reactive/not reactive breaking change risk
-- #1680

KraXen72:
using solid signals for an sdk is kind of crazy lol. 
surprised it managed to live for this long 😅

NiklasBuchfink:
I think this is more of a cleaning task and making sure everything is set and passed as it should. We didn't even use stores, because it is complex to understand, but this could help eliminate this issue with handling complex data like our linting.

There is room for improvement. I dug more into it in the last days with podcasts and having the Solid docs as good night reading :)

samuelstroschein:
> using solid signals for an sdk is kind of crazy lol.
> surprised it managed to live for this long 😅

The SDK has the odd requirement that close to every call is reactive. 

Getting the name of a project, querying a message, getting project settings, installing a plugin is/are all reactive properties that should bind to a GUI without leading to constant re-renders. Otherwise, the fink editor and parrot will be noticeably slow. 

The perfect match for our use case does not seem to exist. The question is: What solution comes somewhat close?

samuelstroschein:
Some research over the holidays: 

- we need an async runtime/state management/reactive programming solution 
- scala has multiple e.g. https://typelevel.org/cats-effect/ and https://zio.dev/
- JS ecosystem seems empty handed except for RxJS and the brand new Effect (influenced from Scalas ZIO) https://effect.website/

samuelstroschein:
i just got off a call with Datner regarding effect ts , see his comment [here](https://www.reddit.com/r/typescript/comments/16w3iwn/comment/k42tu1h/?utm_source=share&utm_medium=web2x&context=3) 

- effect ts solves most problems but is not fully stable yet, requires full buy-in and is complex :/
- we need observables/signals somewhere to expose changes to UI frameworks 
- RxJS would be the go-to. 
- we should take a look at xstate 
[...]

PR #1789, opened 2023-11-30
Granular linting

NiklasBuchfink:
This PR adds granular linting to the SDK to reduce performance issues and enable synchronized updates.

- [x] Fix executing loadProject multiple times within Fink
- [x] Create reactiveMap for messageLintQuery
- [x] Remove debounce
- [x] Handle adding new messages for Paraglide JS and VS code extension 
- [x] Non-blocking lint calculation / lazy-linting by debouncing
- [x] Update & pass all tests
- [x] Make sure other apps still work


inlang/source-code/editor/src/pages/@host/@owner/@repository/State.tsx
```diff
@@ -208,9 +208,6 @@ export function EditorStateProvider(props: { children: JSXElement }) {
 	// get lix errors
 	const [lixErrors, setLixErrors] = createSignal<ReturnType<Repository["errors"]>>([])
 	createEffect(() => {
-		// reset errors to empty on repo changes
-		setLixErrors([])
-
 		repo()?.errors.subscribe((errors) => {
 			setLixErrors(errors)
 		})
```
samuelstroschein, line 213: @NiklasBuchfink why can't you directly subscropbe to `repo().errors`? 

Does no wrapper for repo reactivity exist?

NiklasBuchfink: You are right, moved it behind the repo resource and removed the effect and reset. Everything works, including the resets after changing the branch 👍

NiklasBuchfink:
@samuelstroschein The last difficult step is now the non-blocking linting calls. This will save up to one minute of startup time with extremely large repos (in terms of translations).

samuelstroschein:
@NiklasBuchfink we will look into this as part of #455

NiklasBuchfink:
> @NiklasBuchfink we will look into this as part of #1772

Then I have to debounce the write file again because updating a single message with more than 20 languages and four lint rules has already become a performance problem.

The route course of #1772:
Reactivity with SolidJS is granular; it just needs to be engineered that way.
The file watcher is **not** granular and therefore stupidly recreates everything instead of updating the change diff here and there in the reactivity system.

inlang/source-code/sdk/src/createMessageLintReportsQuery.ts
```diff
@@ -1,50 +1,78 @@
...
-				}).then((report) => {
-					if (
-						report.errors.length === 0 &&
-						JSON.stringify(index.get(message.id)) !== JSON.stringify(report.data)
-					) {
-						index.set(message.id, report.data || [])
+	createEffect(() => {
+		if (rulesArray) {
+			for (const messageId of messagesQuery.includedMessageIds()) {
+				messagesQuery.get.subscribe({ where: { id: messageId } }, async (message) => {
```
samuelstroschein, line 47: why do you use subscribe? 

you can directly use `get()` without the subscribe callback.

inlang/source-code/sdk/src/createMessageLintReportsQuery.ts
```diff
@@ -1,50 +1,78 @@
...
+						await lintSingleMessage({
+							rules: rulesArray,
+							settings: settingsObject(),
+							messages: messages,
+							message: message,
+						}).then((report) => {
+							if (report.errors.length === 0 && index.get(messageId) !== report.data) {
+								index.set(messageId, report.data)
+							}
+						})
```
samuelstroschein, line 58: Why are you awaiting and using a `.then` callback? The `await` keyword is probably ignored. Decide for one. Given that you do not want to run lints when the project loads, `.then` seems to be the correct choice.

inlang/source-code/sdk/src/createMessageLintReportsQuery.ts
```diff
@@ -1,50 +1,78 @@
...
+									messages: messages,
+									message: message,
+								}).then((report) => {
+									if (report.errors.length === 0 && index.get(messageId) !== report.data) {
+										index.set(messageId, report.data)
+									}
+								})
+							},
+							{ atBegin: false }
+						)(message)
```
samuelstroschein, line 75: 1. Why a debouncer here?
2. Async then problem as above.

inlang/source-code/sdk/src/createMessageLintReportsQuery.ts
NilsJacobsen: image

I don't think so. It's not being converted to a signal. It only returns a value this way.

inlang/source-code/sdk/src/createMessageLintReportsQuery.ts
NiklasBuchfink: 1. Debouncing is required until this is implemented https://github.com/inlang/monorepo/issues/1817
**Background:** Linting is triggered on every key change; no longer on the debounced save. This works for projects with only a few languages and lint rules, but for projects with 20+ languages and 4 lint rules, it blocks the main thread and the pattern editor field lags massively behind (loss of real-time feedback).

2. Agreed, I'll remove it

inlang/source-code/sdk/src/createMessageLintReportsQuery.ts
NilsJacobsen: haha sdk trapped me :D

NiklasBuchfink: @samuelstroschein We changed this. It is now a signal within a createEffect.

inlang/source-code/sdk/src/createMessageLintReportsQuery.ts
```diff
@@ -44,7 +44,8 @@ export function createMessageLintReportsQuery(
 	createEffect(() => {
 		if (rulesArray) {
 			for (const messageId of messagesQuery.includedMessageIds()) {
-				messagesQuery.get.subscribe({ where: { id: messageId } }, (message) => {
+				createEffect(() => {
+					const message = messagesQuery.get({ where: { id: messageId } })
```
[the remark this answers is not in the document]
NiklasBuchfink: Its ugly but:
Effect 1. is triggered when a new message is created and all reports are created again (requirement for the VS code extension)
Effect 2. is triggered when a message is updated and a single new report is created.

PR #1813, opened 2023-12-04
feat: add changelog to product pages

flornkm:
https://github.com/inlang/monorepo/assets/92092993/67b8ef81-f296-4bf3-9c37-cd9c6d06d640


NilsJacobsen (approved):
I like it 👍 gives us the flexibility to add reviews / examples ... what every we need

PR #1819, opened 2023-12-04
Merge versioning & npm publish into CI workflow

### LINKED ISSUE #1693: [refactor] Improve Github Actions efficiency

LorisSigrist:
Closes #1693

This PR merges the `versioning` and `npm_publish` workflows into the main Ci workflow. They are both marked as dependent on the CI job succeeding, which saves the need for testing.  They only run on main, so pull-requests should not trigger publishes (also jobs running on pull requests don't get the necessary secrets anyways). 

Copying over the build artifacts would be nice, but I don't see a convenient way of doing that right now. We would likely need a centralised `dist` folder for the entire monorepo.


janfjohannes (approved):

LorisSigrist:
I would also prefer to do it in separate workflows. I didn't know that was possible. Let's do that then!

PR #1823, opened 2023-12-05
Minor Cleanup for the Solid Adapter

LorisSigrist:
@thetarnav just merged the solidstart adapter 🎉

This PR does some minor modifications to bring the setup more inline with the other adapters.
- Use `@inlang/paraglide-js-adapter-vite` for running the paraglide commands
- Simplify command setup
- Use `_dev` instead of `dev` to avoid the example project being run when the global `dev` command is used


thetarnav:
Thank you, looks good.

PR #1824, opened 2023-12-05
add `listProjects()` to SDK

flornkm:
This PR introduces a function for searching through a given filesystem with a depth of 3 directories.
The goal is to allow apps to find and list inlang projects e.g. in a GitHub Repo.


flornkm:
See a full flow with an example: https://github.com/inlang/monorepo/pull/1733#issuecomment-1840660416

inlang/source-code/sdk/src/listProjects.ts
```diff
@@ -0,0 +1,37 @@
+import type { NodeishFilesystem } from "@lix-js/fs"
+
+export const listProjects = async (
+	nodeishFs: NodeishFilesystem,
+	from: string,
+	recursionLimit: number = 3
```
samuelstroschein, line 6: I propose to remove the recursion limit as an argument. 

We will remove this argument in the future when we have caching. If we expose the argument and then remove it -> breaking change. Hence, I propose to hardcode a number (lets increase it to 5?).

flornkm: @samuelstroschein alright, makes sense

NiklasBuchfink (approved):

PR #1825, opened 2023-12-05
Update sdk docs

NilsJacobsen:
- Added a main nav where you can switch between docs goals (build app, build plugin, build lint-rule, publish content)

-> See https://github.com/inlang/monorepo/issues/1815


inlang/source-code/website/src/pages/documentation/index.page.tsx
```diff
@@ -143,32 +139,26 @@ export function Page(props: PageProps) {
...
-							>
+						<div class="flex flex-1 mb-8 md:ml-4 lg:ml-12 min-h-screen">
+							<div class="flex-1 md:max-w-[500px] lg:max-w-[724px] w-full justify-self-center md:col-span-3">
 								<Markdown markdown={props.markdown} />
 								<EditButton href={editLink()} />
 								<Feedback />
 							</div>
 						</div>
 					</Show>
+					<div class="top-[116px] hidden sticky xl:block h-[calc(100%_-_112px)] w-[230px] py-[60px] pl-4">
```
flornkm, line 150: I'm missing a overflow in y direction with the custom overflow scrollbar here in case we have too many anchors.

Like already introduced at the left navbar:
```tsx
<nav class="max-h-[96vh] overflow-y-scroll overflow-scrollbar -ml-4 pl-4">
```

inlang/source-code/website/src/pages/documentation/index.page.tsx
```diff
@@ -143,32 +139,26 @@ export function Page(props: PageProps) {
...
-								</Callout> */}
-							</Show>
-							<div
-								// change the col-span to 2 if a right side nav bar should be rendered
-								class="w-full justify-self-center md:col-span-3"
-							>
+						<div class="flex flex-1 mb-8 md:ml-4 lg:ml-12 min-h-screen">
+							<div class="flex-1 md:max-w-[500px] lg:max-w-[724px] w-full justify-self-center md:col-span-3">
 								<Markdown markdown={props.markdown} />
 								<EditButton href={editLink()} />
```
flornkm, line 145: The edit link isn't working for me. Ping me if I should take a look – guess it's because of reactivity.

inlang/source-code/website/src/interface/sdkDocs/SdkDocsLayout.tsx
```diff
@@ -6,8 +6,8 @@ const SdkDocsLayout = (props: { children: JSXElement }) => {
 	return (
 		<div class="bg-surface-50 min-h-screen">
 			<SdkDocsHeader />
-			<div class="w-full px-4 min-h-[calc(100vh_-_107px_-_480px)]">
-				<div class="max-w-5xl mx-auto">{props.children}</div>
+			<div class="w-full px-4 md:px-8 xl:px-4 min-h-[calc(100vh_-_107px_-_480px)]">
+				<div class="max-w-7xl mx-auto">{props.children}</div>
 			</div>
 			<Footer />
```
flornkm, line 12: Maybe we can change the categories in the footer when visiting the SDK docs to the ones on the tab (SDK and plugins right now), if that's a quick change. Would feel more like a separated page then imo.

inlang/documentation/sdk/tableOfContents.json
flornkm, line 1: tableOfContents looks good to me. Only thing is that the subpath `/plugin` and not free as of the plugin subpage in the docs. Other contributors might not know that and inserting something with the `/plugin` path here, which then redirects them to the other "tab". But that's a rare case, I would say it is good like it is.

flornkm (changes requested):
Just some minor things / questions. @NilsJacobsen super nice work! I like the UI a lot and that we can improve it over time.

inlang/source-code/website/src/pages/documentation/index.page.tsx
NilsJacobsen: fixed ✅

inlang/source-code/website/src/pages/documentation/index.page.tsx
NilsJacobsen: Should be fixed too. We can try that only when the files exist on main.

inlang/source-code/website/src/interface/sdkDocs/SdkDocsLayout.tsx
NilsJacobsen: Replaced Category with Documentation

inlang/documentation/sdk/tableOfContents.json
NilsJacobsen: I don't think that there are huge amounts of people contributing to our docs with new pages. So agree: rare. That was the easiest way I can think of but let's see if we need to iterate.

flornkm (approved):
Looks good to me!

PR #1828, opened 2023-12-05
[skip preview] Implementation of `doc-embed` web component

flornkm:
This isn't working properly yet. It only works with URLs and can't load the accurate CSS.

Additionally, fetching relative paths seems impossible as the web component (sitting in another corner in the repo) doesn't know from where it should begin reading the paths.


samuelstroschein:
@flornkm close this PR as stale?

PR #1829, opened 2023-12-05
Fix/multidomain bug in isogit and cors proxy middleware [fixes #1812]

### LINKED ISSUE #1812: isomorphic git middleware does not call next() and blocks our cors headers middleware function

janfjohannes:
@flornkm  new try this time with the brute force method ;) please check with your branch if it works and verify on prod after merging and deployment, so we dont have the same heart attack if we notice 2 h later prod is down...


flornkm:
Getting the following errors when running locally in production @janfjohannes 
Weird that it works in development

Screenshot 2023-12-06 at 10 24 28 AM

flornkm (approved):

Issue #1835, opened 2023-12-06
[bug] Inlang not generate paraglide folder when project using monorepo


harrytran998:
### Problem

Like the title, I don't know why this does not work but seems this is a bug on Inlang internal.  Hope your team can check this 😆.

Here is the overview.
<img width="177" alt="image" src="https://github.com/inlang/monorepo/assets/24296018/136918c4-02b4-4d64-8446-72eb2c11caa1">

### Expected behavior

Everything works like what you do 😌.

### Reproduction

Clone this repo --> Do like the readme
https://github.com/harrytran998/repro-inlang-paraglide-monorepo

### Other information

MacOs 13 - Any browser

LorisSigrist:
Will take a look!

LorisSigrist:
@samuelstroschein 
Should we log a warning if no messages are found?

samuelstroschein:
@LorisSigrist sounds good!

PR #1838, opened 2023-12-06
Paraglide Bundle Plugins - Warn when no messages are found

LorisSigrist:
The Paraglide bundler plugins skip compilation when no messages are found. This is because something is most likely wrong & it's not a great idea to compile in that case. However, this can cause confusion like in #1835. 

This PR logs a warning whenever compilation get's skipped because no messages were found.


LorisSigrist:
Sorry wrong branch

PR #1849, opened 2023-12-08
[Paraglide] Check if files would change before writing

LorisSigrist:
This PR adds a check to see if the expected files already exist before writing to the output directory.

It currently does not include an automated test for if the skipping works, because of [streamich/memfs#967](https://github.com/streamich/memfs/issues/967), but manual testing / logging says it works!

### DIFF
--- a/inlang/source-code/paraglide/paraglide-js/src/cli/commands/compile.test.ts
+++ b/inlang/source-code/paraglide/paraglide-js/src/cli/commands/compile.test.ts
@@ -73,34 +73,35 @@ test("it should compile into the default outdir", async () => {
 	expect(_fs.existsSync("./src/paraglide/messages.js")).toBe(true)
 })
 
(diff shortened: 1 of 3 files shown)


samuelstroschein:
I propose to change the approach to contain state in the file writing logic that stores hashes of previous files and don't rewrite if the hashes are identical. 

**Why**

The problem is about watching changes. A watcher is a long-living process that can contain runtime state.  

**Pros**

- more robust because diffing of objects (seems flaky!) is not required 
- faster because no files need to be read before writing the files

**Pseudocode**

```
1. (initial) write file stores hashes for each file
2. (subsequent) write file compares previous hashes with new files. if the hashes differ, write to disk
```

The problem is only about HMR

LorisSigrist:
That's a sensible approach. I'll reimplement it like that!

samuelstroschein (approved):

PR #1862, opened 2023-12-11
Add "Valid JS identifier message id" lint rule

### LINKED ISSUE #1855: [bug] reserved keywords
### Problem

There's is nothing stopping me from adding keys with reserved keywords:

```
{
   "delete": "Delete me"
}
```

compiles into:
```
export const delete = (params = {}, options = {}) => {
	const tag = options.languageTag ?? languageTag();
	if (tag === "en") return en.delete()
	return undefined
}
```

Which is not allowed.

`> npx inlang lint` no errors / warning is thrown, nor when running `> npx inlang machine translate ...`

Will only fail when client tries to compile.

### Expected behavior

Warnings / errors should be thrown if you supply reserved keys in .json files.

### Reproduction

Just add any of https://www.w3schools.com/js/js_reserved.asp in a `en.json` file.

### Other information

_No response_

LorisSigrist:
Closes #1855 

This PR adds a lint rule to enforce that all messageIDs are valid Javascript Identifiers that can be used as variable names. It fails if:
- The MessageID isn't a valid Javascript Identifier
- The MessageID is a reserved word

We should consider having this be part of the default setup with Paraglide


samuelstroschein (approved):
Yes, let's add this lint rule to `paraglide init` and put this lint rule into the docs of paraglide "recommended lint rule". Agree?

LorisSigrist:
The CI error comes from the fact that the lint-rule isn't available on npm yet. But of course it won't be on npm until we merge this PR. Bit of a chicken and egg problem. I say we send it and just merge. Everything seems to work in local testing. Also, this PR won't trigger a paraglide release, so we could revert if there is an issue

LorisSigrist:
Actually, nevermind, the publishing Action won't run until CI passes. I'll need to revert the changes to paraglide & add them separately

Issue #195, opened 2022-12-16
conditionally render login button to prevent flickering on intial render


samuelstroschein:
## Problem 

The login button loads slot loads eagerly which renders a lot of text in the nav bar on the initial render which leads to flickering.

## Proposal 

Use Solid's conditional render to prevent flickering.

samuelstroschein:
closed with #196

PR #1877, opened 2023-12-12
Fix internal #195

jannesblobel:
This PR fixes the internal bug

https://github.com/inlang/internal/issues/195


LorisSigrist:
Could you remove paraglide from the changeset and instead manually do a version bump & changelog update? Chaneget currently doesn't support the `1.0.0-prerelease.x` naming scheme we're using

LorisSigrist (approved):

samuelstroschein (approved):

PR #1880, opened 2023-12-12
Fixes in fs.watch and in getPrivateEnvVariables to allow browserAuth run in env with process defined

martin-lysk:
fixes watching on root folder, fixes watcher when multiple listener are attached, fixes private env in env where process is defined but


lix/source-code/fs/src/implementations/memoryFs.ts
```diff
@@ -183,24 +183,30 @@ export function createNodeishMemoryFs(): NodeishFilesystem {
 			})
 
 			const listener = (event: FileChangeInfo) => {
-				if (event.filename === null) {
+				const eventToForward = {
+					eventType: event.eventType,
+					filename: event.filename,
+				}
+
+				if (eventToForward.filename === null) {
```
janfjohannes, line 191: @martin-lysk  could you rename eventToForward to event and  just destructure the incoming event directly? into eventType, filename ertc?

janfjohannes (changes requested):

janfjohannes:
@martin-lysk  also needs manual rebase due to conflicts

lix/source-code/client/src/helpers.ts
```diff
@@ -18,7 +18,8 @@ export const withLazyFetching = ({
 }: Args): NodeishFilesystem => {
 	return new Proxy(nodeishFs, {
 		get(getTarget: typeof nodeishFs, prop, receiver) {
-			if (getTarget[prop as keyof typeof nodeishFs]) {
+			// eslint-disable-next-line no-prototype-builtins
+			if (((getTarget as any).__proto__ as any).hasOwnProperty(prop)) {
```
janfjohannes, line 22: not sure i want this in general, why is this needed except for your testing?

janfjohannes (changes requested):

janfjohannes (changes requested):
sorry i forgot to submit this time. could you have a look at the protoype handling? i would not want to have to think if something is part of a prototype chain or a high level feature . i would expect i can replace any function call with the proxy.

lix/source-code/client/src/helpers.ts
martin-lysk: @janfjohannes This problem occured when you subclass a nodish fs and mark a method as private - instead of the functional approach since private functions become visible to the target. The Proxy should only proxy methods that are defined in the NodishFs type but this is not the case with the current implementation.

janfjohannes: @martin-lysk  not sure what you mean, lets discuss in a call.

samuelstroschein:
What is the status?

janfjohannes:
@samuelstroschein  martin tries to have a look at my comments today.

janfjohannes (approved):

PR #1915, opened 2023-12-18
initial implementation of projectIds [closes  #1891]

### LINKED ISSUE #1891: project ids
## Problem 

We have difficulty improving inlang because projects have no project identifier for telemetry. We want answers to questions like "What plugins are used most often, and we should prioritize?", "What settings should we remove because they are not used?", etc. 

## Proposal

Introduce project identifiers. 

## Requirements

- [x] persistent in the project directory
- [x] no merge conflicts for users 
- [x] two (distributed) users should generate the same project id on first run
- [x] statistically unique

janfjohannes:
and minimal sdk refactor to support the feature


janfjohannes:
@samuelstroschein  unfortunatly i gave up on implementing the actual telemetry.group call as this would probably be redone in the remove of _capture ticket.   i got lost in the different packages for posthog/ node /browser,  the proxy wrapping with types etc.   this would be proboably best finished by someone who understands all the constraints of the posthog package.

inlang/source-code/sdk/src/loadProject.ts
```diff
@@ -80,13 +82,47 @@ export const loadProject = async (args: {
...
+						await nodeishFs.writeFile(projectPath + "/projectId", projectId)
+					}
+				} else {
+					console.error(error)
+				}
+			}
+
+			// TODO:
+			// if (projectId) {
+			// 	telemetryBrowser.group("project", projectId, {
```
janfjohannes, line 115: missing piece, to be implemented when removing _capture

inlang/source-code/sdk/src/loadProject.ts
```diff
@@ -80,13 +82,47 @@ export const loadProject = async (args: {
...
 
 		const [settings, _setSettings] = createSignal<ProjectSettings>()
-		createEffect(() => {
+		createEffect(async () => {
+			let projectId: string | undefined
+
+			try {
+				projectId = await nodeishFs.readFile(projectPath + "/projectId", {
+					encoding: "utf-8",
+				})
```
samuelstroschein, line 91: Why not `/project-id.txt` it's a txt file right?

inlang/source-code/sdk/src/loadProject.ts
```diff
@@ -80,13 +82,47 @@ export const loadProject = async (args: {
...
+							new TextEncoder().encode(`${repoMeta.id + projectPath}`)
+						)
+						projectId = [...new Uint8Array(idDigest)]
+							.map((b) => ("00" + b.toString(16)).slice(-2))
+							.join("")
+
+						await nodeishFs.writeFile(projectPath + "/projectId", projectId)
+					}
+				} else {
+					console.error(error)
```
samuelstroschein, line 109: Why not push to project errors?

inlang/source-code/sdk/src/loadProject.ts
```diff
@@ -80,13 +82,47 @@ export const loadProject = async (args: {
...
+
+					if (repoMeta && !("error" in repoMeta)) {
+						const idDigest = await crypto.subtle.digest(
+							"SHA-256",
+							new TextEncoder().encode(`${repoMeta.id + projectPath}`)
+						)
+						projectId = [...new Uint8Array(idDigest)]
+							.map((b) => ("00" + b.toString(16)).slice(-2))
+							.join("")
+
```
samuelstroschein, line 105: 1. Can we turn this code into a dedicated function? 
2. Can you test this function? 
3. Can you place this function outside of the effect? This function doesn't seem to require an effect. This is an initial load problem that does not seem to need reactivity.

inlang/source-code/editor/src/pages/@host/@owner/@repository/State.tsx
```diff
@@ -265,18 +265,15 @@ export function EditorStateProvider(props: { children: JSXElement }) {
 				const project = solidAdapter(
 					await loadProject({
 						nodeishFs: newRepo.nodeishFs,
+						repo: newRepo,
```
samuelstroschein, line 268: Why is `nodeishFs` still required as a parameter? 

I see substantial risk where the SDK uses a different `fs` than the repo. I don't see a reason why we should not immediately remove `nodeishFs` as a parameter. At least only allow allow either nodesihFs or only repo but not both.

samuelstroschein (changes requested):
Where are the tests? 😱

inlang/source-code/sdk/src/loadProject.ts
janfjohannes: its not a real text file. usually fiels that only have an id dont have an ending, see pid files etc.

inlang/source-code/editor/src/pages/@host/@owner/@repository/State.tsx
janfjohannes: @samuelstroschein  because not everything is using lix and even has a repo object that uses the loadproject, things like ide extension would first need to migrate to lix and remove the direct fs usage before we can deprecate the parameter.

samuelstroschein: @janfjohannes can we migrate now? otherwise, ide extension projects won't have an id. 

the migration should be simple. the @inlang/ide-extension passes a repo object instead of an fs object.

inlang/source-code/sdk/src/generateProjectId.test.ts
```diff
@@ -0,0 +1,24 @@
+import { generateProjectId } from "./generateProjectId.js"
+import { describe, it, expect } from "vitest"
+import { openRepository } from "@lix-js/client"
+import { createNodeishMemoryFs } from "@lix-js/fs"
+
+describe("generateProjectId", async () => {
+	const repo = await openRepository("https://github.com/inlang/ci-test-repo", {
+		nodeishFs: createNodeishMemoryFs(),
+	})
```
samuelstroschein, line 9: @janfjohannes just to double check: 

are we sure that we don't want to introduce `createVirtualRepo` with this change? See https://github.com/inlang/monorepo/issues/1898#issuecomment-1861852281, we can "ignore" github APIs. 

- the test is long running now because it makes a network request
- the test is fragile because it makes a network request

samuelstroschein:
```ts

const repo = await createVirtualRepo({
   // optional arguments
  origin: "https://mock.com"
})

samuelstroschein (approved):

samuelstroschein:
@janfjohannes in light of me being OOO tmr, I approve this PR but would like to see `createVirtualRepo()` https://github.com/inlang/monorepo/pull/1915#discussion_r1432003004. Ultimately your choice whether you implement it (without GitHub APIs) or merge this PR without.

janfjohannes:
@samuelstroschein unfortunately i dont get the tests to pass in github ci anyways and have to get back fresh to this, im totally "durch" for today... have a great flight!

samuelstroschein:
@janfjohannes man this multi platform development is such a pain. have a good night!

jannesblobel (approved):

janfjohannes:
@samuelstroschein  not merging this as i had to move the id generation back into the effect and i have no idea why:
on node 18  there are plenty of fialing tests  "crypto undefined" only if project id generation is not inside the effect , node 20 works.  feel free to merge if we can leave it in the effect.

janfjohannes:
aaaaaaargh another nx cache  issue!! node 16 tests are not run at all and the tests from node 18 are reused! actually the same for the platforms, thats why tests were looking faster: they just did not run !

inlang/source-code/sdk/src/generateProjectId.test.ts
janfjohannes: @samuelstroschein  replaced the network requests with mock data

PR #1923, opened 2023-12-19
Fix error in language switcher example

tomas-correia:
The "withoutLanguageTag" function in the language switcher example provided had a mistake in the first return statement. The mistake was a missing '/' at the beginning of the returned string, which caused wrong paths to be returned.


LorisSigrist (approved):

LorisSigrist:
Thanks! I'll merge this tomorrow

PR #1928, opened 2023-12-20
make cache depend on node ver and platform

janfjohannes:


janfjohannes:
as far as i undestand we reuse tests between platforms and  node versions...

LorisSigrist (approved):
Good change! This might explain the occasional flakey test

PR #1948, opened 2023-12-27
[Paraglide] improve message Index Code

LorisSigrist:
When doing the benchmarks it became obvious that paraglide has a massive weakness. All the if-statements in the message-index minify extremely poorly. This was just as big a drawback as shipping the messages in multiple languages.

This PR replaces the massive list of `if` statements with an indexing operation into a Record.

```js
//before (minified)
const a = (p,o) => {
  const L = o.languageTag ?? l();
  if(L == "de") return a(p)
  if(L == "en") return b(p)
  if(L == "en-US") return c(p)
  return undefined;
}
```

```js
//after (minified)
const a = (p,o) => {
  const m = {
    de:a,
    en:b,
    "en-US":c
  }[o.languageTag ?? l()]
  return m ? m(p) : undefined;
}
```

This scales much better than the `if` chain
Once there is some fallback behaviour for the languages it will be possible to simplify this even further.

```js
//future w/fallback (minified)
const a = (p,o) => {
  return {
    de:a,
    en:b,
    "en-US":c
  }[o.languageTag ?? l()](p)
}
```


samuelstroschein:
Interesting. I would have expected the bundler to minify the if statements automatically. @LorisSigrist wouldn't it be good to get rid of the if statements altogether in the compiler in favor of directly emitting the record? 

PS I vote to return the message id as fallback. It makes searching for missing messages easier (but ofc increases the bundle size)

LorisSigrist:
I would have expected that too, but it's just not the case. 

This PR already emits the record directly from the compiler. That's why there are no more if-else statements. 

I'll switch it so that it returns the messageID as a fallback.

PR #1952, opened 2023-12-28
Paraglide: Add Language Fallbacks

### LINKED ISSUE #1949: [Paraglide] Language Fallbacks
Continuation from: #1728

## Problem 
Currently, when the current language doesn't have a given message `undefined` is returned. 

This is an issue for a few reasons:
- `undefined` is not helpful to users
- The message files (`./paraglide/messages/{lang}.js`) don't share a single interface (-> can't be substituted for one another)

You _can_ catch missing messages using the lint rule, but paraglide should work independently of which lint rules are configured. 

## Proposal
Define a default fallback order & substitute messages from other languages if they are missing in the current language. 

I would propose the following order:
1. Exact match: `de-CH-ZH`
2. Regional in descending specificity: `de-CH` then `de`
3. Source Language: `en`

This should guarantee that there is always some message that can be shown to users.
For the moment I wouldn't make the fallback order configurable, we can always add that in the future.

LorisSigrist:
Closes #1949 

With this PR paraglide now attempts to use messages from fallback languages if they are missing in the current language. 

```ts
//messages/de.js
export { missing_in_german } from "./en.js"
```

If no fallback language can be determined and a message is missing, it will return the message ID. 

Because _all_ languages now return something for _all_ messages, I was able to further compact the message index code.


samuelstroschein (changes requested):
The lookup implementation does not seem spec compliant. [Section 3.4.1](https://datatracker.ietf.org/doc/html/rfc4647#section-3.4.1) is missing. 

> Each application, protocol, or specification that uses lookup MUST
   define the defaulting behavior when no tag matches the language
   priority list.

The correct lookup function type seems to be:

```ts
export function getLookupOrder(
	availableLanguageTags: LanguageTags,
	languageTag: LanguageTag
        default: LanguageTag
)
```

- source language tag is an inlang concept that is not defined in the RFC
- I would also rename the function to `lookup()` or `matchByLookup()` to align with the name of the RFC "matching language tags [by strategy]".

inlang/source-code/versioned-interfaces/language-tag/src/lookup.ts
```diff
@@ -0,0 +1,30 @@
...
+		if (languageTagParts[i - 1] === "x") continue
+
+		//Stringify the language tag parts
+		const fallbackLanguageTag = languageTagParts.slice(0, i).join("-")
+		if (!availableLanguageTags.includes(fallbackLanguageTag)) continue
+		fallbackLanguages.push(fallbackLanguageTag)
+	}
+
+	return fallbackLanguages[0] ?? defaultLanguageTag
+}
```
samuelstroschein, line 30: Should the function throw if the `defaultLanguageTag` is not included in `availableLanguageTags` according to the spec?

inlang/source-code/versioned-interfaces/language-tag/src/lookup.test.ts
```diff
@@ -0,0 +1,44 @@
...
+	expect(lookup("zh-Hans-CN", ["en"], "en")).toEqual("en")
+})
+
+it("does not get confused by the x separator", () => {
+	expect(
+		lookup("zh-Hans-CN-x-private1-private2", ["en", "zh-Hans-CN", "zh-Hans", "zh"], "en")
+	).toEqual("zh-Hans-CN")
+	expect(lookup("zh-Hans-CN-x-private1-private2", ["en", "zh-Hans", "zh"], "en")).toEqual("zh-Hans")
+	expect(lookup("zh-Hans-CN-x-private1-private2", ["en", "zh"], "en")).toEqual("zh")
+	expect(lookup("zh-Hans-CN-x-private1-private2", ["en"], "en")).toEqual("en")
```
samuelstroschein, line 43: Use an object as an argument if you have more than two arguments. 

- increases readability 
- possibility to change (deprecate/replace) arguments (HUGE WIN)

```ts
lookup("zh-Hans-CN-x-private1-private2", { availableLanguageTags: ["zh-Hans-CN-x-private1-private2"], default: "en" })
```

samuelstroschein: On that note, we could even remove `availableLanguageTags` in favor of just `languageTags`. 

```ts
lookup("zh-Hans-CN-x-private1-private2", { languageTags: ["zh-Hans-CN-x-private1-private2"], default: "en" })
```

inlang/source-code/versioned-interfaces/language-tag/src/lookup.ts
LorisSigrist: I don't see the spec mentioning that the default must be part of the options. I don't really see a benefit to enforcing that either. It would make consuming the API a little more awkward.

inlang/source-code/versioned-interfaces/language-tag/src/lookup.test.ts
LorisSigrist: Good point! The options-object looks much nicer.
I vote for just calling the argument `languageTags`, as that's more inline with the spec's language.

PR #1966, opened 2024-01-03
Add `pnpm audit` to CI

LorisSigrist:
This PR adds a new CI action that audits our dependencies.  If there is a vulnerability of severity `high` or above it will error.


samuelstroschein:
@LorisSigrist I would put this check in the `ci` script https://github.com/opral/monorepo/blob/main/package.json#L26

- faster feedback loop that works locally
- no dependency github actions

janfjohannes:
@samuelstroschein  @felixhaeberle @LorisSigrist    i removed an unused vulnerable dependency from the markdown module and ovverrode the vscode test dependency.  tests seem to pass now.

janfjohannes (approved):

PR #1967, opened 2024-01-04
Fix project ids

janfjohannes:


janfjohannes:
- added support for local repo opening in lix
- fixed optional use of repo for project id generation in loadProject of sdk
- added proejct id generation by using repo in badge and cli

inlang/source-code/sdk/src/loadProject.test.ts
```diff
@@ -197,6 +197,42 @@ describe("initialization", () => {
 		expect(result.data).toBeDefined()
 	})
 
+	it("should generate projectId on missing projectid in local repo", async () => {
+		const repo = await mockRepo({ openLocal: true })
```
samuelstroschein, line 201: I would NOT expect an argument to `mockRepo`. @janfjohannes Please change the API to

```ts
const repo = await virtualRepository()

const repo = await virtualRepository({ fromSnapshot: "" })

const repo = await virtualRepository({ fromFilesystem: ""})
```

inlang/source-code/sdk/src/loadProject.ts
```diff
@@ -45,15 +45,24 @@ const settingsCompiler = TypeCompiler.Compile(ProjectSettings)
 export const loadProject = async (args: {
 	projectPath: string
 	repo?: Repository
-	nodeishFs: NodeishFilesystem
+	nodeishFs?: Repository["nodeishFs"]
```
samuelstroschein, line 48: Correct the type to EITHER repo OR nodeisfh (deprecated)

samuelstroschein (changes requested):

inlang/source-code/cli/src/utilities/getInlangProject.ts
```diff
@@ -10,17 +11,43 @@ export async function getInlangProject(args: { projectPath: string }): Promise<I
...
+			nodeishFs: fs,
+			_capture(id, props) {
+				telemetry.capture({
+					// @ts-ignore the event types
+					event: id,
+					properties: props,
+				})
+			},
+		})
+	} else {
```
samuelstroschein, line 33: I vote for just throwing. 

We will complicate our code and architecture if we bake conditional logic everywhere, whether a repo exists or not. Inlang requires lix and lix requires repositories.

samuelstroschein: What do you think?

samuelstroschein: The throw can happen in `openRepository` a la "no repository path provided. If you are using `findRoot` no root has been found."

janfjohannes: @samuelstroschein  i could throw in the cli and show the error: you need to use the cli inside a git repository, effectively removing support for other version control systems and also using the cli without having a repo. i figured we better do that step when we remove the deprecated nodishFS argument.

samuelstroschein (approved):
I hope I didn't over see something. My last approvals were not top notch

janfjohannes:
@samuelstroschein  you mean our hickup yesterday or something else?

samuelstroschein:
@janfjohannes hickup re mockRepo (which should have been generic)

janfjohannes:
@samuelstroschein  @felixhaeberle  the ide extension build is stuck forever, cancel and retrigger did not solve, did we have this before?

janfjohannes:
@samuelstroschein  @felixhaeberle  the ide extension build is stuck forever and even cancel/ restart does not fix this. did we have this before? any ideas?

samuelstroschein:
@janfjohannes no ideas why this is happening

samuelstroschein:
call notes with @janfjohannes: 

1. remove repo id generation from `openRepository`
2. use conditional in `loadProject` that check if a project id exists
2.1. if no ids exist, use `repo` param in `loadProject` to get repo history for first commit and generate a project id
2.2. else skip

**Pros**

- no repo id discussions now
- we can persist the (project) id -> no performance because history does not need to be fetched if project id already exist 

cc @martin-lysk

inlang/source-code/sdk/src/generateProjectId.ts
```diff
@@ -5,10 +5,11 @@ export async function generateProjectId(repo: Repository, projectPath: string) {
 	if (!repo || !projectPath) {
 		return undefined
 	}
-	const repoMeta = await repo.getMeta()
+	const firstCommitHash = await repo.getFirstCommitHash()
+	const originHash = await repo.getOrigin({ safeHashOnly: true })
```
samuelstroschein, line 9: do we need the origin hash? 

while safe has only should be "safe". do we need to take the risk? Just `firstCommitHash+projectPath` seems sufficient, no?

samuelstroschein: If we don't need it, we can remove the `safeHashOnly` logic in lix too (reduce maintenance)

inlang/source-code/sdk/src/loadProject.ts
```diff
@@ -36,24 +36,49 @@ const settingsCompiler = TypeCompiler.Compile(ProjectSettings)
...
+	} else if (args.repo) {
+		fs = args.repo.nodeishFs
+	} else {
+		throw new LoadProjectInvalidArgument(`Repo missing from arguments.`, { argument: "repo" })
+	}
+
 	// -- migrate if outdated ------------------------------------------------
 
-	await maybeMigrateToDirectory({ nodeishFs: args.nodeishFs, projectPath })
+	await maybeMigrateToDirectory({ nodeishFs: fs, projectPath })
```
samuelstroschein, line 81: This place is a good opportunity to increase the readability by placing `await maybeCreateProjectId()` right beneath `maybeMigrateToDirectory`.

samuelstroschein: the following code (from below) could be encapsulated in a function

```ts

		let projectId: string | undefined

		try {
			projectId = await nodeishFs.readFile(projectPath + "/project_id", {
				encoding: "utf-8",
			})
		} catch (error) {
			// @ts-ignore
			if (error.code === "ENOENT") {
				if (args.repo) {
					projectId = await generateProjectId(args.repo, projectPath)
					if (projectId) {
						await nodeishFs.writeFile(projectPath + "/project_id", projectId)
					}
				}
			} else {
				idError = error as Error
			}
		}


```

samuelstroschein (approved):

samuelstroschein (approved):
i'd like to see the comments resolved BUT can be merged

inlang/source-code/sdk/src/generateProjectId.ts
janfjohannes: @samuelstroschein 

@martin-lysk voted for using this as a salt in the project id to prevent being able to check if a repo specific repo was used by someone, i think its a good idea

samuelstroschein:
@janfjohannes oh :D you have added no changeset. aka this won't be released https://github.com/opral/monorepo/pull/1967#issuecomment-1877041876. I'll implement the telemetry in the SDK and write a changeset for this which will also push this change to depoyment

Issue #1970, opened 2024-01-04
[bug] searching for parrot yields no results


samuelstroschein:
### Problem

<img width="1122" alt="CleanShot 2024-01-04 at 12 32 20@2x" src="https://github.com/opral/monorepo/assets/35429197/42699b1b-a5cf-4e54-bd59-e8d37dc7f887">

### Expected behavior

show parrot

### Reproduction

1. go to inlang.com 
2. search for "parrot"

flornkm:
Parrot doesn't have the keyword, this is not a bug. It just wasn't added, I will ping Martin as soon as he's back from vacation.

<img width="567" alt="Screenshot 2024-01-04 at 9 03 58 PM" src="https://github.com/opral/monorepo/assets/92092993/6004841c-b1df-4298-be6c-ff245fe0258e">

samuelstroschein:
@flornkm why does it need a keyword? Parrot has a title and description. I expect to find parrot if i search for "parrot"

flornkm:
@samuelstroschein didn't we already discuss this? I remember that we decided to have keywords as only rule for how something is being listed so the publishers have all the power.

If we would derive that from the title which I remember we already had one time, Parrot would be listed under the plugins category too (but it is only an app, not an inlang plugin :D) or we would need two Algolia instances:

<img width="326" alt="Screenshot 2024-01-04 at 10 26 24 PM" src="https://github.com/opral/monorepo/assets/92092993/b9f6549b-055e-43a0-ad84-9a42142b5a24">

flornkm:
I will look into how I can get this to run, it is a RPC function so I can derive the info if it should search after title + description from the query. Will look into it tomorrow and try to fix that in an uncomplicated way

samuelstroschein:
@flornkm i mean algolia has a category feature https://www.algolia.com/doc/guides/solutions/ecommerce/browse/tutorials/category-pages/

search should not exclude title and description but ofc filter if i am in a particular category

PR #1971, opened 2024-01-05
Algolia search fix

flornkm:
Reducing costs, getting rid of RPC and making it possible to search for displayName + descriptions.


samuelstroschein:
Does this fix work with alglia category pages or is an interim hack? https://github.com/opral/monorepo/issues/1970#issuecomment-1877795624

flornkm:
@samuelstroschein it doesn't work as the searchFilters have to be set **before** every request and they seem to not be compatible as async functions because it just works on every second search, not on the first one. I couldn't get it running yet. 

I think I have to remove the RPC and include algolia directly on the website so it is pre-rendered based on the route, not on the request.

flornkm:
@NilsJacobsen please review it as I got rid of a component on the homepage (an old one and the component was the initial reason we used RPC).

NilsJacobsen (approved):
You are right we do not need it anymore. 👍

inlang/source-code/website/package.json
```diff
@@ -16,10 +16,10 @@
 		"node": ">=18.0.0"
 	},
 	"dependencies": {
-		"@lix-js/server": "workspace:*",
 		"@isomorphic-git/cors-proxy": "^2.7.1",
 		"@lix-js/client": "workspace:*",
 		"@lix-js/fs": "workspace:*",
+		"@lix-js/server": "workspace:*",
```
samuelstroschein, line 22: @flornkm this can't be. The website can't depend on the server when the server depends on the website.

samuelstroschein: My mistake this is the lix server.

samuelstroschein: @opral/lix well well this is even not a change. the dependency just changed from line 19 to 22. good example that semantic diffing is also desired for code

PR #1989, opened 2024-01-09
Update all old monorepo links to Opral ones

flornkm:
This PR just tries to update all old links.


samuelstroschein:
@flornkm merge? 

and why am "I" deploying the preview envs. this is so annoying

flornkm:
@samuelstroschein will do it tomorrow right away so in case it breaks something I can fix it

PR #1990, opened 2024-01-09
Update astro example

LorisSigrist:
This PR slims down the Astro example & uses Paraglide more prominently. This should help us pass inspection for being added to the Astro docs.

It also removes the blog-section, which is where image-transformations needed to happen. This might fix @felixhaeberle's issues with "sharp".


felixhaeberle:
@LorisSigrist fixed it!

PR #1994, opened 2024-01-09
[fix] Run Paraglide compile during test command in Paraglide Example

LorisSigrist:
This PR also runs `paraglide-js compile` during the `test` script of the paraglide example. This avoids a race condition


janfjohannes:
@LorisSigrist  lol now you ran into the github  bug that happens all day :D

janfjohannes:
i retrigger the tests...

Issue #2013, opened 2024-01-12
sdk tests migration to `loadProject(repo)


samuelstroschein:
## Problem 

The inlang SDK tests are blocking the deprecation of the `nodeishfFs` argument in `loadProject` in favor of `repo` 

For context,project ids yielded the hard requirement for the SDK to directly depend on a repo instance. We (@martin-lysk, @janfjohannes, @samuelstroschein) are confident that future requirements will also have a hard requirement for a repo instance and therefore believe that the inlang SDK depending on lix repos directly is the right decision.  

## Proposal

Replace `nodesihFs` with a repo instance in tests. 

```diff
const project = await loadProject({ 
- nodeishFs, 
+ repo
  projectPath: PATH_TO_INLANG_CONFIG 
})
```

- [ ] refactor tests
- [ ] get rid of conditional logic that handles a potentially undefined repo 
- [ ] remove (now redundant) `loadProject` overloads

LorisSigrist:
#2020 contains the same migration for Paraglide. It might be useful as a reference

jannesblobel:
in loadproject.ts is still a todo
line85 
`// TODO remove tryCatch after https://github.com/opral/monorepo/issues/2013`

Could we remove the TODO? or is this still a todo?

Issue #2012, opened 2024-01-12
paraglide migration to `loadProject(repo)`


samuelstroschein:
## Problem 

Paraglide JS is blocking the deprecation of the `nodeishfFs` argument in `loadProject`. 

_For example_

https://github.com/inlang/monorepo/blob/9eabe589795df8cd8b869685dfb76133bb58e7b2/inlang/source-code/paraglide/paraglide-js-sveltekit-old/src/adapter-sveltekit/vite-plugin/config/virtualModule.ts#L60-L61 

## Proposal

Replace `nodesihFs` with a repo instance. 

```diff
const project = await loadProject({ 
- nodeishFs, 
+ repo
  projectPath: PATH_TO_INLANG_CONFIG 
})
```

`mockRepo()` from lix can be used where no access to a repo exists. cc @janfjohannes who implemented `mockRepo()`

PR #2015, opened 2024-01-12
refactor: some clean up cc @janfjohannes

samuelstroschein:
- added missing types
- changed arguments to `args.*` to align with codebase and increase type readability
- remove redundant `nodeishFs` as argument because exposed via `repo`
- if `repo` is undefined, no project id will be generated (no need for an error). after #2012 and #2013 we can throw if no repo exists
- moved `maybeCreateFirstProjectId` into migrations for less clutter


inlang/source-code/sdk/src/generateProjectId.test.ts
```diff
@@ -1,28 +0,0 @@
...
-
-	it("should return undefined if repoMeta contains error", async () => {
-		const errorRepo = await openRepository("https://github.com/inlang/no-exist", {
-			nodeishFs: createNodeishMemoryFs(),
-		})
-
-		const projectId = await generateProjectId(errorRepo, "mocked_project_path")
-		expect(projectId).toBeUndefined()
-	})
-})
```
janfjohannes, line 28: if you commit a rename first it is possible to track changes across ;)

inlang/source-code/sdk/src/loadProject.ts
```diff
@@ -94,23 +90,25 @@ export async function loadProject(args: any): Promise<InlangProject> {
...
+	if (args.nodeishFs) {
+		// TODO: deprecate
+		fs = args.nodeishFs
+	} else if (args.repo) {
+		fs = args.repo.nodeishFs
+	} else {
+		throw new LoadProjectInvalidArgument(`Repo missing from arguments.`, { argument: "repo" })
+	}
+
 	const nodeishFs = createNodeishFsWithAbsolutePaths({
```
janfjohannes, line 103: maybe add the comment // we narrow the fs down to a NodeishFilesystemSubset because we only need the subset from here

inlang/source-code/sdk/src/loadProject.ts
```diff
@@ -64,19 +64,15 @@ export async function loadProject(args: {
 	_capture?: (id: string, props: Record<string, unknown>) => void
```
janfjohannes, line 64: if you swap the overloads position the right intellisense pops up...

samuelstroschein: removing this in the next step

inlang/source-code/sdk/src/generateProjectId.test.ts
samuelstroschein: great. new requirement for lix. do this automatically pls :D

PR #2016, opened 2024-01-13
SDK add `loadedProject` event [skip preview]

samuelstroschein:
## Changes

- add capture function for telemetry
- add `loaded project` event

## Todo 

- [x] env variable injection to have different posthog keys for dev and production


samuelstroschein:
@jannesblobel this is an example event from our repo. Good? 

CleanShot 2024-01-12 at 20 33 12@2x

jannesblobel:
It looks good atm.
Can you rename the project to project_Id?

samuelstroschein:
@jannesblobel this is a group. the group "project" is identified by the id

jannesblobel:
This "project" is a group? Which values can be stored in it? 
Or what does the number after it stand for?
CleanShot 2024-01-14 at 16 24 34

jannesblobel:
Btw. 
> HogQL allows many insights with this schema, e.g

that's correct, but when you use HogQL, you don't get any charts. That's the problem

samuelstroschein:
@jannesblobel lets have a call tmr

inlang/source-code/sdk/src/loadProject.ts
```diff
@@ -46,6 +46,16 @@ const settingsCompiler = TypeCompiler.Compile(ProjectSettings)
...
+	/**
+	 * The app id is used to identify the app that is using the SDK.
+	 *
+	 * We use the app id to group events in telemetry to answer questions
+	 * like "Which apps causes these errors?" or "Which apps are used more than others?".
+	 *
+	 * @example
+	 * 	appId: "app.inlang.badge"
+	 */
+	appId?: string
```
jannesblobel, line 58: Shouldn't we simply make this mandatory?

samuelstroschein: no. we don't want tests to have an appId

jannesblobel: Don't we have to do it sooner or later anyway because of the marketplace?

samuelstroschein: No it's redundant. Telemetrty is not a hard requirement for the SDK to work. Forcing appids for the SDK would be wrong

samuelstroschein:
@jldec review whether you are OK with the env variable approach I took (https://github.com/opral/monorepo/blob/990a4fe425364ba35c3cd18d9fae16b8dbead9bb/inlang/source-code/sdk/src/env-variables/createIndexFile.js). 

i haven't found a better way than creating a script that is invoked before compiling the project. The alternative would be a bundler that injects env variable into the build output. introducing such a large dependency seemed unnecessary.

jldec:
> @jldec review whether you are OK with the env variable approach I took. If you have "are not in the topic" yet, I can also go ahead and merge this myself.
> 
> See https://github.com/opral/monorepo/blob/990a4fe425364ba35c3cd18d9fae16b8dbead9bb/inlang/source-code/sdk/src/env-variables/createIndexFile.js
> 
> I haven't found a better way than creating a script that is invoked before compiling the project. The alternative would be a bundler that injects env variable into the build output. introducing such a large dependency seemed unnecessary.

Good for now - My quick google for build-time env var substitution [did not yield](https://stackoverflow.com/questions/62682043/transform-env-variables-during-build-time-with-typescript) anything cleaner. I expect that as our builds get fancier we'll find a nicer way to do this e.g. PUBLIC static env vars [from sveltekit](https://kit.svelte.dev/docs/modules#$env-static-public).

jldec (commented):
LGTM

Issue #1898, opened 2023-12-15
Sdk needs a repo  reference instead of only the fs


janfjohannes:
## Problem 
We often run into issues not having lix/repo features in the sdk that lead to implementing features at the wrong abstraction level or hacky testing (cc @martin-lysk ) The current blocker is that predictable project ids should be generated based on the identity of the repository the are in but this info is only known to lix/git, not in the fs directly. 

## Proposal

As discussed with @samuelstroschein  we decided to add a required repo argument to loadProject next to the fs and then phase out the use of fs and deprecate that argument.

samuelstroschein:
@janfjohannes is `createVirtualRepo()` for testing purposes (replaces `createMemoryFs()`) in the scope of the issue? 

```ts
import { createVirtualRepo } from "@lix-js/client"

const repo = await createVirtualRepo()
const project = await loadProject({ repository: repo, projectPath: "blalba.inlang" }) 

```

martin-lysk:
Great we need a repo in the sdk for sure. Not for persistence i hope - but for a lot of other things like history queries, change detection, review logs and many other things. I guess there are cases where the project need to exist before the repo - like in parrot where project may start in figma but the connection to git comes at a later time.

felixhaeberle:
Off-topic but I love how the SDK is getting more love since a few days 😍

janfjohannes:
@samuelstroschein  createVirtulalRepo would require mocks for all github apis that are used which would be quite an undertaking and probably better done when we add support for other git providers and need to do this anyways.

 i recomend closing this ticket and i can of course provide mock features for any repo functions that a test you want reqires, if you want more tests now, please give a rough description for me to add them.

samuelstroschein:
> createVirtulalRepo would require mocks for all github apis that are used which would be quite an undertaking 

Would throwing if external APIs are used be an option (via a proxy for example)? 

Seems logical that SDK devs want to pass a mock repo object to `loadProject` instead of: 

1. Create a virtual filesystem
2. open a repo with the virtual filesystem
3. load a project

felixhaeberle:
> Seems logical that SDK devs want to pass a mock repo object to loadProject

Really important ❗️(even for ourselves to have a nice way to test functionality with loadProject)

janfjohannes:
@here i added ar repo = mockRepo() function to lix/client

PR #2020, opened 2024-01-15
No longer use `nodeisFs` in `loadProject`.

### LINKED ISSUE #2012: paraglide migration to `loadProject(repo)`

LorisSigrist:
Closes #2012

### DIFF
--- a/inlang/source-code/paraglide/paraglide-js-adapter-unplugin/src/index.ts
+++ b/inlang/source-code/paraglide/paraglide-js-adapter-unplugin/src/index.ts
@@ -1,5 +1,6 @@
 import { createUnplugin } from "unplugin"
 import { Message, ProjectSettings, loadProject, type InlangProject } from "@inlang/sdk"
+import { openRepository } from "@lix-js/client"
 import path from "node:path"
 import fs from "node:fs/promises"
 import { compile, writeOutput, Logger } from "@inlang/paraglide-js/internal"
@@ -59,9 +60,14 @@ export const paraglide = createUnplugin((config: UserConfig) => {
 	let project: InlangProject | undefined = undefined
 	async function getProject(): Promise<InlangProject> {
 		if (project) return project
+
+		const repo = await openRepository("file://", {
+			nodeishFs: fs,
+		})
+
 		project = await loadProject({
 			projectPath: path.resolve(process.cwd(), options.project),
-			nodeishFs: fs,
+			repo,
 		})
 		return project
 	}
--- a/inlang/source-code/paraglide/paraglide-js-sveltekit-old/src/adapter-sveltekit/vite-plugin/config/virtualModule.ts
+++ b/inlang/source-code/paraglide/paraglide-js-sveltekit-old/src/adapter-sveltekit/vite-plugin/config/virtualModule.ts
@@ -14,6 +14,7 @@ import { shouldContentBePrerendered } from "./utils/shouldContentBePrerendered.j
 import { getSettings } from "./utils/getSettings.js"
 import { doesPathExist } from "./utils/utils.js"
 import { getNodeishFs } from "./utils/getNodeishFs.js"
+import { openRepository } from "@lix-js/client"
 
 type VersionString = `${number}.${number}.${number}${string}`
 
@@ -57,7 +58,10 @@ export const initVirtualModule = async (): Promise<VirtualModule> => {
 
 	// eslint-disable-next-line no-async-promise-executor
 	return (VirtualModule = new Promise<VirtualModule>(async (resolve, reject) => {
-		const inlang = await loadProject({ nodeishFs, projectPath: PATH_TO_INLANG_CONFIG })
+		const repo = await openRepository("file://", {
+			nodeishFs,
+		})
+		const inlang = await loadProject({ repo, projectPath: PATH_TO_INLANG_CONFIG })
 
 		const errors = inlang.errors()
 		if (errors.length) {
@@ -90,7 +94,7 @@ export const initVirtualModule = async (): Promise<VirtualModule> => {
 			return {
 				default: new InlangSdkException(
 					`Could not find 'svelte.config.js' file (${PATH_TO_SVELTE_CONFIG})`,
-					error as Error,
+					error as Error
 				),
 			}
 		})) as { default: SvelteConfig | InlangSdkException }
@@ -101,15 +105,15 @@ export const initVirtualModule = async (): Promise<VirtualModule> => {
 		const files = {
 			appTemplate: path.resolve(
 				PATH_TO_CWD,
-				svelteConfig.kit?.files?.appTemplate || path.resolve("src", "app.html"),
+				svelteConfig.kit?.files?.appTemplate || path.resolve("src", "app.html")
 			),
 			routes: path.resolve(
 				PATH_TO_CWD,
-				svelteConfig.kit?.files?.routes || path.resolve("src", "routes"),
+				svelteConfig.kit?.files?.routes || path.resolve("src", "routes")
 			),
 			serverHooks: path.resolve(
 				PATH_TO_CWD,
-				svelteConfig.kit?.files?.hooks?.server || path.resolve("src", "hooks.server"),
+				svelteConfig.kit?.files?.hooks?.server || path.resolve("src", "hooks.server")
 			),
 		}
 
@@ -123,7 +127,7 @@ export const initVirtualModule = async (): Promise<VirtualModule> => {
 
 		const usesTypeScript = await doesPathExist(
 			nodeishFs,
-			path.resolve(PATH_TO_CWD, "tsconfig.json"),
+			path.resolve(PATH_TO_CWD, "tsconfig.json")
 		)
 
 		const svelteKitVersion = await getSvelteKitVersion()
--- a/inlang/source-code/paraglide/paraglide-js/src/cli/commands/compile.ts
+++ b/inlang/source-code/paraglide/paraglide-js/src/cli/commands/compile.ts
@@ -42,10 +42,14 @@ export const compileCommand = new Command()
 			logger.warn(`Could not find repository root for path ${path}`)
 			// We still support projects without git repo for now.
 
+			const repo = await openRepository("file://", {
+				nodeishFs: fs,
+			})
+
 			project = exitIfErrors(
 				await loadProject({
 					projectPath: path,
-					nodeishFs: fs,
+					repo,
 					_capture(id, props) {
 						telemetry.capture({
 							// @ts-ignore the event types
--- a/inlang/source-code/paraglide/paraglide-js/src/cli/commands/init.ts
+++ b/inlang/source-code/paraglide/paraglide-js/src/cli/commands/init.ts
@@ -11,6 +11,7 @@ import { version } from "../state.js"
 import { telemetry } from "../../services/telemetry/implementation.js"
 import { Logger } from "../../services/logger/index.js"
 import dedent from "dedent"
+import { openRepository } from "@lix-js/client"
 
 // TODO add a project UUID to the tele.groups internal #196
 // import { gitOrigin } from "../../services/telemetry/implementation.js"
@@ -84,10 +85,13 @@ export const maybeAddVsCodeExtension = async (args: { projectPath: string }, log
 		return
 	}
 
+	const repo = await openRepository("file://", {
+		nodeishFs: fs,
+	})
+
 	const project = await loadProject({
 		projectPath: resolve(process.cwd(), args.projectPath),
-		//@ts-ignore
-		nodeishFs: fs,
+		repo,
 	})
 
 	const settings = project.settings()
@@ -159,11 +163,16 @@ export const existingProjectFlow = async (
 	if (selection === "newProject") {
 		return createNewProjectFlow(logger)
 	}
-	const project = await loadProject({
-		projectPath: resolve(process.cwd(), args.existingProjectPath),
-		//@ts-ignore
+
+	const repo = await openRepository("file://", {
 		nodeishFs: fs,
 	})
+
+	const project = await loadProject({
+		projectPath: resolve(process.cwd(), args.existingProjectPath),
+		repo,
+	})
+
 	if (project.errors().length > 0) {
 		logger.error("The project contains errors: ")
 		for (const error of project.errors()) {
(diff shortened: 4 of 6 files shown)


inlang/source-code/paraglide/paraglide-js-adapter-unplugin/src/index.ts
```diff
@@ -59,9 +60,14 @@ export const paraglide = createUnplugin((config: UserConfig) => {
 	let project: InlangProject | undefined = undefined
 	async function getProject(): Promise<InlangProject> {
 		if (project) return project
+
+		const repo = await openRepository("file://", {
```
janfjohannes, line 64: @LorisSigrist  this will try to open the whole hard drive as a repo and fail. did you try this or am i missing something?

janfjohannes: take a look at uses  findRepoRoot eg.    await findRepoRoot({ nodeishFs: fs, path: projectPath })

janfjohannes:
@LorisSigrist  ahh if you dont use any repo features it will work, but you will not get project_ids, to see if it works please also check that project_ids are generated if you delete the previous porjectfolder/inlang_id

inlang/source-code/paraglide/paraglide-js-adapter-unplugin/src/index.ts
```diff
@@ -59,9 +61,18 @@ export const paraglide = createUnplugin((config: UserConfig) => {
...
 	async function getProject(): Promise<InlangProject> {
 		if (project) return project
+
+		const repoRoot = await findRepoRoot({ nodeishFs: fs, path: projectPath })
+
+		const repo = repoRoot
+			? await openRepository(repoRoot, {
+					nodeishFs: fs,
+			  })
+			: await openRepository("file://", { nodeishFs: fs })
```
janfjohannes, line 71: not sure we should open the root, i would either fail without repo or try opening the cwd...

inlang/source-code/paraglide/paraglide-js-sveltekit-old/src/adapter-sveltekit/vite-plugin/config/virtualModule.ts
```diff
@@ -57,7 +58,10 @@ export const initVirtualModule = async (): Promise<VirtualModule> => {
 
 	// eslint-disable-next-line no-async-promise-executor
 	return (VirtualModule = new Promise<VirtualModule>(async (resolve, reject) => {
-		const inlang = await loadProject({ nodeishFs, projectPath: PATH_TO_INLANG_CONFIG })
+		const repo = await openRepository("file://", {
```
janfjohannes, line 61: see above.

inlang/source-code/paraglide/paraglide-js/src/cli/commands/compile.test.ts
```diff
@@ -1,7 +1,7 @@
 import { vi, test, expect, beforeEach } from "vitest"
 import memfs from "memfs"
-import mockedFs from "node:fs/promises"
-import fs from "node:fs"
+import mockedFs from "fs/promises"
```
janfjohannes, line 3: should this not be prefixed with node: ?

inlang/source-code/paraglide/paraglide-js/src/services/file-handling/write-output.ts
```diff
@@ -1,4 +1,4 @@
-import type nodeFsPromises from "node:fs/promises"
+import type nodeFsPromises from "fs/promises"
```
janfjohannes, line 1: see above

inlang/source-code/paraglide/paraglide-js/src/cli/commands/compile.ts
```diff
@@ -35,17 +35,21 @@ export const compileCommand = new Command()
...
 
-		const repoRoot = await findRepoRoot({ nodeishFs: fs, path })
+		const repoRoot = await findRepoRoot({ nodeishFs: nodeFsPromises, path })
 
 		let project: Awaited<ReturnType<typeof loadProject>>
 		if (!repoRoot) {
 			logger.warn(`Could not find repository root for path ${path}`)
 			// We still support projects without git repo for now.
 
+			const repo = await openRepository("file://", {
```
janfjohannes, line 45: see above

janfjohannes (changes requested):
added a few comments, it would also be fine to fail without repo as discussed today and on the week end with @samuelstroschein

inlang/source-code/paraglide/paraglide-js/src/cli/commands/init.ts
```diff
@@ -11,18 +9,41 @@ import { version } from "../state.js"
 import { telemetry } from "../../services/telemetry/implementation.js"
 import { Logger } from "../../services/logger/index.js"
 import dedent from "dedent"
+import { findRepoRoot, openRepository, type Repository } from "@lix-js/client"
+import type { NodeishFilesystem } from "@lix-js/fs"
+import nodeFsPromises from "fs/promises"
```
janfjohannes, line 14: see above

inlang/source-code/paraglide/paraglide-js-adapter-unplugin/src/index.ts
samuelstroschein: Given that we want a repo to exists in the inlang SDK, opening a mockRepo should be the go-to solution no?

inlang/source-code/paraglide/paraglide-js/src/cli/commands/init.ts
```diff
@@ -442,3 +471,36 @@ const prompt: typeof consola.prompt = async (message, options) => {
 	}
 	return response
 }
+
+export async function fileExists(filePath: string, nodeishFs: NodeishFilesystem) {
+	try {
+		await nodeishFs.access(filePath)
```
janfjohannes, line 477: not familar with access. the common pattern to do exeists checks is do a fs.stat and catch on throwing for non existing. access is not available on memfs.

inlang/source-code/paraglide/paraglide-js/src/cli/commands/init.test.ts
```diff
@@ -810,21 +853,20 @@ const mockUserInput = (testUserInput: any[]) => {
 const mockFiles = (files: memfs.NestedDirectoryJSON) => {
 	const _memfs = memfs.createFsFromVolume(memfs.Volume.fromNestedJSON(files))
 	const lixFs = createNodeishMemoryFs()
```
janfjohannes, line 855: not sure we should use another memfs, why dont you test with our own memoryFS?

janfjohannes (changes requested):
and some more questions added

inlang/source-code/paraglide/paraglide-js/src/cli/commands/init.test.ts
samuelstroschein: not all methods are available that are required and/or the memory fs was not able to be initiated via a JSON for testing

janfjohannes: @samuelstroschein  can we document this? if just the json format is missing i can add this to the snapshot as a feature in a few lines.

inlang/source-code/paraglide/paraglide-js-adapter-unplugin/src/index.ts
janfjohannes: @samuelstroschein  this is the implementation not the test, this nearly got me confused too...

inlang/source-code/paraglide/paraglide-js/src/cli/commands/init.test.ts
samuelstroschein: @janfjohannes sure, coordinate with @LorisSigrist who owns this code now

inlang/source-code/paraglide/paraglide-js-adapter-unplugin/src/index.ts
samuelstroschein: @janfjohannes now i am confused. wdym with implementation not the test? The file:// seems wrong. If no repo exists, `loadProject` should be called with `loadProject(mockRepo)`

samuelstroschein: @LorisSigrist what wording do you find better `loadProject(virtualRepo())` or `loadProject(mockRepo())` in cases where no repo exists?

LorisSigrist: If it's meant to be used in code `virtualRepo` seems more appropriate. `mockRepo` sounds like it's just for testing

samuelstroschein: @janfjohannes i vote for `virtualRepo()` instead of `mockRepo()` given that we decided to build a hard dependency on repos in inlang, which requires the use of "virtual repos" where no repo is available. 

What do you think?

inlang/source-code/paraglide/paraglide-js/src/cli/commands/compile.test.ts
LorisSigrist: My editor was being goofy & only recognized the types when they did not have the `node:` prefix

inlang/source-code/paraglide/paraglide-js/src/cli/commands/init.ts
LorisSigrist: I switched it to the pattern you described so that it can work with lix's memfs

inlang/source-code/paraglide/paraglide-js/src/cli/commands/init.test.ts
LorisSigrist: @janfjohannes The JSON initialisation would be greatly appreciated. I tried doing it with `fromSnapshot`, but that doesn't seem to work unless `fsStat` is provided, which I don't feel like hand-authoring.

Apart from that, I believe lixFs should cover all uses in Paraglide, so ditching memfs would be feasable

inlang/source-code/paraglide/paraglide-js-adapter-unplugin/src/index.ts
LorisSigrist: I will add that `virtualRepo()` is not useable in paraglide in it's current form. It doesn't operate on the real file-system, but on a memfs. Calling `project.settings()` will not actually read the settings from a user's `project.inlang/settings.json` and instead error, because the file is not present in the virtual file-system.

For now the best solution if there is no repo is to open the `cwd` as if it were a repo. Git-operations will fail, but filesystem operations do work just fine. Given that we want lint rules to run in CI where a git repo may not be present, basic operations like reading the settings will always need to run without a real repo, so paraglide can rely on that continuing to work.

inlang/source-code/paraglide/paraglide-js/src/cli/commands/compile.test.ts
LorisSigrist: Fixed it

inlang/source-code/paraglide/paraglide-js/src/cli/commands/init.test.ts
janfjohannes: @LorisSigrist  ok i created a ticket for the json format we need for testing , will ping you when available, lets merge this as is

inlang/source-code/paraglide/paraglide-js-adapter-unplugin/src/index.ts
janfjohannes: @LorisSigrist  @samuelstroschein  hmm there is a big confusion about this feature.  mockRepo is ONLY  for testing with a mock repository that provides some conveneince to setup a memfs from an optional snapshot. virtual repo  as described here as a lix feauture to allow more features for opening folders without an existing repo has nothing to do with mockRepo but is a seperate thing. I am not sure it is a good idea or if we need this as many subtle things would work in unexpected ways or are hard to reason about eg. project ids would  randomly change etc.  currently if you open a folder in lix and it does not have a repo it will just fail all the function calls that require a repo (which is currently all except the nodeishFs  functions.) we could argue if its better to explicitly enable this behaviour or if openRepository should fail without a rpo present, but this is another discussion.

samuelstroschein: @janfjohannes Yes, we need to think about a solution here that is _easily communicatable_. We have confusion here that we need to lift. 


My thought line is:

i wanted a virtualRepo for testing and production use cases. I don't see a difference between `mockRepo` and `virtualRepo`. A developer gets a "virtual repository" whether that is in memory or using files form the real fs doesn't seem to make a large difference? 

```ts
// in memory repository
const repo = await virtualRepo()

// for testing (in memory fs)
const repo = await virtualRepo({ fromSnapshot })

// in memory fs (for repo stuff) with real fs (the use case from @LorisSigrist) 
const repo = await virtualRepo({ usingFs: nodeFs })

``` 

we can also go for 

```ts
// tree-shakable imports
import * as virtualRepo from "@lix-js"

const repo = await virtualRepo.fromSnapshot()
const repo = await virtualRepo.inMemory()
const repo = await virtualRepo.withFs()
```

janfjohannes: we should have a sync call. not sure we are on the same page @samuelstroschein

samuelstroschein: agree. setting a reminder for tmr

jldec: This was also discussed in https://github.com/opral/monorepo/issues/1898 

After talking with @janfjohannes, I pushed a commit in PR #2060 to simplify the use of mockRepo() in loadProject. test.

jldec: and opened a discussion in https://github.com/opral/monorepo/discussions/2061

PR #2031, opened 2024-01-16
[Lix fs] Add encoding option to `writeFile` type

LorisSigrist:
This tripped me up when updating the Paraglide tests to use `NodeishFilesystem`. In regular node, you can specify the encoding when writing a file like so: `fs.writeFile("./file.txt", "I'm some content", {  encoding: "utf-8" })`. `utf-8` is the default, so this doesn't actually do anything, but being explicit is nice. 

This PR updates the type-definition of `NodeishFilesystem.writeFile` to also accept this encoding options. Only `utf-8` is allowed, so no behaviour changes are needed.

### DIFF
--- a/lix/source-code/fs/src/NodeishFilesystemApi.ts
+++ b/lix/source-code/fs/src/NodeishFilesystemApi.ts
@@ -11,7 +11,11 @@
 export type NodeishFilesystem = {
 	// we need to expose internal state _State to the snapshotter without adding it to code completions
 	[x: string]: any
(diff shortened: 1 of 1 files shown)


janfjohannes:
@LorisSigrist    we do not handle that argument and it is not implemented, given that it behaves different from the readfile encoding argument i think its better to leave this out.

Screenshot 2024-01-16 at 13 00 57


- the default is utf8 anyways
- if we handover a buffer it is ignored

janfjohannes:
@LorisSigrist  please just re-open the pr if i missed something or you disagree

PR #2053, opened 2024-01-17
 hotfix for project id fork conflicts

janfjohannes:


.changeset/lovely-balloons-reflect.md
```diff
@@ -0,0 +1,8 @@
+---
+"@inlang/sdk": minor
+"@inlang/cli": minor
+"@lix-js/client": minor
+"@lix-js/fs": minor
```
samuelstroschein, line 5: Please remove all changesets and only include `@inlang/sdk`. 

- This changeset can't be right. You only changed the source code in the SDK. 
- Changeset will automatically bumb the versions of package that use `@inlang/sdk`. There is no need to manually specify versions

samuelstroschein (changes requested):

.changeset/lovely-balloons-reflect.md
janfjohannes: @samuelstroschein  i added those to the changeset as they have changes that were missed befere, can we use the merge to trigger a release for all of them at the same time?

samuelstroschein: @janfjohannes roger!

PR #1890, opened 2023-12-14
inlang tab [skip preview]

felixhaeberle:
fixes #1879

PR #2066, opened 2024-01-19
Fix last uses of loadProject(fs) and remove deprecated api

### LINKED ISSUE #2013: sdk tests migration to `loadProject(repo)

jldec:
resolves #2013 

- [x] 1177f06 fix loadProject(fs) call in cli
- [x] a41ac52 remove create-project and ide-extension auto-config  
- [x] 8fc05a5 fix loadProject(fs) call from ide-extension initProject 
- [x] 93409f5 remove loadProject(fs) overload from sdk (breaking change) 
- [x] changesets

Note that a41ac52 and 8fc05a5 will be superseded by #1890 which also gets rid of auto-config in the ide-extension.


jldec:
🤦 of course, removing the overload [breaks create-project](https://github.com/opral/monorepo/actions/runs/7588957682/job/20672583596?pr=2066#step:8:2818)

samuelstroschein (approved):

samuelstroschein (approved):
Looks good!

I changed the changeset versioning from patch to minor as we don't really use patches. 

Simple heuristic: Major if the user needs to pay attention (breaking change, major marketing update). Everything else is minor.

samuelstroschein (changes requested):
The API of a `project.id` should not be potentially `undefined` anymore with this update. A project id should always be generated.

jldec:
@samuelstroschein I'm trying to figure out what changes you are requesting.

loadProject() requires a repo arg now, so it will always attempt to create a project id if it does not exist.

Was your request motivated by the windows vs posix path delimiter issues, or some leftover checks for repo, or do you think there are still cases where the project will come back undefined with this PR?

inlang/source-code/sdk/src/loadProject.ts
```diff
@@ -99,15 +67,7 @@ export async function loadProject(args: {
...
-		// TODO: deprecate
-		fs = args.nodeishFs
-	} else if (args.repo) {
-		fs = args.repo.nodeishFs
-	} else {
-		throw new LoadProjectInvalidArgument(`Repo missing from arguments.`, { argument: "repo" })
-	}
+	const fs = args.repo.nodeishFs
 
 	const nodeishFs = createNodeishFsWithAbsolutePaths({
```
jldec, line 72: @samuelstroschein I think you're asking to remove the tryCatch on L88.

samuelstroschein: nope this one https://github.com/opral/monorepo/issues/2069

samuelstroschein: https://discord.com/channels/@me/1196510316717154314/1198999268531380225

jldec: aah - ok - 🙏 👀

samuelstroschein (approved):
OK to merge. we can follow up with subsequent changes

jldec:
ok - for clarification - there are two separate outstanding issues revealed by changing the projectId api and looking at the resulting failing tests.

1.  generateProjectId still returns undefined if it's not a "real" repo. To fix that in this PR we would also have to add new logic to do something different like generate a temporary projectId.

2. loadProject() calls normalize() before calling maybeCreateFirstProjectId(). This does not work as expected in Windows because the nodeishFs.stat can't find the normalized Windows path and thinks the .inlang directory passed in does not exist.

Issue #1892, opened 2023-12-14
human readable and random ids for messages by default 


samuelstroschein:
## Problem 

1. @inlang/editor needs a message id generation algorithm to add a "create a message" button 
2. devs choosing message ids leads to multiple problems like namespacing, renaming IDs and thereby breaking the relation to translations, or https://github.com/inlang/monorepo/issues/1889

## Proposal

Apps choose message IDs for users that are human-readable and memorizable but have no "meaning" by default.

```
blue_dot_map
car_sky_keyboard
phone_table_chocolate
...
```

**Pros**

- already best practice for large projects because ids should (must!) have no meaning
- better UX/DX because apps don't need to prompt users for ids 
- a wide class of bugs is eliminated like choosing paraglide incompatible ids
- users can still search and memorize messages 
  - "did you change blue_dot_map?"
  - "we have a missing translation for car_sky_keyboard"

**Cons**

- maybe unexpected behavior for devs. I propose to test implement this and wait for user reactions.

```ts
// own function for tree-shakability 
import { generateMessageId } from "@inlang/sdk"

// checks the project for id conflicts
const id = generateMessageId({ project })

project.query.message.create({ id }) 

```

samuelstroschein:
@inlang/editor @inlang/ide-extension @inlang/paraglide-js good idea? The implementation cost is relatively simple

samuelstroschein:
The idea is inspired by what3words.com

samuelstroschein:
Lovely, a lib exists for this https://www.npmjs.com/package/human-id. three words have 15 million possibilities. 15 million possibilities should be enough for even the largest enterprise use cases. adding a forth word increases the possibilities manifold further

NiklasBuchfink:
It just depends on the developer's workflow and whether they like it or not. I only mention this because these could be possible thoughts:
1. I'm building a modal for the user; why isn't everything prefixed with "modal_user_" since autocomplete can help me with that? (human readable ids are good to memorize short-term and for autocompletion)
2. Clean code says good variable names don't need explanation. We link content with these ids and without context inside the id, I don't know what is behind it (... unless I use the vs code extension and as we know the problems begin when we start renaming ids)

Maybe an alias/comment/description/context field will be necessary. Again something to fill in and find names.

samuelstroschein:
> @NiklasBuchfink  Clean code says good variable names don't need explanation. We link content with these ids and without context inside the id,

[...]

Issue #1844, opened 2023-12-07
SDK persistence of messages in project directory


samuelstroschein:
## Context

We need persistence of messages before starting #1585. The storage format will influence the importers and exporters. 

## Scope

Test different storage persistence formats. Does splitting messages 

- [ ] implement storage persistency in SDK 
  - [ ] test persist files by message and language 
  - [ ] test persist split by message only
- [ ] "degrade" `saveMessages` and `loadMessages` to act as "importers" (until #1585 is implemented)

```
1. loadProject "loads" message from project directory
2. IF loadMessages API defined -> [NOT BLOCK UI] loadProject calls `loadMessages` -> write messages to project directory
3. CRUD operations happen on project directory
4. on CRUD operations, `saveMessages` is called to "export" messages
```

samuelstroschein:
@martin-lysk interesting observation. The per-file splitting for version control would be great as can be seen in https://www.loom.com/share/6636f233340340d9b7f5e237c922592f. I was able to re-construct how message changes based on the diff, given that the format is human-readable. 

Maybe a human-readable format would also be good for persisting our messages to enable easier diffing of messages (until we have semantic diffing and diffing visualization in @inlang/lix; @inlang/lix are pings like this one helpful to provide information where requirements are coming from?)

janfjohannes:
@samuelstroschein  yes thanks, super helpful!
this looks really great progress, please plan this in a way that we can limit the number of files to 1000 per folder long term. no need to support this straight away but keep in mind that once we support huge project we need to have a folder prefix that can limit the files per folder to ~1000 . @martin-lysk   i changed my mind regarding the namespace folder structure we talked about earlier, limiting number of files per folder would be a nice sideeffect of that structure, correct?

martin-lysk:
**Should filenames represent the message name or the message id?** - yes we have this question again ;-) 

Message id and message name are currently the same. @samuelstroschein suggested to have files without a speaking name to explain one should not touch them and make them the id of the message. A lint could be written that disallows duplicate message "names". 

_This would require_
* to introduce a message "name" as an additional property of the message that stores the imported names from importer/exports like json android or ios.
* change of the sdk api or introduce a query by name that could result in multiple messages 
* all places where we use the id need to be checked - if they meant to use the id or the name of a message (header column in the editor)
* The you would need to be able to explain why multiple messages with the same name can exist
* a lint rule that fires when we find multiple messages with the same name 

**How should the sdk handle corrupt message files?**

[...]

Issue #1585, opened 2023-11-03
Design for Import/Export APIs to replace storage plugins


samuelstroschein:
This issue has been raised by @martin-lysk. We agreed that importers/exporters is the right way to go but decided at the Berlin Offsite in Oct 23 to work around the issue as long as possible. First users are confused now why they storage plugins are limiting them.  

## Problem 

Inlang's set up to be "provide your storage plugin" leads to numerous issues: 

- inlang's features are limited by the provided storage plugin (a no-go)
- users don't understand why certain features work or don't work with a provide storage plugin (e.g. https://github.com/inlang/monorepo/issues/1577#issuecomment-1791655712)

## Proposal 

Storing messages should be controlled by the SDK to ensure that inlang is not limited by external plugins. 

- `loadMessages` should be succeeded by `importMessages`
- `saveMessages` should be succeeded by `exportMessages`
```
message-29sn82
-> json import: login.button
-> paraglide export: login_button
-> ios export: LOGIN_BUTTON
-> android export: login-button
```

**Pros**

- inlang is not limited by external plugins
- import/export APIs can expose what features they support
- multi-platform exports (export for iOS, Android, Paraglide) become possible
- users are communicated what is supported by a target platform and what not. 
- we can performance optimize the storage instead of naively calling `saveMessages()` and `loadMessages()`
- importer/exporter plugins could store additional data like message id `ieb3s` should be exported as `login-button`, or `ieb2s` exists in the namespace file `en/login.json`. requires https://github.com/inlang/monorepo/discussions/1418

**Cons**

- effort. 
- the introduction of https://github.com/inlang/monorepo/discussions/1418 should be completed with this change too. have a `project.inlang` folder to avoid massive scatter across a repo

## Requirements

- [ ] allow multiple importers and exporters to be used in a project (load and save messages only support one "storage" plugin at a time)
- [ ] must allow for "namespacing" logic see https://github.com/inlang/monorepo/issues/1577#issuecomment-1791655712. otherwise existing projects can't migrate to inlang 
- [ ] https://github.com/inlang/monorepo/issues/1769#issuecomment-1830270023
- [ ] when is export triggered? onSave [can be ignored for now]
- [ ] how to deal with creation/deletion

samuelstroschein:
@martin-lysk this seems to be a great issue for you. after all, you raised this issue and now it is hurting our growth because users don't understand why feature limitations exist for different storage formats

felixhaeberle:
[...]

PR #2108, opened 2024-01-24
WIP 1844 Part 1: auto-generated human-IDs and aliases

### LINKED ISSUE #1968: [bug] Can't `machine-translate` multiple new keys
### Problem

https://github.com/osmosis-labs/osmosis-frontend/pull/2639/commits/c564b8773ef5007e6d1b5130c3682b21132f30f6

See diff for this commit. I didn't notice as I thought it was fixed, but running machine-translate once seemed to generate the title, p1, and p3, but for some reason missed p2.

I'm wondering if it has to do with there being multiple new keys? Or the nesting?

Running multiple times as needed to get the remaining keys translated seems to work. 

### Expected behavior

Should generate translations for _all_ new keys in the ref file. Should not have to run multiple times.

### Reproduction

Add multiple new keys to the ref language file. Run machine translate. Notice some of the new keys are missing new translations.

### Other information

MacOS. Using JSON langage files.

martin-lysk:
This PR introduces auto-generated human-IDs and aliases behind the `experimental: aliases` flag. 

For background, please see #1892 and https://github.com/opral/monorepo/issues/1844#issuecomment-1894358226 and the [message docs](https://inlang.com/documentation/concept/message).

This PR also adds file-system-based locking to improve the behavior of concurrent message updates through the load/store plugin api. Most of these change are in [loadProject](https://github.com/opral/monorepo/blob/9c855da3e21dbdf53ee33d03cd1dd8e21d2cf5f7/inlang/source-code/sdk/src/loadProject.ts).

In order to validate these changes, there is also a new sdk/load-test, and a mock machine translate api. 

#### tasks

- [x] feature flag for human ids and aliases
- [x] enable feature flag for opral projects 
- [x] add message.alias (non-optional) property
- [x] autogenerate human ids and set alias = plugin id on load
- [x] message file locking on load/save, read/write
- [x] validate file locking (should fix #1968)
- [x] validate with large project
- [x] add mock translation to rpc
- [x] add test for file locking with real FS and mock translation
- [x] uncomment cli translate progress bar or make it optional
- [x] replace console logs with debug logger flag
- fink.
  - [x] show aliases alongside message ids
  - [x] seach messages by alias
- ide-extension 
  - [x] recognize message aliases returned by matchers.
  - [x] show aliases in inlang tab (design tbd)
  - (separate PR) ~~auto-generate human ids for new (e.g. extracted) messages~~
- paraglide
  - [x] compile aliases into message functions
  - [x] generate deprecation comments for aliases  
- [x] add [SDK docs](https://inlang.com/documentation/concept/message) for aliases and human IDs
- [x] revisit TODOs
- [x] PR review 
- [x] changeset
- [ ] 🚢 

...onward to 1844 Part 2: inlang message persistence

#### human ids (opral only)
Message ids from a plugin are replaced with human ids derived from the message id.  The id becomes the default `alias`.

Existing message references in code may continue to use the plugin id, now stored in the default alias. A new api `messages.getByDefaultAlias()` resolves aliases e.g. when matched by the ide-extension:
 
```ts
const highlightedMessages = matchedMessages
  // resolve messages from id or alias
  .map((message) => {
    return (
      state().project.query.messages.get({ where: { id: message.messageId } }) ??
      state().project.query.messages.getByDefaultAlias(message.messageId)
    )
  })
```   

#### In fink, the alias is shown alongside the human id
Screenshot 2024-02-05 at 11 37 40

#### the ide-extension checks for aliases as well as ids when it matches references to messages
Screenshot 2024-02-11 at 23 48 16

The `message.alias` is a non-optional property of the [message api](https://github.com/opral/monorepo/blob/9c855da3e21dbdf53ee33d03cd1dd8e21d2cf5f7/inlang/source-code/versioned-interfaces/message/src/interface.ts#L61). NOTE: This change to the TypeScript `Message` interface impacts a large number of tests. Additional modification was required in those tests which expect to be able to get() messages using the original message key.


martin-lysk:
Hey we just had a great call

To give you an update on Persistence checkout the following to understand HOW

We introduce the alias system and we don't try to run our own persistence and the plugin storage at the same time - its one OR the other.

Our new persistence will be only running ONLY if NO loadMessages and saveMessages plugin is configured AND for now only on opral repos.
Users that configured a load and save plugin will see a deprecation some time in the future - at that time we have an import function that can be triggered manually
We will dog food persistence in our own projects and sample projects
When we are happy - remove the opral repo constraint and it can be used for project create in inlang.manage

Times follow next week!

To understand WHY this way - and get a grasp why this takes so long:

Having messages synced with (1.) a JSON file or whatsoever - the plugin names it - (2.) our own persistence (one file per Message) and multiple SDK(s) running in different processes is a classical example of the consensus Problem (https://en.wikipedia.org/wiki/Consensus_(computer_science) ). 

To deal with two representations in the file system didn't make thinks easier - and side effects are hard to tackle:

Take a rather simple example:
Someone edits one message in the message.json (managed by the plugin)
The sdk updates the messages in the sdk and informs the plugin that messages have been changed -> writes the message.json (for those thinking Quick here: No we can't know who initialized the change and skip that step - since we have to relay on fs.watch)
1.a A message is edited in the message.json 

Outcome: 
Change in 1.a is overwritten by 2. 

And this is just one of many cases....

While this can't happen in a single threaded env like Fink or Node and or at least can be synchronized things become nasty if multiple processes are involved with two target stores with fs properties 😕 UX will be way to bad!

jldec:
> Our new persistence will be only running ONLY if NO loadMessages and saveMessages plugin is configured AND for now only on opral repos

I have updated the OP and tasklist to reflect this.

inlang/source-code/sdk/src/loadProject.ts
```diff
@@ -448,3 +846,420 @@ export function createSubscribable<T>(signal: () => T): Subscribable<T> {
...
+			const messagesToExport: Message[] = []
+			for (const message of currentMessages) {
+				if (messageDirtyFlags[message.id]) {
+					const importedEnecoded = stringifyMessage(message)
+					// TODO #1844 use hash instead of the whole object JSON to save memory...
+					persistedMessageHashs[message.id] = importedEnecoded
+				}
+
+				const fixedExportMessage = { ...message }
+				// TODO #1585 here we match using the id to support legacy load message plugins - after we introduced import / export methods we will use importedMessage.alias
```
samuelstroschein, line 1050: Would this code become redundant if `repo.fs.readFile()` would be "reactive" and take care of triggering only if a file, and thus message, changed?

inlang/source-code/sdk/src/createNodeishFsWithAbsolutePaths.ts
```diff
@@ -42,5 +42,11 @@ export const createNodeishFsWithAbsolutePaths = (args: {
 			path: string,
 			options: { signal: AbortSignal | undefined; recursive: boolean | undefined }
 		) => args.nodeishFs.watch(makeAbsolute(path), options),
+		// This might be surprising when symlinks were intended to be relative
```
jldec, line 45: @martin-lysk i added implementations for the new methods which you had already added to the test. Could you check that they look ok.

Not 100% sure why we need this wrapper at all, but will focus on getting tests to pass for this PR.

inlang/source-code/sdk/src/test-utilities/createMessage.test.ts
```diff
@@ -35,32 +52,31 @@ test("should create a message with pattern", () => {
 				{ type: "Text", value: '"' },
 			],
 		})
-	).toMatchInlineSnapshot(`
+	).toStrictEqual({
```
jldec, line 55: I wasn't sure why we were string matching here but it seemed a bit off - do you know @martin-lysk ?

inlang/source-code/sdk/src/createNodeishFsWithAbsolutePaths.ts
samuelstroschein: @jldec We need this wrapper because relative paths do not exist in non-server environments. 

Any path that is resolved by the SDK needs to be absolute and resolve from the project directory itself. Alternatively, we could set a fake CWD maybe

jldec: thanks for clearing that up - @samuelstroschein

jldec:
@NiklasBuchfink I have inserted an image of the modified fink UI showing the id nest to the alias. 

Would you mind reviewing the implementation to make sure that is what you expected. Happy to discuss if modifications are needed. I do wonder if we should add additional labels ('alias' and 'id') to make it a bit clearer.

cc: @samuelstroschein

NiklasBuchfink (approved):

jldec:
@felixhaeberle I added `getByDefaultAlias()` to the MessageQueryApi, and added it to a few places in the ide-extension where matched messages were being retrieved with just `project.query.messages.get({ where: { id: message.messageId } })`. This API is non-subscribable -- it only makes sense for resolving matched messages 

I also added a screenshot and code-sample in the OG above.

Thesee are the TODOs for the ide-extension that I noted:
1. Improve the message display in the inlang tab to show message ids with aliases.
2. Fix the edit (pencil) links in the ide tooltip for matched messges in the code which use an alias (I can take a shot at this)

#### Regarding the current prompt for a new message ID during extract
I propose that we leave the current behavior unchanged in this PR, and introduce auto-generated ids, for internally-stored messages only. 

In order to replace the prompt with auto-generated ids for all plugins in this PR, we would have to store those human-ids in the plugin-managed message files, adding new requirements to the alias logic, like recognizing and not-creating aliases for plugin-owned messages with human-ids.

cc: @samuelstroschein @martin-lysk - happy to hear your opinions on these points too.

samuelstroschein:
>  This API is non-subscribable -- it only makes sense for resolving matched messages

Inconsistent API behavior and the assumption seems wrong. 

If an app wants to resolve a message by alias, an app also wants to subscribe to it. At any given point, a message can change. The UI needs to reflect such changes. Such change can also be an alias. Expecting frontend developers to somehow manually handle those changes (only manual polling comes to mind) will result in bug reports from users and frustrated inlang ecosystem developers. 

**Example - IDE extension**

@felixhaeberle correct me if I am wrong. You can use 👍 to indicate agreement.

```
1. User sees messages in the tab view searched by alias
2. User changes the alias of a message in the resource file
3. Tab view doesn't update because `getMessageByAlias()` is not reactive. Bug report coming 💥 
```

**Exampe = Fink**

@NiklasBuchfink correct me if i am wrong please. Indicate with 👍 if you agree

```
1. Some component in fink allows searching for alias and lists the results 
2. User updates the alias of a message in the search results in a different component
3. 💥 search component list doesn't update because `getMessageByAlias()` is not reactive
```

In both examples, the component responsible for each has no knowledge about another component (or even an external change!) mutating the message and, therefore, must be able to subscribe. Sure, we can say, "developers should deal with this themselves," but that misses the point of the SDK. The job of the SDK team is to take as much work from app developers as possible. 

This comment is a good example of my belief that every property in the inlang `project.*` API and lix `repo.*` API needs to be reactive. Any assumption of non-reactivity will lead to bugs in apps because close to every API is derived state.

NiklasBuchfink:
@samuelstroschein The list of messages only updates when one of the search and filter parameters (lint, language, search, id) changes. Therefore the filter and search function itself is **not** a derived signal that depends on reactivity. Otherwise:
-> You would start typing
-> This resolves the "missing translation rule" that is set as a filter
-> The message is removed from the list (User: "Where is my message? I want to read it a second time!")

Currently, there is no option to update the alias in Fink. Do you think this should be added?

@jldec I added the alias to the search function (showFilteredMessage.ts).

jldec:
@samuelstroschein I have made getByDefaultAlias subscribable - I will add a test for that also.

We were not planning to include the new persistence in this PR. We thought it made sense to avoid the "multi-master" issues of syncing messages across 2 storage locations, and to reduce the scope of this PR, @martin-lysk removed the persistence code, for use in a followup PR.

This also means that to avoid storing our ids in resource files from plugins, we won't prompt for ids for new messages yet.

Happy to discuss this if you still have reservations.

samuelstroschein:
> This also means that to avoid storing our ids in resource files from plugins, we won't prompt for ids for new messages yet.

@jldec but how do human readable ids get introduced then if they are not saved somewhere?

jldec:
> @jldec but how do human readable ids get introduced then if they are not saved somewhere?

@samuelstroschein my understanding is that human-ids are generated deterministically from the incoming plugin message IDs, allowing references (by human id) from code, without modifying the ids in the plugin stored files. 
cc: @martin-lysk  

> When messages are loaded via plugin loadMessages, the message id is replaced with a human ID derived from the message content. The original id from the plugin becomes the default alias. [loadProject L666](https://github.com/opral/monorepo/blob/9c855da3e21dbdf53ee33d03cd1dd8e21d2cf5f7/inlang/source-code/sdk/src/loadProject.ts#L666)

martin-lysk:
Yes the human id's are a computed hash based on the message id coming from the load plugin - this is predictable and doesn't change when the message name stays consistent.

samuelstroschein:
Understood. @martin-lysk @jldec Not deploying human-readable IDs together with storage has a high risk of broken UX in the entire ecosystem because IDs potentially change everytime a project is loaded https://www.loom.com/share/c94f307c60884fe083bd359e9fa748c1. 

1. Did you propose to split the PR but **not** deploy the changes until persistency is introduced?  
2. If not, the loom should make you reconsider deploying both changes independently.

martin-lysk:
> Yes the human id's are a computed hash based on the message id coming from the load plugin - this is predictable and doesn't change when the message name stays consistent.

So the short answer to your loom is no - we dont use the content. we use the message id comming from the plugin - the one we put into the alias on load.

samuelstroschein:
@martin-lysk read my edit. using the key only is even worse. we risk multiple messages with the same id now

samuelstroschein:
we introduced the derivation of message ids from existing keys **to avoid migration merge conflicts**. after the migration, the ID "system" was supposed to take over and ensure uniqueness and immutability. what you are proposing is to ship an ID migration with no ID properties of immutability and uniqueness. 

ecosystem users (especially @opral/inlang-paraglide-js due to https://github.com/opral/monorepo/pull/2212!) will start to use message ids only to be burned because the id changes when someone renames the alias/key. this is a recipe for disaster and calls for "human ids are sh**. i want keys". 

if you want to split the PRs okay, but deploy them in one go (aka one PR must be a large one). I am open to discussions but it seems like you are going the wrong path here

samuelstroschein:
**Call with @martin-lysk **

1. we will ship this PR #2108 and follow up https://github.com/opral/monorepo/issues/1585 and https://github.com/opral/monorepo/issues/1844 behind feature flag "is opral owned project" to test out all 
2. nothing changes for users until #2108, storage and importer/exportet are complete and bulletproof

**Why**?

Users will have no benefit of human ids until https://github.com/opral/monorepo/issues/1844 and https://github.com/opral/monorepo/issues/1585 are deployed. We don't want to alienate users in the meantime. 

**Tasks**

- [ ] apps (paraglide js, fink, ide extension) need conditional logic if `message.alias.default` exists 
- [ ] SDK needs conditional logic for "is opral repo" 

cc @martin-lysk @jldec

jldec:
Putting aliases and human-ids behind a feature flag for this PR makes sense.
I have updated the task list in the OP
cc @martin-lysk .

inlang/source-code/editor/src/pages/@host/@owner/@repository/Message.tsx
```diff
@@ -103,9 +103,11 @@ export function Message(props: { id: string }) {
 					slot="summary"
 					class="flex-grow-0 flex-shrink-0 text-[13px] font-medium text-left text-on-surface before:text-on-surface"
 				>
-					{message() && Object.keys(message()!.alias).length > 0
-						? message()!.alias[Object.keys(message()!.alias)[0]!]
-						: "alias"}
+					{
+						message() && Object.keys(message()!.alias).length > 0
+							? message()!.alias[Object.keys(message()!.alias)[0]!]
+							: "" /* // TODO featureFlag // "alias" */
```
jldec, line 109: better to show nothing if there is no alias
(no need for feature flag)

inlang/source-code/versioned-interfaces/project-settings/src/interface.ts
```diff
@@ -76,6 +76,9 @@ const InternalProjectSettings = Type.Object({
 			],
 		})
 	),
+	featureFlags: Type.Optional(
+		Type.Record(Type.String(), Type.Union([Type.Boolean(), Type.Number(), Type.String()]))
+	),
```
jldec, line 81: @samuelstroschein what do you think about adding featureFlags to project settings?
This would simplify testing of projects with/without a feature flag
We could still have logic to feature-flag by owner/repo in addition to this?

inlang/source-code/versioned-interfaces/project-settings/src/interface.ts
```diff
@@ -76,9 +76,7 @@ const InternalProjectSettings = Type.Object({
 			],
 		})
 	),
-	featureFlags: Type.Optional(
-		Type.Record(Type.String(), Type.Union([Type.Boolean(), Type.Number(), Type.String()]))
-	),
+	featureFlags: Type.Optional(Type.Record(Type.String(), Type.Boolean())),
```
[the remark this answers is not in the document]
jldec: @samuelstroschein i was going for ease of manual testing (just toggle the flag between true and false without having to lookup the flag name and recreate the line) but I don't mind forcing it to literal "true" if that encourages folks to remove old flags.
```ts
	featureFlags: Type.Optional(Type.Record(Type.String(), Type.Literal(true))),
```

Here's a CLI experience after changing the value from `true` to `false` 

```
$ inlang machine translate --project project.inlang/

 WARN  Human translations are better than machine translations.                                                                                                  5:08:39 PM

We advise to use machine translations in the build step without commiting them to the repo. By using machine translate in the build step, you avoid missing translations in production while still flagging to human translators that transaltions are missing. You can use the force flag (-f, --force) to skip this prompt warning.

✔ Are you sure you want to machine translate? … yes

 ERROR  The project settings are invalid:                                                                                                                        5:08:41 PM

"/featureFlags/aliases":

Expected true

  
  "/featureFlags/aliases":
  
  Expected true
  at parseSettings (/Users/jldec/inlang/monorepo/inlang/source-code/cli/dist/main.js:109522:13)
  at loadSettings (/Users/jldec/inlang/monorepo/inlang/source-code/cli/dist/main.js:109515:10)
```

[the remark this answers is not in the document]
samuelstroschein: the testing one is a good argument at the cost of outdated feature flags. what trade off should we do here? 

- we don't want people to manually edit .inlang files (argument against testing stuff) 
- an automatic clean up of outdated feature flags will be just like just as annoying for testing than removing the key entirely
- have a manual clean up functions requires apps to build show feature

an alternative might be a map/array to avoid the weird setting to true thing

```
featureFlags: [
  "aliases",
  "humanIds"
]
```

```ts
if (project.settings.featureFlags.includes("aliases"){

}
```

@jldec what do you think about the array/map approach?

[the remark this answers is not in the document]
jldec: I prefer the map, even with the forced true (maps are more easily tested with `featureflags?.foo`)
I don't think we really gain a lot by preventing "false" - we'd gain more by maintaining the list of valid feature flags somewhere and validating against that - but I don't want to do it in the interface declaration because that seems too hard to release and rollout changes (maybe I'm missing the process for that).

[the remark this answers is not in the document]
samuelstroschein: > we'd gain more by maintaining the list of valid feature flags somewhere and validating against that

We will likely not be able to validate the feature flags:

- apps will use the feature to drop their own feature flags into the settings 
- an old SDK version can't crash once new feature flags are added to the validation list



@jldec i vote for only setting to `true` for now. we can loosen the restriction in the future. the beauty of allowing only `true` is a) automatic removal of unused feature flags and b) one way to achieve things:  

**Comparing both**

1. loading a project can get slower: 100 feature flags increase the size of the settings json by approx 1.5kb. assuming that apps will also use the feature flag feature, i easily foresee 100's of feature flags "rotten" in the settings file of which only a few are active.

2. multiple ways to achieve the same thing (remove a key or set it to false. which method should be preferred?)  

```ts

// 2 ways to achieve the same things :/ 
project.settings.featureFlags.aliases = false
project.settings.featureFlags.aliases = undefined

```

**Conclusion**

can we agree on: 

1. `settings.featureFlags` should not be undefinable to increase ergonomics of using the API (have a migration that creates sets feature flags to `{}` if not existent).
2. have a map instead of an array
3. only allow `true` for now. can be loosened in the future. 
4. have a good error message when a feature flag is set to false that prompts the user to remove the key entirely.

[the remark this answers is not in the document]
jldec: not sure how to implement 4, but will work on 1-3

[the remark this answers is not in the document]
jldec: > have a good error message when a feature flag is set to false 

The more important challenge will to clean up code and config for feature flags that are no longer needed. Migration of settings to auto-remove those would be a big win (future.)

I can take another looking, but I'm not convinced that we should add extra code now to warn users (or change the default error message) when the value of the flag is set to false.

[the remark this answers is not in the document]
samuelstroschein: > When we ship the new project-settings schema, old SDKs and old projects, need to keep loading without error - we don't want to force everyone to upgrade (break their projects) with this PR.

Argh. I forgot about this. 

In the future, we should have a public telemetry dashboard (cc @jannesblobel) on API usage of the SDK over time and formulate a rule like "if a deprecated API is used by less than X% of projects, it will be removed with the next major version". I hope that such a public dashboard aligns our users with deprecations. 

In this example, we could formulate an insight/rule/deprecation like "once less than 5% of projects use SDK version <X (which includes the featureFlag migration), feature flags can not be undefined"

inlang/source-code/cli/src/commands/machine/translate.ts
```diff
@@ -52,6 +52,7 @@ export async function translateCommandAction(args: { project: InlangProject }) {
 			log.error(`No inlang project found`)
 			return
 		}
+		const ffAliases = args.project.settings().featureFlags?.aliases
```
samuelstroschein, line 55: had to double check what `ff` is

```suggestion
		const aliasesFeatureFlag = args.project.settings().featureFlags?.aliases
```

inlang/source-code/sdk/src/adapter/solidAdapter.test.ts
```diff
@@ -278,6 +350,73 @@ describe("messages", () => {
 			).value
 		).toBe("test2")
 	})
+
+	it("should react to message delete", async () => {
+		const repo = await mockRepo()
+		const fs = repo.nodeishFs
+		await fs.mkdir("/user/project.inlang.inlang", { recursive: true })
```
samuelstroschein, line 357: the tests seem ill-placed. 

the solid adapter acts as reactivity bridge. the only tests that are required is briding reactivity, not testing business logic. the added tests (and old tests too btw) seem to test business logic.

inlang/source-code/sdk/src/resolve-modules/plugins/types.ts
```diff
@@ -18,7 +18,7 @@ import type { ProjectSettings } from "@inlang/project-settings"
  */
 export type NodeishFilesystemSubset = Pick<
 	NodeishFilesystem,
-	"readFile" | "readdir" | "mkdir" | "writeFile" | "watch"
+	"readFile" | "readdir" | "mkdir" | "writeFile" | "watch" | "rm" | "stat"
 >
```
samuelstroschein, line 22: which plugins need access to `rm` and `stat`?

inlang/source-code/sdk/src/storage/human-id/human-readable-id.ts
```diff
@@ -0,0 +1,27 @@
+import murmurhash3 from "murmurhash3js" // we use murmur for best distribution https://medium.com/miro-engineering/choosing-a-hash-function-to-solve-a-data-sharding-problem-c656259e2b54
```
samuelstroschein, line 1: increase readability by moving the comment above the import, not next to the import

```suggestion
// we use murmur for best distribution https://medium.com/miro-engineering/choosing-a-hash-function-to-solve-a-data-sharding-problem-c656259e2b54
import murmurhash3 from "murmurhash3js"
```

inlang/source-code/sdk/src/storage/helper.ts
```diff
@@ -0,0 +1,52 @@
...
+
+	// Extract until the last dot (excluding the dot)
+	return messageFileName.slice(0, Math.max(0, lastDotIndex))
+}
+
+export function getPathFromMessageId(id: string) {
+	const path = id.replace("_", "/") + fileExtension
+	return path
+}
+
```
samuelstroschein, line 24: Exporting individual functions instead of nesting them in objects like `utils.getMessageIdFromPath` enables tree-shaking. Good choice 👍

inlang/source-code/sdk/src/storage/helper.ts
```diff
@@ -0,0 +1,52 @@
...
+export function parseMessage(messagePath: string, messageRaw: string) {
+	return JSON.parse(messageRaw)
+}
```
samuelstroschein, line 27: why is this function needed? IT just calls JSON.parse

inlang/source-code/sdk/src/createMessagesQuery.test.ts
```diff
@@ -19,6 +20,19 @@ describe("create", () => {
 		expect(created).toBe(true)
 	})
 
+	it("should create a message with a default alias", () => {
```
samuelstroschein, line 23: The reasoning is missing in the test description on why a default alias should automatically be created. 

I don't understand why this should happen. This seems wrong due to implicit behaviour that we won't be able to remove in the future to ensure backwards compatibility. If this logic exists to ensure compatibility with `loadMessages()` and `saveMessags()`, I recommended to process every message passed to `saveMessages()` and (dynamically) create the dfault alias.

inlang/source-code/sdk/src/loadProject.test.ts
```diff
@@ -52,7 +52,7 @@ const mockPlugin: Plugin = {
 	description: { en: "Mock plugin description" },
 	displayName: { en: "Mock Plugin" },
 
-	loadMessages: () => exampleMessages,
+	loadMessages: () => exampleMessages as Message[],
```
samuelstroschein, line 55: the casting seems redundant. as typescript identifies the messages as `Message[]`


```suggestion
	loadMessages: () => exampleMessages,
```

inlang/source-code/sdk/src/loadProject.ts
```diff
@@ -448,3 +561,450 @@ export function createSubscribable<T>(signal: () => T): Subscribable<T> {
 		},
 	})
 }
+
+// --- serialization of loading / saving messages.
```
samuelstroschein, line 565: 1. move the code to `storage/locking` to declutter `loadProject`? 

2. @martin-lysk @jldec are you sure that the variables should not be scoped within `loadProject`?

inlang/source-code/sdk/package.json
```diff
@@ -39,8 +39,10 @@
 		"@inlang/result": "workspace:*",
 		"@inlang/translatable": "workspace:*",
 		"@sinclair/typebox": "^0.31.17",
+		"@types/murmurhash3js": "^3.0.7",
```
samuelstroschein, line 42: @types are a dev depenedency

inlang/source-code/versioned-interfaces/plugin/src/interface.ts
```diff
@@ -19,7 +19,7 @@ import type { ExternalProjectSettings, ProjectSettings } from "@inlang/project-s
  */
 export type NodeishFilesystemSubset = Pick<
 	NodeishFilesystem,
-	"readFile" | "readdir" | "mkdir" | "writeFile" | "watch"
+	"readFile" | "readdir" | "mkdir" | "writeFile" | "watch" | "rm" | "stat"
```
samuelstroschein, line 22: I repeat the question here. Which plugin needs access to `rm` and `stat`? 

The subset exists to keep APIs that plugins use under control and ease a migration to the upcoming lix fs API

inlang/source-code/versioned-interfaces/message/src/interface.ts
```diff
@@ -58,6 +58,7 @@ export const Variant = Type.Object({
 export type Message = Static<typeof Message>
 export const Message = Type.Object({
 	id: Type.String(),
+	alias: Type.Record(Type.String(), Type.String()),
```
samuelstroschein, line 61: Shouldn't the alias by undefineable to ensure backwards compatibility similar to `settings.featureFlags`?

lix/source-code/client/src/helpers.ts
```diff
@@ -18,7 +18,8 @@ export const withLazyFetching = ({
 }: Args): NodeishFilesystem => {
 	return new Proxy(nodeishFs, {
 		get(getTarget: typeof nodeishFs, prop, receiver) {
-			if (getTarget[prop as keyof typeof nodeishFs]) {
+			// eslint-disable-next-line no-prototype-builtins
+			if (((getTarget as any).__proto__ as any).hasOwnProperty(prop)) {
```
samuelstroschein, line 22: why is the proto access needed? add a comment above please

project.inlang/messages/v1/after/dragon_scram_inside.json
```diff
@@ -0,0 +1,79 @@
...
+    },
+    "id": "after_dragon_scram_inside",
+    "selectors": [],
+    "variants": [
+        {
+            "languageTag": "de",
+            "match": [],
+            "pattern": [
+                {
+                    "type": "Text",
```
samuelstroschein, line 13: committing this file is intended?

inlang/source-code/sdk/src/resolve-modules/plugins/types.ts
martin-lysk: Good catch. We need those for file locking - and this was a Problem with the way loadProject was structured when we had both persistency and plugins in this PR should be changed again now.

inlang/source-code/sdk/src/storage/helper.ts
martin-lysk: not used and can be removed leftover from our persistence

inlang/source-code/cli/src/commands/machine/translate.ts
jldec: 👍 
Agree - will fix across both files.

inlang/source-code/sdk/src/adapter/solidAdapter.test.ts
jldec: Will remove the delete test from here. They don't go very deep anyway.
Deletion in now tested in loadProject.test and covers similar ground.

inlang/source-code/sdk/src/storage/human-id/human-readable-id.ts
jldec: 👍

inlang/source-code/sdk/src/storage/helper.ts
jldec: will remove

inlang/source-code/sdk/src/createMessagesQuery.test.ts
jldec: Agree - changed the test description to 
`query.getByDefaultAlias should return a message with a default alias`

inlang/source-code/sdk/src/loadProject.test.ts
jldec: 👍

inlang/source-code/sdk/src/loadProject.ts
jldec: Will cleanup this file in separate PR - it's still helping us to have these together while we debug locking logic.

inlang/source-code/sdk/package.json
jldec: 👍

lix/source-code/client/src/helpers.ts
jldec: @martin-lysk could you fill in reason for this please

project.inlang/messages/v1/after/dragon_scram_inside.json
jldec: @martin-lysk was this intentional?

> $ git log project.inlang/messages/v1/after/dragon_scram_inside.json
>
> commit 93102a11c4a1c62f9ca9aa5da4873f37f735a4dd
> Author: Martin Lysk <martin.lysk@gmail.com>
> Date:   Thu Jan 25 13:02:52 2024 +0100
>
>     #1844 adds one message for a test on git

inlang/source-code/cli/src/commands/machine/translate.ts
samuelstroschein: btw what about calling feature flags `experimental`? aligns with communication of other apps that use feature flags e.g. node cli, github etc. 

```diff
-aliasesFeatureFlag
+experimentalAliases
```

```diff
-args.project.settings().featureFlags.aliases
+args.project.settings().experimental.aliases
```

this is a naming discussion. no large impact expected. hence, either name it feature flags or experimental. both will be fine :)

project.inlang/messages/v1/after/dragon_scram_inside.json
martin-lysk: Nope not anymore

lix/source-code/client/src/helpers.ts
martin-lysk: I drop it

lix/source-code/client/package.json
```diff
@@ -27,7 +27,7 @@
 	"dependencies": {
 		"@lix-js/fs": "workspace:*",
 		"@octokit/types": "12.4.0 ",
-		"isomorphic-git": "1.24.5",
+		"isomorphic-git": "git@github.com:martin-lysk/isomorphic-git.git#build",
```
jldec, line 30: @martin-lysk I assume this was left behind by mistake right?

inlang/source-code/versioned-interfaces/plugin/src/interface.ts
jldec: resolved.

inlang/source-code/sdk/src/loadProject.ts
jldec: @samuelstroschein I moved the message load and save state up into loadProject scope
@martin-lysk offered to review
https://github.com/opral/monorepo/pull/2108/commits/b066babde301e54277c9e3ee7ab2d501a5cbf4ed

inlang/source-code/sdk/src/test-utilities/createMessage.test.ts
martin-lysk: TBH - no i think toMatchInlineSnapshot should be enough

inlang/source-code/versioned-interfaces/message/src/interface.ts
martin-lysk: Objects from type Message currently only exist in memory - even if one plugin uses the AST to store them - this one will be replaced by the import export before the feature flag is active which will take care of that.
 
Making the alias type non optional saves us an extra check if alias is present and the overall structure is cleaner - i would say - we keep it like that.

inlang/source-code/sdk/src/loadProject.ts
```diff
@@ -154,9 +189,25 @@ export async function loadProject(args: {
...
+
+		const [saveMessagesViaPluginError, setSaveMessagesViaPluginError] = createSignal<
+			Error | undefined
+		>()
+
+		const messagesQuery = createMessagesQuery(() => messages())
+
+		// const messageBaseFolderFolderPath = projectPath + "/messages"
+		// const messageFolderPath = messageBaseFolderFolderPath + "/v1"
+		const messageLockFilePath = projectPath + "/messagelock"
```
martin-lysk, line 207: We should add a .gitignore underneath the inlang folder and add the messagelock file to it

jldec: How about renaming lockFilePath => lockDirPath
I'm not sure we need the gitignore since the directory is empty.

I added sentry logging to the tasklist, so when we find a leftover lock dir (unexpected) it would be nice to record that.

inlang/source-code/sdk/src/createNodeishFsWithWatcher.ts
```diff
@@ -7,9 +7,9 @@ import type { NodeishFilesystemSubset } from "@inlang/plugin"
  * The paths are resolved from the `projectPath` argument.
  */
 export const createNodeishFsWithWatcher = (args: {
-	nodeishFs: NodeishFilesystemSubset
+	nodeishFs: NodeishFilesystem
 	updateMessages: () => void
-}): NodeishFilesystemSubset => {
+}): NodeishFilesystem => {
```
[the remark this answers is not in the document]
jldec: @martin-lysk do you know which of the changes introduced in https://github.com/opral/monorepo/commit/6e6359cd4bc34d11cc7fd8e2927a3c9cec41e713 are still necessary?

inlang/source-code/sdk/src/loadProject.ts
martin-lysk: hahaha nice one! you are right - the naming confused me as well. its an empty folder no need to git ignore - thats great

inlang/source-code/ide-extension/src/diagnostics/linterDiagnostics.ts
```diff
@@ -25,58 +25,66 @@ export async function linterDiagnostics(args: { context: vscode.ExtensionContext
...
-						where: {
-							messageId: message.messageId,
+				// resolve message from id or alias
+				const _message =
+					state().project.query.messages.get({
+						where: { id: message.messageId },
+					}) ?? state().project.query.messages.getByDefaultAlias(message.messageId)
+
+				if (_message) {
+					state().project.query.messageLintReports.get.subscribe(
```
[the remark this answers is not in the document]
samuelstroschein: i thought via the scope of the solid root. nobody tested if this works with a reactivity bridge though

inlang/source-code/sdk/src/loadProject.ts
samuelstroschein: >  folder no need to git ignore  

in case you need a gitignore, introducing a gitignore seems OK. paraglide js also uses a generated gitignore file.

jldec: renamed lockFilePath to lockDirPath in the code.

inlang/source-code/ide-extension/src/diagnostics/linterDiagnostics.ts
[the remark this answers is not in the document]
jldec: Every call to subscribe has the side effect of calling createEffect.
It's unclear to me whether those are automatically disposed or not. 😬 
What happens if not? Would such a leak be benign or would it produce errors?

inlang/source-code/sdk/src/loadProject.ts
```diff
@@ -165,28 +219,56 @@ export async function loadProject(args: {
...
 			const fsWithWatcher = createNodeishFsWithWatcher({
 				nodeishFs: nodeishFs,
+				// this message is called whenever a file changes that was read earlier by this filesystem
+				// - the plugin loads messages -> reads the file messages.json -> start watching on messages.json -> updateMessages
 				updateMessages: () => {
-					loadAndSetMessages(nodeishFs)
+					// preserving console.logs as comments pending #
+					// eslint-disable-next-line no-console
+					debug("load messages because of a change in the message.json files")
+					// TODO FINK check error handling for plugin load methods (triggered by file change) -> move to separate ticket
```
martin-lysk, line 239: Create a ticket together with @NiklasBuchfink for this TODO and link it via the issue number.

inlang/source-code/sdk/src/loadProject.ts
```diff
@@ -165,28 +219,56 @@ export async function loadProject(args: {
...
+						.then(() => {
+							if (loadMessagesViaPluginError() !== undefined) {
+								setLoadMessagesViaPluginError(undefined)
+							}
+						})
 				},
 			})
 
-			loadAndSetMessages(fsWithWatcher)
+			// TODO FINK check error handling for plugin load methods (initial load)
```
martin-lysk, line 257: Create a ticket for this todo and link it via the issue number  - same ticket as above

inlang/source-code/sdk/src/loadProject.ts
```diff
@@ -309,6 +458,10 @@ export async function loadProject(args: {
 			errors: createSubscribable(() => [
 				...(initializeError ? [initializeError] : []),
 				...(resolvedModules() ? resolvedModules()!.errors : []),
+				...(loadMessagesViaPluginError() ? [loadMessagesViaPluginError()!] : []),
```
martin-lysk, line 461: Those are the errors that we should report in fink as well (for clarification with @NiklasBuchfink )

martin-lysk (commented):
@jldec The only thing that i found from logical perspective is the usage of the alias also without the flag set (we use the existence of the property in messages in fink and the fallback to getByDefault alias even if the flag is not set). Not so clean but we could leave it like that - you decide ;-)

I will also push another commit with some cleanups like removing uneeded // eslint-disable-next-line no-console since you use debug now.

There are some cleanup tasks like Todos that we introduce that are not tracked i would suggest to fix this before they hit main. (I will suggest to use a lint rule for that in a discussion at the onsite).

inlang/source-code/editor/src/pages/@host/@owner/@repository/Message.tsx
```diff
@@ -113,6 +113,14 @@ export function Message(props: { id: string }) {
 				>
 					<IconLink />
 				</div>
+				<h3
```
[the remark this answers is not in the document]
jldec: yes - I didn't feature flag the alias UI, just alias creation, so this should be fine - thanks for checking.

inlang/source-code/editor/src/pages/@host/@owner/@repository/helper/showFilteredMessage.ts
```diff
@@ -49,7 +49,10 @@ export const showFilteredMessage = (message: Message | undefined) => {
 	const filteredBySearch =
 		searchLower.length === 0 ||
 		(message !== undefined &&
-			(message.id.toLowerCase().includes(searchLower) || patternsLower.includes(searchLower)))
+			(message.id.toLowerCase().includes(searchLower) ||
+				(typeof message.alias === "object" &&
```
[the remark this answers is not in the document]
jldec: https://github.com/opral/monorepo/issues/2346

inlang/source-code/ide-extension/src/decorations/contextTooltip.ts
```diff
@@ -32,9 +32,11 @@ export function contextTooltip(
...
 	>[number]
 ) {
-	const message = state().project.query.messages.get({
-		where: { id: referenceMessage.messageId },
-	})
+	// resolve message from id or alias
+	const message =
+		state().project.query.messages.get({
+			where: { id: referenceMessage.messageId },
+		}) ?? state().project.query.messages.getByDefaultAlias(referenceMessage.messageId)
```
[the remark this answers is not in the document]
jldec: Yes - only alias creation is feature flagged

inlang/source-code/ide-extension/src/decorations/messagePreview.ts
```diff
@@ -46,9 +46,11 @@ export async function messagePreview(args: { context: vscode.ExtensionContext })
...
 
 			return messages.map(async (message) => {
-				const _message = state().project.query.messages.get({
-					where: { id: message.messageId },
-				})
+				// resolve message from id or alias
+				const _message =
+					state().project.query.messages.get({
+						where: { id: message.messageId },
+					}) ?? state().project.query.messages.getByDefaultAlias(message.messageId)
```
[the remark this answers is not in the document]
jldec: Yes - only alias creation is feature flagged

inlang/source-code/ide-extension/src/diagnostics/linterDiagnostics.ts
```diff
@@ -25,58 +25,66 @@ export async function linterDiagnostics(args: { context: vscode.ExtensionContext
...
 			for (const message of messages) {
-				state().project.query.messageLintReports.get.subscribe(
-					{
-						where: {
-							messageId: message.messageId,
+				// resolve message from id or alias
+				const _message =
+					state().project.query.messages.get({
+						where: { id: message.messageId },
+					}) ?? state().project.query.messages.getByDefaultAlias(message.messageId)
```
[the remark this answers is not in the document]
jldec: Yes - only alias creation is feature flagged

inlang/source-code/ide-extension/src/utilities/messages/messages.ts
```diff
@@ -127,8 +127,12 @@ export function createMessageWebviewProvider(args: {
 				).flat()
 
 				const highlightedMessages = matchedMessages
+					// resolve messages from id or alias
 					.map((message) => {
-						return state().project.query.messages.get({ where: { id: message.messageId } })
+						return (
+							state().project.query.messages.get({ where: { id: message.messageId } }) ??
+							state().project.query.messages.getByDefaultAlias(message.messageId)
```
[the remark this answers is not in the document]
jldec: Yes - only alias creation is feature flagged

inlang/source-code/sdk/src/createMessagesQuery.ts
```diff
@@ -13,19 +13,35 @@ export function createMessagesQuery(
 	// @ts-expect-error
 	const index = new ReactiveMap<string, Message>()
 
+	// Map default alias to message
+	// Assumes that aliases are only created and deleted, not updated
+	// TODO handle updates to aliases
+	// TODO refine to hold messageId[], if default alias is not unique
```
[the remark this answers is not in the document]
jldec: https://github.com/opral/monorepo/issues/2346

martin-lysk (commented):
Some followup comments - but we are getting there

inlang/source-code/sdk/src/createMessageLintReportsQuery.ts
```diff
@@ -21,6 +22,7 @@ export function createMessageLintReportsQuery(
 	settings: () => ProjectSettings,
 	installedMessageLintRules: () => Array<InstalledMessageLintRule>,
 	resolvedModules: () => Awaited<ReturnType<typeof resolveModules>> | undefined,
+	// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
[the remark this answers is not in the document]
martin-lysk: Great i removed the used parameter and pushed a new commit that resolves this

inlang/source-code/sdk/src/createMessageLintReportsQuery.ts
```diff
@@ -10,8 +10,9 @@ import type { resolveModules } from "./resolve-modules/index.js"
 import type { MessageLintReport, Message } from "./versionedInterfaces.js"
 import { lintSingleMessage } from "./lint/index.js"
 import { ReactiveMap } from "./reactivity/map.js"
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
 import { debounce } from "throttle-debounce"
```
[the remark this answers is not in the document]
martin-lysk: removed this

inlang/source-code/sdk/.eslintrc.json
```diff
@@ -1,17 +1,4 @@
 {
-	"rules": {
-		"no-restricted-imports": [
```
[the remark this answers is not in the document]
martin-lysk: ah got it its moved from rules to overrides

inlang/source-code/sdk/src/loadProject.ts
jldec: removed TODO (per @NiklasBuchfink all errors are handled and shown in Fink)
added #2358 to improve Fink validation of error handling

inlang/source-code/sdk/src/loadProject.ts
jldec: removed TODO (per @NiklasBuchfink all errors are handled and shown in Fink)
added #2358 to improve Fink validation of error handling

inlang/source-code/sdk/src/loadProject.ts
jldec: per @NiklasBuchfink all errors are handled and shown in Fink
added #2358 to improve Fink validation of error handling

inlang/source-code/sdk/src/createMessageLintReportsQuery.ts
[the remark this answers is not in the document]
jldec: @NiklasBuchfink and I tested manually on cal.com and other large projects.

The earlier removal of the debounce from this code _did_ result in UX problems when typing into Fink on larger projects with significant lint activity. A new debounce was added to the autosave on the Front-end in dabd05d. After this change the behavior appears to be better.

NOTE: The real fix for this will come with an API change for #2358 - the debounce is just a band-aid.

PR #2119, opened 2024-01-26
fix wrong error messages in paraglide [closes #2118]

### LINKED ISSUE #2118: [bug] "username password" is unsupported error message in paraglide js
### Problem

Paraglide JS logs an error message from lix saying "username and password is unsupported". 

image

### Expected behavior

1. If no auth is required, like in opening local repos, lix shouldn't initiate auth activity
2. no console logs in lix :) 

### Reproduction

NA

### Other information

From https://discord.com/channels/897438559458430986/1069566340031066172/1200200906734587914

janfjohannes:


janfjohannes:
@LorisSigrist   can you also double check that you do  not pass empty string to openRepo by accident if findRepoRoot does not find a git repo? i think i saw that happen somewhere but it will be a hard error in lix form now on, not silently accepted

janfjohannes:
@LorisSigrist  should @inlang/paraglide-js not be part of the affected packages?
Screenshot 2024-01-26 at 12 02 06

LorisSigrist:
It should be.
I believe `@lix-fs/client` is currently a `devDependency`, as it's bundled into paraglide. We need to manually add a changeset for paraglide.

PR #2128, opened 2024-01-26
CLI: Fix broken dependency

LorisSigrist:
With this pr `@inlang/telemetry` is no longer listed as a dependency in `@inlang/cli`


janfjohannes (approved):
just make sure to update lock file and rerun tests before merge

PR #2137, opened 2024-01-29
Paraglide Adapter Sveltekit: `dir` option & moe

LorisSigrist:
This PR adds a few QOL features to Paraglide & cleans up the configuration:
- Adds a `dir` option to `createI18n` where the text-direction can be configured.
- Adds default `%paraglide.lang%` and `%paraglide.dir%` string replacements for `app.html`
- Sets `lang` and `dir` on `event.locals.paraglide`, so the values can easily be used throughout the App
- Consolidates the config so that all config goes through `createI18n`

### DIFF
--- a/.changeset/famous-socks-leave.md
+++ b/.changeset/famous-socks-leave.md
@@ -1,0 +1,5 @@
+---
+"@inlang/paraglide-js-adapter-sveltekit": minor
+---
+
+breaking: move `noAlternateLinks` option from `<ParaglideJS>` component to `seo` option on `createI18n`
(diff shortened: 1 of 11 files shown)


samuelstroschein:
@LorisSigrist can you name `dir` `direction`? 

i thought it's a new `directory` option. don't using abbreviations makes things simpler to understand

LorisSigrist:
Good point!
`textDirection` would be the least ambiguous, so I'll go with that. I initially went with `dir` because that's the name of the HTML attribute.

PR #2152, opened 2024-01-30
Paraglide Stack Info

### LINKED ISSUE #2093: Paraglide get Stack info
It would be interesting to know which Frameworks Paraglide is being used with. We should gather some insights here

LorisSigrist:
Closes #2093 

This PR attempts to detect which tech-stack is being used & responds accordingly


inlang/source-code/paraglide/paraglide-js/src/compiler/compile.ts
```diff
@@ -25,8 +28,14 @@ export const compile = (args: {
...
-		event: "PARAGLIDE-JS compile executed",
+	getPackageJson(fs, process.cwd()).then((packageJson) => {
+		const stack = getStackInfo(packageJson)
+		telemetry.capture({
+			event: "PARAGLIDE-JS compile executed",
+			properties: {
+				stack,
+			},
+		})
 	})
```
samuelstroschein, line 39: How is ensured that the event will be captured? 

Shouldn't we change the compile function to be async?

inlang/source-code/paraglide/paraglide-js/src/services/telemetry/stack-detection.ts
```diff
@@ -0,0 +1,72 @@
...
+			if (dependencyName in allDependencies) {
+				const dependencyVersion = allDependencies[dependencyName]
+				if (typeof dependencyVersion !== "string") continue
+				packages[dependencyName] = dependencyVersion
+			}
+		}
+
+		return { packages }
+	} catch (error) {
+		console.error(`Failed to parse package.json`, error)
```
samuelstroschein, line 69: Use sentry or something. Logging this error to the user seems to be of no value

inlang/source-code/paraglide/paraglide-js/src/services/telemetry/stack-detection.ts
```diff
@@ -0,0 +1,72 @@
...
+export type StackInfo = {
+	/**
+	 * A map of relevant packages to their versions.
+	 */
+	packages: {
+		[packageName in RelevantPackage]?: string
+	}
+}
+
+export function getStackInfo(packageJson: unknown): StackInfo {
```
samuelstroschein, line 48: can you add a comment to the function what information is gathered?

inlang/source-code/paraglide/paraglide-js/src/services/telemetry/events.ts
```diff
@@ -1,12 +1,22 @@
...
-	| "PARAGLIDE-JS init finished"
+export type EventMap =
+	| {
+			event: "PARAGLIDE-JS compile executed"
+			properties: {
+				stack: StackInfo
+			}
+	  }
+	| {
+			event:
```
samuelstroschein, line 16: Putting props in the type here seems unnecessarly complex. 

- the value of typesafety goes mostly for the event name to avoid telemetry problems
- properties regulaly change for events. that's ok

I would revert this change and keep things simple.

samuelstroschein (commented):
I don't find an example event in posthog "development". Did you trigger a test event? 

To trigger a test event, you can use the `development-internal` env variables and run paraglide. it should automatically capture the event except of course the async function problem in `compile` leads to no capture

inlang/source-code/paraglide/paraglide-js/src/services/telemetry/stack-detection.ts
LorisSigrist: Good Point, I'll add a comment. You did guess the correct return type

I nested the packages. because we may want to add further information about the "stack". Having the `packages` be a property would mean we could add that without breaking existing queries in Posthog

inlang/source-code/paraglide/paraglide-js/src/compiler/compile.ts
```diff
@@ -17,16 +20,23 @@ const ignoreDirectory = `# ignore everything because the directory is auto-gener
...
-}): Record<string, string> => {
+}): Promise<Record<string, string>> => {
 	const compiledMessages = args.messages.map((message) =>
 		compileMessage(message, args.settings.languageTags, args.settings.sourceLanguageTag)
 	)
 
+	const pkgJson = await getPackageJson(fs, process.cwd())
+	const stack = getStackInfo(pkgJson)
+
+	//Don't wait for this, the request fires synchronously, we don't care about a response
```
samuelstroschein, line 34: the comment is misleading. nothing needs to be awaited here anyways :)

```suggestion
```

samuelstroschein (approved):

PR #2158, opened 2024-01-31
fix: communication and wrong links

NilsJacobsen:
- changed app to inlang app
- fixed wrong links
- removed the word tooling as it creates a new concept that is only used once

@samuelstroschein do you have some more hot fixes?


samuelstroschein:
@NilsJacobsen nope not for now.

PR #2165, opened 2024-02-01
Use Fully qualified URL in alternate links

### LINKED ISSUE #2164: [bug] Lighthouse: Document doesn't have a valid hreflang (Relative href value)
### Problem

image

### Expected behavior

Link for hreflang should not be a relative URL.

### Reproduction

Create a Svelte app using Paraglide so that an `hreflang` is injected to `<head>` post hydration like that:

```html
<link rel="alternate" hreflang="en" href="/photos">
```

### Other information

https://stackoverflow.com/questions/28291574/are-relative-links-valid-in-link-rel-alternate-hreflang-tags

LorisSigrist:
Closes #2164 

With this PR `<link rel="alternate">` now always has a fully qualified URL, as the spec demands


LorisSigrist:
@felixhaeberle do you have any idea why the IDE e2e tests keep failing here? I don't see how any of my changes are relevant, and CI runs just fine on main. Is this a permissions problem?

felixhaeberle:
This bugs me since yesterday! I don't know why they are failing. Seems random & an internal issue. The error also reads random.

Issue #1841, opened 2023-12-06
[bug] Message counter should to lazy linting messages


NiklasBuchfink:
### Problem

`messageCount` is calculated before messages are linted

![image](https://github.com/inlang/monorepo/assets/59048346/690de2b0-a39e-4ed0-9312-6efaac43251a)

### Expected behavior

Show the correct number of filtered messages
(The function `messageCount` should react to lazy linting)

### Reproduction

Go to https://inlang.com/editor/github.com/AppFlowy-IO/AppFlowy?lint=messageLintRule.inlang.missingTranslation

NiklasBuchfink:
The solution to this was to update one global counter if the show state of a single message changes.

Issue #1127, opened 2023-07-18
treat "isolated view" as a regular filter option


ivanhofer:
![image](https://github.com/inlang/inlang/assets/21335119/6f5a939d-2ed8-43c3-978b-a3df8dedb445)

We kind of render two filter states in the UI. The "Isolated view" state is rendered different than any other filter.
In my opinion, this should be treated like any other filter and it should be rendered after the "Language" filter on the same line.

NilsJacobsen:
@inlang/editor

felixhaeberle:
although this button looks nice, the proposal makes sense 😄

NiklasBuchfink:
I agree, we will find a suitable place for it in the filter row 👍

felixhaeberle:
@NiklasBuchfink Maybe also related #1380

NiklasBuchfink:
@felixhaeberle It's unrelated
This is about how we display this filter state to the user and that it should be part of our filter section.

The mentioned issue is about reacting to all filtering options (lint, language, id, search) with our lintSummary calculation. We removed this feature, as it resulted in lags.

PR #2176, opened 2024-02-05
[skip preview] setup action test

NiklasBuchfink:
General:
- [x] insert GITHUB_TOKEN via env
- [x] scope to PRs and file changes

Styling:
- [x] comment styling (markdown)
- [x] list missing messages

Test:
- [x] create comment
- [x] update comment
- [ ] release action
- [ ] adding to a new repo

Content:
- [ ] Branding
- [ ] Create a product page
- [ ] Extend Automation guide
- [ ] Update README


samuelstroschein:
@NiklasBuchfink is the diffing so complicated? the github action provides better value if it diffs and only triggers if new lint reports come in. 

@janfjohannes interim workaround to trigger the github action when a project file changes substantially increases the barrier for adoption and risk for breaking the github action. If someone renamed an inlang project, the message.json file, adds a new project or whatever, the github action is broken.

samuelstroschein:
seeing this message https://github.com/opral/monorepo/pull/2176#issuecomment-1952860075 increases the desire for diffing even more. yes the project file changed but this PR introduces no message updates. Hence, the message provides no value

NiklasBuchfink:
> The view needs to be split by lint rule.

Ok, I can change that. I was not sure if this should be listed by errors/warnings or lint rules id

felixhaeberle:
I would also suggest to:

- get rid of the warning ⚠️ icon, we have errors & warnings & in the future other lint rules. the best would be (also in regard to change control): 🛎️ Update translations

- give multiple links for each lint rule (scales better and lint rules can eventually be linked = what does this lint rule do?)

- naming fink prominently without explaining what the user can do hurts conversion: I would go for contribute (via Fink) as @samuelstroschein has proposed

NiklasBuchfink:
> @NiklasBuchfink is the diffing so complicated?

@janfjohannes Therefore these two need to work:
1. Checkout a different branch (e.g. PR branch `github-lint-action` -> `main`)
2. Checkout a different repos (e.g. PR fork `NiklasBuchfink/monorepo` -> `opral/monorepo`)
Can you give me an ETA for this?

@samuelstroschein Would this approach be ok if we don't use the `paths` and instead:
1. use project path which is an input anyway
2. get the store location of the language files from the SDK (needs to be exposed) -> check if they are in the changedFiles of the PR (simple filter of file names)

janfjohannes:
@NiklasBuchfink   if the paths filter is not an option and you cannot wait for lix to fix brnach switching, you could also switch branch with command line git,  then open the repo again with lix and compare the new output with the saved one.

NiklasBuchfink:
> @NiklasBuchfink if the paths filter is not an option and you cannot wait for lix to fix brnach switching, you could also switch branch with command line git, then open the repo again with lix and compare the new output with the saved one.

@janfjohannes Can I execute the Git commands from the Typescript file similar to the CLI?

samuelstroschein:
> @NiklasBuchfink  Can I execute the Git commands from the Typescript file similar to the CLI?

Yes, you can. Search for "node subprocess" or "node execute" 

@janfjohannes good workaround idea to call the git CLI!

NiklasBuchfink:
> Yes, you can. Search for "node subprocess" or "node execute"

Ok, I will try this tomorrow 👍

NiklasBuchfink:
### Quick Update:
- [x] comment is updated
- [x] diff works for branches using Git CLI
- [ ] diff works for forks using Git CLI (to be tested)
- [ ] multiple projects (comment must include project path/id to update the correct comment)

samuelstroschein:
>  multiple projects (comment must include project path/id to update the correct comment)

Having the comment point to the project is a good idea! That avoids config for the user.

NiklasBuchfink:
The last issue is the following:
If a PR comes from a fork, the permissions of the GITHUB_TOKEN are set to read only. This way I cannot create a comment. This is a [known issue](https://stackoverflow.com/questions/76952023/how-to-make-github-actions-safely-access-secrets-for-prs-created-from-forks).

Everything else works now.

samuelstroschein:
@NiklasBuchfink let's ignore forks then? I assume commenting still work for pull requests opened on a fork to the upstream repo?

NiklasBuchfink:
@samuelstroschein The check works now (See second action PR from NiklasBuchfink/monorepo-copy), but the commenting is the issue. We need the GitHub token to comment. For the CLA workflow it works somehow with the same token. 🤔

NiklasBuchfink:
**Update**
- [x] commenting works for PRs from forks and branches
- [x] automatically checks all projects without the need for an input
- [x] lists results of all projects in one comment
- [x] debug logs in action in case some of the projects have errors (Skip project ... because of errors)
- [x] diffing if a project went from "no errors" to "some errors" and provide an error message:
image

NiklasBuchfink:
PR from fork: https://github.com/opral/monorepo/pull/2291
PR from branch: https://github.com/opral/monorepo/pull/2297

NiklasBuchfink:
Related issues to make this even better:
- https://github.com/opral/monorepo/issues/1127 (required to only show relevant messages/ filter multiple IDs)
- https://github.com/opral/monorepo/issues/1841 (show the correct number of messages)

samuelstroschein:
@NiklasBuchfink should I review this? The PR is marked as draft (not ready for review)

NiklasBuchfink:
@samuelstroschein I will implement the proposed changes before you review this.

I would argue for not listing all project errors in the comment, as it would get very messy with 6 errors like this:
image

Maybe I can just show the error type and a link to Fink where the details can be viewed. What do you think?

samuelstroschein:
@NiklasBuchfink have you considered showing all errors but behind an accordion https://gist.github.com/pierrejoubert73/902cc94d79424356a8d20be2b382e1ab ?

NiklasBuchfink:
@samuelstroschein Good idea, I can do it that way 👍

module.js
```diff
@@ -0,0 +1,35 @@
+// mathUtils.js
+
```
janfjohannes, line 2: whats this file?

NiklasBuchfink: This was a test module during the module import problem. I forgot to delete it 👍

inlang/source-code/github-lint-action/src/main.ts
```diff
@@ -1,97 +1,388 @@
...
-			deletions: 0,
-			changes: 0,
+		// Prepare head repo
+		let repoHead
+		if (isFork) {
+			core.debug("Fork detected, cloning base repository")
+			process.chdir("../../../")
+			await cloneRepository(headMeta)
+			process.chdir(headMeta.repo)
+			repoHead = await openRepository(process.cwd(), {
```
janfjohannes, line 86: see above

inlang/source-code/github-lint-action/src/main.ts
```diff
@@ -1,97 +1,388 @@
...
+		core.debug(`Is fork: ${isFork}`)
 
-		let diffData = {
-			additions: 0,
-			deletions: 0,
-			changes: 0,
+		// Prepare head repo
+		let repoHead
+		if (isFork) {
+			core.debug("Fork detected, cloning base repository")
```
janfjohannes, line 82: why would we handle forks here?

inlang/source-code/github-lint-action/src/main.ts
```diff
@@ -1,97 +1,388 @@
...
+			process.chdir("../../../")
+			await cloneRepository(headMeta)
+			process.chdir(headMeta.repo)
+			repoHead = await openRepository(process.cwd(), {
+				nodeishFs: fs,
+			})
+		} else {
+			core.debug("Fork not detected, fetching and checking out base repository")
```
janfjohannes, line 90: not sure why this is needed aat all, the pr will run on the branch after the merge simulated allrady by default in github.

inlang/source-code/github-lint-action/src/main.ts
```diff
@@ -1,97 +1,388 @@
...
+			await fetchBranch(headMeta.branch)
+			await checkoutBranch(headMeta.branch)
+			await pull()
+			repoHead = await openRepository(process.cwd(), {
```
janfjohannes, line 94: see above

inlang/source-code/github-lint-action/src/main.ts
```diff
@@ -1,97 +1,388 @@
...
+			}
+			core.debug(`stdout: ${stdout}`)
+			core.debug(`stderr: ${stderr}`)
+			resolve()
+		})
+	})
+}
+
+// Function to clone the base repository
+async function cloneRepository(repoData: { link: string; branch: string }) {
```
janfjohannes, line 370: not sure why needed, this is allready in the yaml as one of the steps before this code runs, no?

janfjohannes (changes requested):

inlang/source-code/github-lint-action/src/index.test.ts
```diff
@@ -1,6 +1,102 @@
...
+					},
+				],
+			},
+		],
+	},
+]
+
+async function setupProject(enabledLintRule?: MessageLintRule) {
+	const repo = await mockRepo()
+	await fs.mkdir("project.inlang", { recursive: true })
```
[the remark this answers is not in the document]
NiklasBuchfink: The test doesn't work at all. I must split up the action file and write tests for all functions. I can't run the whole action at once, as it relies heavily on GitHub APIs. So, testing is missing completely atm.

inlang/source-code/github-lint-action/src/main.ts
```diff
@@ -1,97 +1,388 @@
...
+		const token = process.env.GITHUB_TOKEN
+		if (!token) {
+			throw new Error("GITHUB_TOKEN is not set")
+		}
+		const { owner, repo } = github.context.repo
+		const prNumber = github.context.payload.pull_request?.number
 
-		// Debug logs are only output if the `ACTIONS_STEP_DEBUG` secret is true
-		core.debug(`I got all inputs: ${owner} ${repo} ${pr_number} ${token}`)
+		const repoBase = await openRepository(process.cwd(), {
```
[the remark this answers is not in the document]
NiklasBuchfink: Done ✅

inlang/source-code/github-lint-action/src/index.test.ts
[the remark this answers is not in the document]
NiklasBuchfink: I deleted the test file as it makes no sense to have something broken in there.

inlang/source-code/github-lint-action/src/index.test.ts
```diff
@@ -1,6 +1,102 @@
...
+	const _import = async (name: string) => {
+		if (name === "lintRule.js") {
+			return { default: enabledLintRule as MessageLintRule } satisfies InlangModule
+		}
+		return {
+			default: _mockPlugin,
+		}
+	}
+
+	return await loadProject({
```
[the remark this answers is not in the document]
NiklasBuchfink: see above

inlang/source-code/github-lint-action/src/main.ts
NiklasBuchfink: The debug message was irritating. We check out the base repo first and need the head repo of a fork or the head branch to compare both. I updated the comment and we talked about it ✅

PR #2188, opened 2024-02-07
Cross-Sell Fink & Parrot from Paraglide Docs

LorisSigrist:
This PR adds a "Working with Translators" section to the Paraglide docs where we link to Fink & Parrot


NiklasBuchfink (changes requested):
The wording could scare developers. Instead of describing that everything is pushed and deployed directly, "commit to a translation branch or submit changes via pull request" would be more attractive.

NiklasBuchfink (approved):

PR #2196, opened 2024-02-08
Lazy checkout with batching

janfjohannes:


janfjohannes:
superseded by lazy clone PR

PR #2197, opened 2024-02-08
chore: remove duplicate command

samuelstroschein:


inlang/source-code/ide-extension/package.json
```diff
@@ -152,8 +152,7 @@
 		"dev": "DEV=true node ./build.js",
 		"package": "pnpm vsce package --no-dependencies",
 		"publish": "pnpm vsce publish --no-dependencies",
-		"vscode:prepublish": "npm run esbuild-base -- --minify",
-		"esbuild-base": "node ./build.js",
```
samuelstroschein, line 156: The esbuild-base command should be redundant given that it just calls the build script. Minify should also be redundant because the build command minifies already

PR #2202, opened 2024-02-08
fix: assign string to _languageTag instead of function

leonardsimonse:
Svelte example shows a type error when converting to TS.

`Type 'string' is not assignable to type '() => "bg" | "cs" | "de" | "en" | "es" | "fr" | "hu" | "pl" | "ro" | "sk" | "tr" | "uk"'.`

Assigning the string value to _languageTag, instead of the function itself, solves this issue.


leonardsimonse:

LorisSigrist:
Good catch!

PR #2226, opened 2024-02-15
Update ide docs

jannesblobel:
I updated the getting started section in the docs and changed the URL on the `create a project` button in the inlang tab
CleanShot 2024-02-15 at 15 43 17@2x

### DIFF
--- a/.changeset/dirty-gorillas-agree.md
+++ b/.changeset/dirty-gorillas-agree.md
@@ -1,0 +1,5 @@
+---
+"vs-code-extension": patch
+---
+
+Changed the hreft from the 'create project’ button in the IDE extension and update docs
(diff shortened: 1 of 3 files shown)


felixhaeberle:
@jannesblobel This getting started guide is party too unspecific and misleading.
1. we don't want people to create settings.json manually
2. people don't know the content of the settings.json
3. 3, 4, 5 are not required - but recommended.

cc @samuelstroschein how would you improve the getting started docs of the vscode extension for now despite we do not have `createNewProject` yet?

jannesblobel:
Points 3 and 4 are necessary, otherwise kebab would not work.

samuelstroschein:
@felixhaeberle i don't know. jannes approach seems to be the best of the worst options we have. the alternative is to have a create project functionality in app that "manually" creates a project for a user. if we do this, we could already ship `createNewProject()` though that will use the plugin json to store messages. The path to use existing messages, i18next or next-intl would be blocked with this approach until #1585 ships. 

I am unsure how to move forward here.

samuelstroschein:
@jannesblobel point 3 syntax matcher is also optional

jannesblobel:
> @jannesblobel point 3 syntax matcher is also optional

From a technical perspective, maybe. But who will use Kebab without a matcher?

PR #2252, opened 2024-02-17
remove fetch-env:public

jldec:
internal [discussion](https://discord.com/channels/897438559458430986/1130765707639001088/1208493531149369364)


package.json
```diff
@@ -17,7 +17,6 @@
 		"build:no-nx": "pnpm --stream --parallel build",
 		"dev:no-nx": "pnpm --stream --parallel dev",
 		"dev:no-clear": "pnpm dev:no-nx | sed -u 's/\\x1Bc//g'",
-		"fetch-env:public": "pnpm --filter=@inlang/env-variables fetch-env:public",
 		"fetch-env:internal": "doppler secrets download --no-file --format env > .env",
```
samuelstroschein, line 20: this should be removed too if we go for `doppler run --` everywhere, or?

samuelstroschein:
- [ ] remove https://github.com/opral/monorepo/blob/0a2ffcc52f2f39785c54066a55d5a3f67f1c1576/inlang/source-code/env-variables/src/build/fetchPublicEnv.ts#L12-L55
- [ ] remove https://github.com/opral/monorepo/blob/0a2ffcc52f2f39785c54066a55d5a3f67f1c1576/inlang/source-code/env-variables/src/build/buildStepVariables.ts#L4-L5
- [ ] remove https://github.com/opral/monorepo/blob/0a2ffcc52f2f39785c54066a55d5a3f67f1c1576/inlang/source-code/env-variables/src/build/buildStepVariables.ts#L24-L25
- [ ] dotenv can likely be removed because env variables come via process.env and never as a file now https://github.com/opral/monorepo/blob/0a2ffcc52f2f39785c54066a55d5a3f67f1c1576/inlang/source-code/env-variables/build.ts#L10-L11

jldec:
How about merging this as-is today (without the additional changes in the env-variables package) and follow up with the env-variables tightening in a separate PR.

samuelstroschein (approved):
okay lets merge this and in a follow up PR remove the entire env var package

PR #2299, opened 2024-02-27
Improve m function matcher & test coverage

### LINKED ISSUE #2265: [bug] VSCode locale previews break when passing function references
### Problem

When passing the locales' function references to valibot schemas instead of the results of the functions, previews break.

https://github.com/opral/monorepo/assets/18720666/c46e4c3d-4aba-4c27-851d-7c03576f295b

### Expected behavior

Previews should show in any case, even when the reference is used instead of the output

### Reproduction

# Paste this code into a new .js/.ts file:

```js
import * as v from 'valibot';

import * as m from '$paraglide/messages';

export const testSchema1 = v.object({
	public_id: v.string([v.minLength(1, m.affect_micro_pets)]),
	start_date: v.optional(v.date(m.dramatic_height_honey), () => new Date()),
	end_date: v.optional(v.date(m.barbados_maintenance_referenced), () => new Date())
});

export const testSchema2 = v.object({
	public_id: v.string([v.uuid()]),
	team_id: v.string([v.uuid()]),
	email: v.string([v.minLength(1, m.patches_fit_means), v.email(m.thats_written_regulatory)]),
	first_name: v.string([v.minLength(2, m.yrs_rack_hobby)]),
	last_name: v.string([v.minLength(2, m.certificate_editions_platforms)])
});

export const testSchema3 = v.object({
	public_id: v.string([v.minLength(1, m.affect_micro_pets)]),
	team_name: v.string([v.minLength(1, m.coral_gets_damage)]),
	team_description: v.optional(v.string([v.minLength(1, m.application_talk_fcc)]))
});
```

### Other information

_No response_

felixhaeberle:
fix #2265

### DIFF
--- a/.changeset/empty-windows-clap.md
+++ b/.changeset/empty-windows-clap.md
@@ -1,0 +1,5 @@
+---
+"@inlang/plugin-m-function-matcher": minor
+---
+
+improve matcher to include function referencing & improve test coverage
--- a/inlang/source-code/plugins/m-function-matcher/src/ideExtension/messageReferenceMatchers.test.ts
+++ b/inlang/source-code/plugins/m-function-matcher/src/ideExtension/messageReferenceMatchers.test.ts
@@ -303,6 +303,49 @@ describe("Paraglide Message Parser", () => {
 		])
 	})
 
+	it("should parse function calls without parentheses", () => {
+		const sourceCode = `import * as m from 'module';
+
+		const a = {
+			b: m.helloWorld,
+			c: m.helloWorld(),
+			d: m.helloWorld().someFunction(),
+			e: m.this_is_a_message123,
+		}
+		`
+		const result = parse(sourceCode)
+		expect(result).toEqual([
+			{
+				messageId: "helloWorld",
+				position: {
+					start: { line: 4, character: 9 },
+					end: { line: 4, character: 19 },
+				},
+			},
+			{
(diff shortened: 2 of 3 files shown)


LorisSigrist (approved):
The test with the hardcode line/column numbers looks a little fragile to me, but since it's unlikely to be edited often it's probably fine.

Great work!

felixhaeberle:
> The test with the hardcode line/column numbers looks a little fragile to me

Where do you see the fragility?

LorisSigrist:
> > The test with the hardcode line/column numbers looks a little fragile to me
> 
> Where do you see the fragility?

The hardcoded line numbers & columns. Any change to the formatting of the test string would mess it up. 
I would've given each message a unique name & string-searched to get the expected line & column dynamically

felixhaeberle:
😅 I do not understand this as I want to test these exact positions and not get them dynamically.
Would retrieving them dynamically increase the possibility of wrong dynamic positions – How would this look like?

LorisSigrist:
I meant to have some sort of helper-function that returns the start & end coords of a given string & then using that. 
```
const { start, end } = findString(sourceCode, "firstMessage()") //The string it's supposed to match (including parens if needed)
```

If you give each of the four messages an easily identifiable, unique name then you wouldn't need to hardcode it.

PR #2306, opened 2024-02-28
Prompt user for Tech Stack when running `paraglide-js init`

LorisSigrist:
As described in #1570

This PR adds a prompt for the user's tech stack to `paraglide-js init`. Depending on which one they select it will recommend one of the Adapters & link to the documentation


felixhaeberle:
this is really dope! 😻

PR #2325, opened 2024-02-29
Allow additionalProperties on inlang-message-format and project-settings

jldec:
This change should allow new properties to be added to the schemas for inlang-message-format and project-settings without breaking existing apps which are unaware of the new properties.

discussion in https://discord.com/channels/897438559458430986/1212819302219120640

I did a couple of manual chacks with this PR, and changes appear to work as intended:

Screenshot 2024-02-29 at 22 15 38
Screenshot 2024-02-29 at 22 15 19

I think we can ignore the json schema warning in the vscode editor. I assume that's based on the schema referenced by the settings file.

### DIFF
--- a/inlang/source-code/plugins/inlang-message-format/src/storageSchema.ts
+++ b/inlang/source-code/plugins/inlang-message-format/src/storageSchema.ts
@@ -4,7 +4,7 @@ const InternalProperties = Type.Object(
 	{
 		$schema: Type.Optional(Type.Literal("https://inlang.com/schema/inlang-message-format")),
 	},
-	{ additionalProperties: false }
+	{ additionalProperties: true }
 )
 
 const Messages = Type.Record(
@@ -15,7 +15,7 @@ const Messages = Type.Record(
 		examples: ["helloWorld", "hello_world", "helloWorld123", "hello_world_123"],
 	}),
 	Type.String(),
-	{ additionalProperties: false }
+	{ additionalProperties: true }
 )
 
 /**
--- a/.changeset/four-pianos-hang.md
+++ b/.changeset/four-pianos-hang.md
@@ -1,0 +1,5 @@
+---
+"@inlang/paraglide-js": patch
+---
+
+bump `@inlang/sdk` dependency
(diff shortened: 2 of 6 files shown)


inlang/source-code/plugins/inlang-message-format/src/storageSchema.ts
```diff
@@ -4,7 +4,7 @@ const InternalProperties = Type.Object(
 	{
 		$schema: Type.Optional(Type.Literal("https://inlang.com/schema/inlang-message-format")),
 	},
-	{ additionalProperties: false }
+	{ additionalProperties: true }
```
jldec, line 7: I was unsure if I should remove the whole object or just the `additionalProperties` property leaving `{ }`, so it was easier to change the value to `true`.

jldec:
working on fixing the tests

samuelstroschein (approved):
> @jldec  I think we can ignore the json schema warning in the vscode editor. I assume that's based on the schema referenced by the settings file.

Yes, we can. The schema is deployed from `main`. Hence, the schema change here is not reflected in the lint.

LorisSigrist:
Probably because Paraglide is bundled and has the SDK as a `devDependency`. I'll take the liberty to add a changeset to this PR

jldec:
@LorisSigrist it sounds like you intentionally do not include the sdk package as a dependency in paraglide-js
to reduce the weight of paraglide when it is added as a dependency of "host" UI projects.

@samuelstroschein it appears that scenarios like this force us to break the automatic dependency tracking provided by changeset, which doesn't support tracking devDependencies [issue/pr](https://github.com/changesets/changesets/pull/1159).

jldec:
here's the versioning PR produced by changesets 
https://github.com/opral/monorepo/pull/2329

samuelstroschein:
@jldec i know. we ran into the versioning issue multiple times with changesets now. it's unmaintained. setting something up like https://rushjs.io/ takes too much effort atm. another changelog creator that is as smooth as changesets doesn't seem to exist :/ 

so either we live with the versioning issues, find another versioning system, or make a switch to rush. i vote for live with it

PR #2335, opened 2024-03-02
Simplify sdk loadSettings

### LINKED ISSUE #2328: Errors for loadProject settings get swallowed
### Problem

@jldec   still not sure at what place the swallowing happens either sdk somewhere above like cli. it happens definitely in the website build.

 when read errors occur in the screenshot. the raw error is not loogged anywhere but a generic error reason instead which makes finding bugs from ci output impossible.
 
 
Screenshot 2024-03-01 at 14 05 54

### Expected behavior

i see the raw error in the log

### Reproduction

remove project.json and run build for the website.  

### Other information

_No response_

jldec:
Fixes #2328 

@janfjohannes could you try this with your Lix setup.
Errors should now be relayed back directly, without wrapping. 

The tryCatch and extra error wrappers in the loadProject code were not adding much value, so I removed them.
 
This PR includes the paraglide VS Code debug configuration, which I used to debug the flow.
It should be useful for others.

### DIFF
--- a/inlang/source-code/sdk/src/loadProject.ts
+++ b/inlang/source-code/sdk/src/loadProject.ts
@@ -8,6 +8,7 @@ import type {
 import { type ImportFunction, resolveModules } from "./resolve-modules/index.js"
 import { TypeCompiler, ValueErrorType } from "@sinclair/typebox/compiler"
 import {
+	ProjectSettingsFileJSONSyntaxError,
 	ProjectSettingsInvalidError,
 	PluginLoadMessagesError,
 	PluginSaveMessagesError,
@@ -327,9 +328,19 @@ const loadSettings = async (args: {
 	settingsFilePath: string
 	nodeishFs: NodeishFilesystemSubset
 }) => {
-	// caller should handle errors
+	// allow readFile errors to pass through to help with debugging
 	const settingsFile = await args.nodeishFs.readFile(args.settingsFilePath, { encoding: "utf-8" })
-	const json = JSON.parse(settingsFile)
+
+	let json: unknown
+	try {
+		json = JSON.parse(settingsFile)
+	} catch (err) {
+		// trap JSON parse errors and help users by including path to settings file
+		throw new ProjectSettingsFileJSONSyntaxError({
+			cause: err,
+			path: args.settingsFilePath,
+		})
+	}
 	return parseSettings(json)
 }
 
--- a/inlang/source-code/sdk/src/loadProject.ts
+++ b/inlang/source-code/sdk/src/loadProject.ts
@@ -8,8 +8,6 @@ import type {
 import { type ImportFunction, resolveModules } from "./resolve-modules/index.js"
 import { TypeCompiler, ValueErrorType } from "@sinclair/typebox/compiler"
 import {
-	ProjectSettingsFileJSONSyntaxError,
-	ProjectSettingsFileNotFoundError,
 	ProjectSettingsInvalidError,
 	PluginLoadMessagesError,
 	PluginSaveMessagesError,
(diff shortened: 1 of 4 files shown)


inlang/source-code/sdk/src/loadProject.ts
```diff
@@ -329,24 +327,10 @@ const loadSettings = async (args: {
...
-		throw new ProjectSettingsFileJSONSyntaxError({
-			cause: json.error,
-			path: args.settingsFilePath,
-		})
-	}
-	return parseSettings(json.data)
+	// caller should handle errors
+	const settingsFile = await args.nodeishFs.readFile(args.settingsFilePath, { encoding: "utf-8" })
+	const json = JSON.parse(settingsFile)
+	return parseSettings(json)
```
samuelstroschein, line 333: The UX of errors seems to be negatively impacted by this change. 

1. an app won't inform the user anymore that a settings file has not been found
2. an app won't inform the user anymore that the settings file contains a syntax error

Both errors were introduced after inlang users ran into them and did not understand the underlying error. Hence, we wrapped them to provide a better error message. 

Please elaborate on why this change is beneficial/required.

samuelstroschein (changes requested):

inlang/source-code/sdk/src/loadProject.ts
jldec: ok - I reverted the JSON parse error so that it includes the settings file path in the error message. You're right, that's helpful for UX.

I think we should allow the readfile error to pass through to users, because the wrapper doesn't add additional information, and based on @janfjohannes experience last week, the wrapper can actually prevent lix failures from being reported accurately. 

before:
```
$ p inlang validate --project ./project.inlang/
ℹ 🔎 Validating the inlang project...                                                                                                  12:05:53 PM

 ERROR  The file at "/Users/jldec/inlang/load-test/project.inlang/settings.json" could not be read. Does the file exists?               12:05:53 PM

  at loadSettings (node_modules/.pnpm/@inlang+cli@2.12.6/node_modules/@inlang/cli/dist/main.js:98230:11)
```

with this PR
```
$ p inlang validate --project project.inlang/
ℹ 🔎 Validating the inlang project...                                                                                                  12:44:00 PM

[12:44:01 PM]  ERROR  ENOENT: no such file or directory, open '/Users/jldec/inlang/monorepo/inlang/source-code/paraglide/paraglide-js/example/project.inlang/settings.json'
```

samuelstroschein: @jldec feel free to call me

samuelstroschein: The CLI is not logging `error.cause`. That's a bug in the CLI. There seems little benefit in removing concrete errors in favor of generic one's just because one app is not logging correctly. 

`error.cause` has been introduced specifically to re-throw with more concrete error messages https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause

inlang/source-code/sdk/src/loadProject.ts
```diff
@@ -329,24 +328,20 @@ const loadSettings = async (args: {
...
-		throw new ProjectSettingsFileNotFoundError({
-			cause: settingsFileError,
-			path: args.settingsFilePath,
-		})
-
-	const json = tryCatch(() => JSON.parse(settingsFile!))
-
-	if (json.error) {
+	// allow readFile errors to pass through to help with debugging
+	const settingsFile = await args.nodeishFs.readFile(args.settingsFilePath, { encoding: "utf-8" })
```
samuelstroschein, line 332: This will throw without a concise warning for the user. "ENOENT" is not an error message users can't solve/know what to do

jldec:
Closing this - will re-visit as we continue to improve the sdk api.

PR #2376, opened 2024-03-12
Plural prototype

LorisSigrist:
This is a prototype for using plurals in paraglide. This PR will likely never be merged. It exists to allow us to play around & figure out requirements. 

### Message Format changes
We can now have a match-statement in our messages to select between different variants. The source
```
match {gender} when female {Hello actress} when * {Hello actor}
```
would compile into the message:
```ts
{
  id: "message_id",
  alias: {},
  // we will need to introduce typed variable references, so we can have more than just exact matches
  selectors: [
    {
      type: "VariableReference",
      name: "gender",
    },
  ],
  variants: [
    //each variant must have exactly one `match` value for each selector
    { match: ["female"], languageTag: "en", pattern: [{ type: "Text", value: "Hello actress" }] },
    { match: ["*"], languageTag: "en", pattern: [{ type: "Text", value: "Hello actor" }] },
  ],
}
```

(this is still very fragile, so changing whitespace may break the parser. I just bodged together the simplest thing that works)

### DIFF
--- a/inlang/source-code/paraglide/paraglide-js/example/e2e.test.ts
+++ b/inlang/source-code/paraglide/paraglide-js/example/e2e.test.ts
@@ -16,4 +16,8 @@ test("should log the messages in the correct language tag", async () => {
 
 	//Test fallback
 	expect(consoleMock.mock.calls[5][0]).toBe("This is missing in German")
+
+	//Test plural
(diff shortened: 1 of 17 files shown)


LorisSigrist:
Closing this - We learned everything we need to know

Issue #2327, opened 2024-03-01
High volume of lint report events consume excessive resources and lead to performance issues


jldec:
Capturing this as a (slightly extreme) example of the reactive api for message lint reports.

In the test scenario, 1000 messages were added directly to the file system, and then translated with the inlang cli.

`messageLintReports.getAll.subscribe()` triggers 2000 lint report events, one for each message added, and one for each message translated. 

Each of those events contains all lint reports for all messages. In the test case there were 36 languages so that means up to 36000 lint reports per event. 😳 

The process takes about a minute with (M1) cpu maxed and many gc's observed. 

![lint-reports-screenshot](https://github.com/opral/monorepo/assets/849592/a63440b0-5047-42ba-8f0a-f633e573b80f)

jldec:
from [internal](https://discord.com/channels/897438559458430986/1212736106077102131)
> TL;DR an api change is required. The existing messageLintReports.getAll.subscribe() api is designed to produce one event on every message change, and each of those events includes all lint reports for all messages. We won't tackle that in PR 2108.

jldec:
and more details from this [internal discord](https://discord.com/channels/897438559458430986/1212192970996981830/1212732489248804955) on feb 29.

> I took another look and ran some experiments with throttling 
>
> bottom line - there are exactly 2000 lint report events in the test because messageLintReports.getAll.subscribe()  produces an event on every message change, and each of those events includes all lint reports for all messages. 
>
> Throttling would mean changing the api not to send an event per message update, or not to send the full list of reports per event.
>
> iow there is no throttling  opportunity left which does not alter the api semantics.
Image

Issue #2230, opened 2024-02-15
logger option for the sdk


jldec:
### Problem 
It would be helpful to have logs that can be turned on and off for debugging e.g. to better understand internal behavior in the sdk. 

### Proposal
__(thanks @samuelstroschein)__
Add a logger to the sdk that can be configured with a loadProject() option like `{ debug: true }`
The option could also be exposed via a  --debug flag on the CLI

samuelstroschein:
@jldec assigning you to make sure this issue will be resolved

PR #2378, opened 2024-03-12
MessageLintReports get() and getAll() async, with no reactivity, to reduce excessive resource consumption

### LINKED ISSUE #2073: lint report query API should be async
## Problem 

Lint reports are async but exposed as sync API. Similar to #1846 

The result is hacky workarounds: 

https://github.com/opral/monorepo/blob/f5f20812f19e3b9399ee942faf031db4f5e3cd99/inlang/source-code/badge/src/badge.ts#L44-L78

## Proposal

Expose the lint report API as async to make workarounds redundant.

jldec:
partial fix for #2327 (memory consumption still limits scale)
resolves #2073

https://www.loom.com/share/3cf3c51d2f644886948c1267fa73f176

### remove reactive map and make lint reports api async

MessageLintReportsQueryApi is simplified to make it non-reactive (no subscriptions), and the lint-report-specific ReactiveMap is replaced with a vanilla Map. (This change replaces the additional includedMessageIds api from the first experiment.)

**Reasoning 💡**  
Lint reports don't require their own separate reactive map over all messages. It should be possible to maintain an up-to-date model of lint reports in memory, and build reactive (auto-updating) UIs with lint reports, both using only the messages reactive interface to watch for changes.

This would the first step toward removing granular (per message) reactivity from the sdk core.

#### status:
- [x] **async** lint reports get() and getAll() 
- [x] remove lint reports reactivity
- [x] load-test changes 
  - [x] find scalability limit at 4GB node heap size:  
  limit = ~10000 messages with translation into 36 languages
- [x] fix tests
- [x] code changes in fink and sherlock to test UX improvement at feature parity
  - [x] fix off-by-one total lint errors value in fink
  - [x] fix empty message list on reload with lint reports filter  
- [x] changesets

### first experiment - new api `IncludedMessageIds` (abandoned)

First attempt to fix the issue replaced subscription to `getAll()` with a subscription to `includedMessageIds`. This new api returns just the message IDs which have lint reports, instead of returning all lint reports in every event, making it much lighter. The change was implemented in d7ae5c118 **but does not fully resolve the issue**.

https://www.loom.com/share/358ae27b6dda403aa4c013a2ac0b723b
The loom show the out of memory failure when translating 4500 messages with 
`project.query.messageLintReports.includedMessageIds.subscribe` in the load test.


package.json
```diff
@@ -8,7 +8,7 @@
 		"production": "pnpm exec nx run-many --nx-bail --target=production --parallel 1000 --exclude @lix-js/server --exclude @inlang/badge --exclude @inlang/manage --exclude @inlang/editor --exclude @inlang/telemetry-proxy --output-style=stream",
 		"build": "pnpm exec nx run-many --nx-bail --target=build --parallel",
 		"dev:lix": "pnpm --filter @inlang/env-variables build && pnpm --stream --parallel --filter @lix-js/* --filter @inlang/server dev",
-		"dev:lixe-fink": "pnpm --filter @inlang/env-variables build && pnpm --stream --parallel --filter @lix-js/* --filter @inlang/serve --filter @inlang/manage --filter @inlang/editor --filter @inlang/website dev",
+		"dev:lix-fink": "MOCK_TRANSLATE=true pnpm --filter @inlang/env-variables build && pnpm --stream --parallel --filter @lix-js/* --filter @inlang/server --filter @inlang/manage --filter @inlang/editor --filter @inlang/sdk --filter @inlang/website --filter @inlang/telemetry-proxy dev",
```
NiklasBuchfink, line 11: If you just want to run Fink with this command, you can leave out the server and the website.

NiklasBuchfink (changes requested):
I found a bug when loading the Fink with a lint rule filter selected:
https://www.loom.com/share/84bc0df455bf45f48435a6f2f6a38ac4

It is not something SDK-related and more about Finks application state, so let me know, if I should fix this.

felixhaeberle (approved):
lgtm 👍

package.json
jldec: The MOCK_TRANSLATE env var at the start actually doesn't work as intended so i'll remove that.
The rest is useful for sdk and lix development so I'll keep those for now.

inlang/source-code/editor/src/pages/@host/@owner/@repository/Message.tsx
```diff
@@ -33,34 +32,37 @@ export function Message(props: { id: string }) {
 		}
 	})
 
+	// TODO: Find out why messages.get() in solidAdapter only works with subscribe.
+	//       By contrast, query.messages.getAll() signal does work. see ListHeader.tsx
 	createEffect(() => {
 		if (!project.loading) {
-			project()!.query.messages.get.subscribe({ where: { id: props.id } }, (message) =>
+			// console.log("Editor subscribing to message udpates for", props.id)
```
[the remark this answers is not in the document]
jldec: I'll remove those - need to find better debug logging solution - #2230 and MESDK-21

jldec:
> It is not something SDK-related and more about Finks application state, so let me know, if I should fix this.

@felixhaeberle I have pushed a fix for this
(surprisingly) the signal produced by createResource only updates itself on the tick after the awaited fetch resolves, so I had to workaround this.

PR #2397, opened 2024-03-18
Add ninja cover image

NilsJacobsen:
image


NiklasBuchfink (approved):
Looks good. We need to update them from time to time when we update the comment styles.

PR #2400, opened 2024-03-18
update: add default converter for settings object

NilsJacobsen:


NilsJacobsen:
@felixhaeberle fixed the passing of objects, but still working on functions.

PR #2392, opened 2024-03-17
introduce settings component in Sherlock 🕵️‍♂️

felixhaeberle:
@NilsJacobsen this is your playground for the settings component in sherlock

TODO's:
- [X] bug: can't pass an object easily in HTML, we need some sort of framework to run the lit component in. does lit itself provide a runtime? 

-> we are now ging with serializable props as decided in https://linear.app/opral/issue/INLMC-26/do-the-ui-component-props-need-to-be-serializable

- [x] move out of the sidebar into big window screen
- [x] introduce settings icon to open settings window
- [x] add styling to component

PR #2411, opened 2024-03-19
update component api

NilsJacobsen:
- according to https://linear.app/opral/issue/INLMC-22/results-after-call


felixhaeberle:
@NilsJacobsen ah I see you are still in the works of updating the API. afterwards, #2392 latest changes should work.

PR #2456, opened 2024-03-22
enable sparse profile for lix

janfjohannes:
enable sparse profile for lix


janfjohannes:

PR #2469, opened 2024-03-26
Fix PARJS-47 - Astro-adapter `languageTag()` not being set properly on windows

LorisSigrist:
This PR fixes https://github.com/opral/inlang-paraglide-js/issues/45

The issue was caused by the `runtime.js` module being instantiated twice. The middleware would correctly set the language, but in a separate instance of the module. The rest of the app didn't get the memo. 

This is fixed by normalizing the path passed to the internal import-alias plugin.

### DIFF
--- a/inlang/source-code/paraglide/paraglide-js-adapter-astro/src/utilts.ts
+++ b/inlang/source-code/paraglide/paraglide-js-adapter-astro/src/utilts.ts
@@ -1,0 +1,13 @@
+// vendored in from vite
+import path from "node:path"
+
+const windowsSlashRE = /\\/g
+function slash(p: string): string {
+	return p.replace(windowsSlashRE, "/")
+}
+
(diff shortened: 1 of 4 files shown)


inlang/source-code/paraglide/paraglide-js-adapter-astro/src/utilts.ts
```diff
@@ -0,0 +1,13 @@
...
+import path from "node:path"
+
+const windowsSlashRE = /\\/g
+function slash(p: string): string {
+	return p.replace(windowsSlashRE, "/")
+}
+
+const isWindows = typeof process !== "undefined" && process.platform === "win32"
+
+export function normaizePath(id: string) {
```
dallyh, line 11: Small typo here normaizePath() => normalizePath()

LorisSigrist:
Just tested this branch on the Windows Shadow-PC & it works!

PR #2497, opened 2024-03-28
refactor parseOrigin

janfjohannes:
refactor parseOrigin


inlang/source-code/ide-extension/src/utilities/settings/getGitOrigin.test.ts
```diff
@@ -1,39 +1,39 @@
...
 		const origin = await getGitOrigin()
-		expect(origin).toEqual("https://github.com/user/repo")
+		expect(origin).toEqual("github.com/inlang/ci-test-repo.git")
 	})
 
-	it("should handle errors and return undefined", async () => {
-		// @ts-expect-error
-		lixClient.findRepoRoot.mockRejectedValueOnce(new Error("Mock error"))
+	it.todo("should handle errors and return undefined", async () => {
+		// FIXME: either return not existing worpsace forlder here as mock or mock the findRepoRoot as a before
```
janfjohannes, line 36: @felixhaeberle  please take over from here and fix this test case, i am not familiar with vi mocking and cannot make this test case work.

janfjohannes:
@felixhaeberle  handing this over to you as i am not able to fix the vscode bugs. i added a comment what needs to be done, but if you feel like it, i would highly recomend making the vscode extensions getGitOrigin functional, eg not importing vscode and node fs there but making function arguments eg. getGitOrigin({ path, fs }) instead and refactoring the usage. this would make testing trivial instead of relying on unreliable and complex module mocking.

felixhaeberle:
@janfjohannes I'll take over and passing those arguments is probably a good idea. Although there might be a reason why this wasn't the case (for telemetry to work when not path + fs is even set or the extension has trouble setting those up).

felixhaeberle:
> Although there might be a reason why this wasn't the case

Indeed, the function is being used in places where the fs object passing would cause troubles because of serialization concerns – therefore this nullary function is needed. I think it is valid to use here because it doesn't have any side effects.

felixhaeberle (dismissed):
Please check on the mocked return values in my adjusted tests.

janfjohannes:
@NiklasBuchfink  could you explain the ninja errors here?

janfjohannes:
@felixhaeberle  does not look like test failure is caused by this pr now....

Screenshot 2024-04-03 at 18 53 26

felixhaeberle (approved):

PR #2505, opened 2024-04-02
feat: add slugs to adapters

NilsJacobsen:
see https://github.com/opral/inlang.com/issues/18

- Added slugs for optimizing SEO
- remove old slugs from link (only the uniqueID is needed)

### DIFF
--- a/inlang/external-projects/remix-paraglidejs/marketplace-manifest.json
+++ b/inlang/external-projects/remix-paraglidejs/marketplace-manifest.json
@@ -1,5 +1,6 @@
 {
 	"id": "library.brikev.remix-paraglidejs",
+	"slug": "paraglide-remix-i18n",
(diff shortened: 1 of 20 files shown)


LorisSigrist (commented):
Is linking to just `m/[id]` acceptable for SEO? Will google recognize this as a proper link to the page?

NilsJacobsen:
Okey I did a little research:

Your Proposal: Changing the internal links to the new slug -> might have a slightly positive effect because we don't do a 301 internally. But in general this change leads to a considerable number of 301 redirects of old links, which might lead to a slightly worse SEO outcome. But we can not prevent that.

So I gonna change it internally.

LorisSigrist (approved):

LorisSigrist (approved):

PR #2507, opened 2024-04-02
PARJS-59 Generate the Messages Folder during `paraglide-js init`

LorisSigrist:
This PR updates the `paraglide-js init` command:

- Prompts the user for which languages they want to support
- The `messages` folder is now generated by default, along with a `messages/{lang}.json` for each language
- The `identical-pattern` lint rule is no longer added by default. If I understand correctly this will also be the behavior of `createNewProject` in the SDK once that lands

These changes mean that most users should never have to touch `project.inlang/settings.json`, reducing the need for documentation


LorisSigrist:
@samuelstroschein I'm requesting a review from you since I don't really have anyone else to ping-pong off of for changes like this & you are familiar with the paraglide core codebase.

LMK if I should ask someone else in the future I know you're very busy

samuelstroschein (approved):

PR #2537, opened 2024-04-08
SHERL-17: fix preferredLanguageTag wrong auto selection in Sherlock 🕵️‍♂️

felixhaeberle:
This PR aims to fix https://linear.app/opral/issue/SHERL-17/[bug]-incorrect-error-message-when-sherlock-preferred-language-is-set


jldec (changes requested):
@felixhaeberle I'm still seeing the same issue as reported.

- set sherlock preview language to `de` in a project which has `de`
- open a project with only `en`
- look for the preview of any message in a code file

The statusbar reports `en`, but the message previews still report errors.

Screenshot 2024-04-09 at 17 55 10

felixhaeberle:
I'll investigate once more. I guess there are two issues involved here. First is already fixed -> Fallback to source language if preferredLanguageTag isn't in project. Second will be to fix the message (first find out why it is appearing).

jldec:
> I'll investigate once more. I guess there are two issues involved here. First is already fixed -> Fallback to source language if VS code setting of preferredLanguageTag isn't in project. Second will be to fix the message (first find out why it is appearing).

The "preferred" language tag logic added to [statusBar.ts](https://github.com/opral/monorepo/pull/2537/files#diff-f2e67eeb4eb388c215c948eade3e27446b2455e9c070875ec474f2c16c99cb3bR37-R39) needs to be extracted and used in [messagePreview.ts](https://github.com/opral/monorepo/blob/b25aaab3e0eff19d83ed36d419cc25ea06686d56/inlang/source-code/ide-extension/src/decorations/messagePreview.ts#L55) also.

felixhaeberle:
@jldec Watch this LOOM :)
https://www.loom.com/share/f4cc4bb7d7454035937d96c1a4122c03

jldec (approved):
works for me now - thanks @felixhaeberle

PR #2548, opened 2024-04-09
Non reactive errors

martin-lysk:
Lix currently returns a reactive property called `errors`. With this one Lix has 3 kinds of error handlings:

1. Function throws
2. Function catches and adds errors to the errors property
3. Function catches and returns an errors object as function result

This PR removes the second case for now and moves error collection logic to the thin layer using the lix SDK.

In the future we will have defined errors returned in the return object of a function called. (compare 3.) for errors that the client can react on and can deal with and throw for unexpected errors like asertation errors wrong url formats or state issues (file system has changed unexpectedly).


janfjohannes:
@martin-lysk  dont forget to fix the ci

Screenshot 2024-04-09 at 23 22 02

martin-lysk:
We looked into it - we cant merge it until openRepo throws in cases like this one:
https://github.com/opral/monorepo/blob/de169bee1a85590e51bc2af6490855a613496772/lix/packages/client/src/openRepository.ts#L385
@janfjohannes shall i fix it in this repo or do you think we should merge it together with your branch?

inlang/source-code/editor/src/pages/@host/@owner/@repository/State.tsx
```diff
@@ -281,23 +291,19 @@ export function EditorStateProvider(props: { children: JSXElement }) {
...
 						window.repo = newRepo
 					}
-
-					if (newRepo.errors().length > 0) {
-						setLixErrors(newRepo.errors())
-						return
-					} else {
-						setLixErrors([])
-					}
+					// TODO replace this with `setLixErrors([])` as soon as repo throws
```
janfjohannes, line 294: @martin-lysk  what is the impact of this? is this a precondition to mergin?

janfjohannes:
> We looked into it - we cant merge it until openRepo throws in cases like this one:
> 
> https://github.com/opral/monorepo/blob/de169bee1a85590e51bc2af6490855a613496772/lix/packages/client/src/openRepository.ts#L385
> 
> 
> @janfjohannes shall i fix it in this repo or do you think we should merge it together with your branch?

as discussed on call. lets fix this iin open repo , i can merge the three lines manually on my side : )

inlang/source-code/editor/src/pages/@host/@owner/@repository/State.tsx
janfjohannes: maybe its just a forgotten comment

inlang/source-code/editor/src/pages/@host/@owner/@repository/State.tsx
```diff
@@ -306,33 +312,86 @@ export function EditorStateProvider(props: { children: JSXElement }) {
 		}
 	)
 
-	repo()?.errors.subscribe((errors: any) => {
-		setLixErrors(errors)
-	})
+	async function pushChanges(args: {
```
janfjohannes, line 315: @martin-lysk  you moved the fuction up, which make 2 pane diffing impossible for the function, is there any reason, or could you move it back to the old position?

janfjohannes (changes requested):

inlang/source-code/editor/src/pages/@host/@owner/@repository/State.tsx
```diff
@@ -459,11 +518,40 @@ export function EditorStateProvider(props: { children: JSXElement }) {
...
-			const repoMeta = await repo()?.getMeta()
+			const repoMeta = await repo()!.getMeta()
+			if ("error" in repoMeta) {
+				setLixErrors([repoMeta.error, ...lixErrors()])
+			}
 			return repoMeta
 		}
 	)
 
+	const isForkSyncDisabled = () =>
```
janfjohannes, line 529: @martin-lysk  you moved this function which makes diffing impossible in the pr, could you move it back or was there a reason?

inlang/source-code/editor/src/pages/@host/@owner/@repository/State.tsx
```diff
@@ -459,11 +518,40 @@ export function EditorStateProvider(props: { children: JSXElement }) {
 			}
 		},
 		async () => {
-			const repoMeta = await repo()?.getMeta()
+			const repoMeta = await repo()!.getMeta()
```
janfjohannes, line 521: we should not do ! here. repo should be an argument from the previous function insteafd of unwrapping repo() signal again in the effect, i think

janfjohannes (changes requested):

samuelstroschein:
> @martin-lysk  In the future we will have defined errors returned in the return object of a function called.

Please elaborate. 

If you refer to exposing a "result" like type e.g. `const { data, error } = await repo.foo()`, I urge reconsideration. The JS ecosystem expects throws. Error handling in frontend frameworks for example, will not trigger anymore. While you could argue that "but frontend frameworks will get a wrapper", the problem is caused by steering away from standards. I know that error handling in JS is frustrating but we win nothing by deviating from the standard.

Supabase switching to result types https://github.com/supabase/supabase-js/issues/32 overwhelmingly downvotes after the change was merged. Counter proposals are happening https://github.com/supabase/supabase-js/issues/885 to revert the change are happening too.

In short, while `Result` is nice internally, exposing a result type externally is breaking standards and thereby compatibility.

martin-lysk:
@samuelstroschein Jan and I concluded this a the way to go yesterday. @janfjohannes please hop in with your thoughts as well on this on.

My thoughts on this:

1. DX
Javascript is just broken when it comes to throwing errors. Why? We need a way to let the consumer know the cases to be handled. I agree on the standard part but the dx of just marking a function as `@throws` pretty poor - you can't let the user know what errors to expect.  

The user of the lib will have to always look at the documentation and keep track of changes in errors a function can produce to be able to react to them vs having a compile time information about the errors that should be handled. 

I see the point of making a "throwing" version as suggested by the followup ticket in supabase but as long as lix is used internally we should focus on making a solid error handling and go for the returned error property that will Definitly help us define the error handling and communicate changes over the team (via typsaftynes). 

2. Inter Process communication
Lix will most likely run in another process/thread/worker/deamon soon so we will have a interprocess communication going on between consumer and Excution and returning a Serializable result also including errors will be easier.

inlang/source-code/editor/src/pages/@host/@owner/@repository/State.tsx
martin-lysk: it removed the repo argument and make it part of the state it self - i needed to move it at the position to have access to repo()

inlang/source-code/editor/src/pages/@host/@owner/@repository/State.tsx
martin-lysk: Good catch, That was meant to fix the catching in openRepo - which it didn't because of its async behaviour :-( now since openRepo throws - this is was removed with the recent push

inlang/source-code/editor/src/pages/@host/@owner/@repository/State.tsx
martin-lysk: it removed the repo argument and make it part of the state it self - i needed to move it at the position to have access to repo()

inlang/source-code/editor/src/pages/@host/@owner/@repository/State.tsx
martin-lysk: @NiklasBuchfink I changed this to a bang (!) because it should be available at this state and i was fine with this for now - we discussed this on the call. but @janfjohannes also noticed that :) please decide if we should change that in this pr

samuelstroschein:
@janfjohannes @martin-lysk I opened https://linear.app/opral/issue/INBOX-60/error-handling-in-lix-and-inlang-message-sdk. 

I don't understand your decision here. If we agree that compatibility issues will arise, a migration to throwing will come because users will complain as they do with supabase.

inlang/source-code/editor/src/pages/@host/@owner/@repository/State.tsx
NiklasBuchfink: I think @janfjohannes is right, we can pass it as an argument as it is done in the previous function. This should eliminate the need for the bang.

inlang/source-code/editor/src/pages/@host/@owner/@repository/State.tsx
martin-lysk: I just double checked we added those lines at the top of the function and removed the repo parameter.

```
const loadedRepo = repo()
if (!loadedRepo) {
	return { error: new PushException("Repo not loaded") }
}
```

inlang/source-code/editor/src/pages/@host/@owner/@repository/State.tsx
martin-lysk: double checked the function it self has not changed

inlang/source-code/editor/src/pages/@host/@owner/@repository/State.tsx
martin-lysk: Last push should have tackled this one

inlang/source-code/sdk/src/migrations/maybeCreateFirstProjectId.test.ts
```diff
@@ -20,15 +19,3 @@ it("should generate a project id", async () => {
 	const projectId = await generateProjectId({ repo, projectPath: "mocked_project_path" })
 	expect(projectId).toBe("432d7ef29c510e99d95e2d14ef57a0797a1603859b5a851b7dff7e77161b8c08")
 })
-
-it("should return undefined if repoMeta contains error", async () => {
```
janfjohannes, line 24: why was this test removed? seems unrelated

janfjohannes (approved):

inlang/source-code/sdk/src/migrations/maybeCreateFirstProjectId.test.ts
janfjohannes: @martin-lysk  approved, just please check if this test was removed by accident, as it was still there last time i reviewed this pr

martin-lysk: The behaviour of openRepository has changed it will fail if the repo does not exist, errors got to be handled before so - no need for this test anymore?

PR #2549, opened 2024-04-10
sdk/multi-project-test with MOCK_TRANSLATE_LOCAL

jldec:
- [x] sdk/muti-project-test runs e2e translation test on 3 proejcts in the same repo (root, /project2-dir, /project3-dir)
- [x] MOCK_TRANSLATE_LOCAL allows the cli to mock translate messages without any rpc server 
- [x] `inlang machine translate --nobar` (-n) option to suppress progressbar  

This is a followup to #2528


jldec:
@janfjohannes the new multi-project-test is failing in CI [here](https://github.com/opral/monorepo/actions/runs/8623893946/job/23637927630#step:8:5661) but it works locally.

The part of the test which fails involves migrating the old project.inlang.json to the new project.inlang directory, so I'm wondering if this could be lix-related (just a guess.)

Unfortunately the logging in the test is not great - so I can't tell for sure if that's the issue. The test invokes the cli with node:childProcess.exec so all console stdio is swallowed.

jldec:
@janfjohannes - i just pushed a slightly more elaborate test with try/catch to try to debug the CI failure

inlang/source-code/sdk/multi-project-test/multi-project.test.ts
```diff
@@ -0,0 +1,95 @@
...
+	"project3_message_key_1": "Mock translate local en to de: Generated message (1)",
+	"project3_message_key_2": "Mock translate local en to de: Generated message (2)",
+	"project3_message_key_3": "Mock translate local en to de: Generated message (3)"
+}`)
+			} catch (e) {
+				console.error(e)
+				throw e
+			}
+		},
+		{ timeout: 20000 }
```
janfjohannes, line 93: could you add new line to end of file setting and maybe reduce the timeout? i assume you dont use google but your translate mock so this should be able to run in 5 seconds or so?

janfjohannes (approved):
just added  a minor comment, but looks good

inlang/source-code/sdk/multi-project-test/multi-project.test.ts
jldec: For some reason the test that migrates the old project.inlang.json to the new project.inlang directory takes a lot longer (several seconds).

jldec: adjusted but left some room for slow CI environments

PR #2551, opened 2024-04-10
PARJS-72 - Eagerly set Language on Client Load

LorisSigrist:
This PR changes _when_ the language get's set on the client. Up until now it would match the behavior on the server where it was unavailable in load functions & only got set for rendering (on the server you would have cross-talk otherwise). However, on the client there is no need for this restriction. 

With this change the language is now set immediately on the client, before load functions run.


LorisSigrist:
Closed for the reasons outlined in opral/inlang-paraglide-js#69

PR #2575, opened 2024-04-11
lix refactor

janfjohannes:
lix refactor

Summary:

Test Plan:

---
* #2610
* #2609
* __->__ #2575


janfjohannes:
@martin-lysk  i still have this error:

Screenshot 2024-04-12 at 01 49 09


you could allready think about if you have an idea, i think you had the same error when you moved the checkout into lix, right?

lix/packages/client/src/git/checkout.ts
```diff
@@ -68,7 +86,7 @@ export async function checkout({
 	force = false,
 	track = true,
 	cache = {},
-}) {
+}: any) {
```
janfjohannes, line 89: the type annotations do not work anymore, if you know how to activate them in ts pls let me know. i think using a big plugin or build step is too much hassle, then we just convert it next time we touch this. as doCheckout is not exposed as lix api this is probably fine until then

martin-lysk: Yeah - did some research https://alexharri.com/blog/jsdoc-as-an-alternative-typescript-syntax#:~:text=JSDoc%20does%20not%20work%20in,frustrating%2C%20especially%20for%20larger%20modules. :-( we can define the types later on for now i agree any should do the trikc

lix/packages/client/src/git/commit.ts
```diff
@@ -18,7 +20,68 @@ type PartialEntry = {
...
+		include,
+	}: // TODO: exclude,
+	{
+		author?: any
+		message: string
+		include: string[]
+		// exclude: string[]
+	}
+) {
+	if (include) {
```
martin-lysk, line 50: include is non optional at the moment - gues this if is not needed?

lix/packages/client/src/git/findRepoRoot.ts
```diff
@@ -0,0 +1,14 @@
...
+export async function findRepoRoot(args: {
+	nodeishFs: NodeishFilesystem
+	path: string
+}): Promise<string | undefined> {
+	const gitroot = await findRoot({
+		fs: args.nodeishFs,
+		filepath: args.path,
+	}).catch(() => undefined)
+
+	return gitroot ? "file://" + gitroot : undefined
```
martin-lysk, line 13: Why do we add file:// to the path here? Guess this is not a change in this pr but we might add a TODO to document the reason

lix/packages/client/src/git/getBranches.ts
```diff
@@ -0,0 +1,29 @@
...
+		throw new Error("Could not find repo url, only github supported for getBranches at the moment")
+	}
+	const serverRefs = await isoGit
+		.listServerRefs({
+			url: ctx.gitUrl,
+			corsProxy: ctx.gitProxyUrl,
+			prefix: "refs/heads",
+			http: makeHttpClient({ verbose: ctx.debug, description: "getBranches" }),
+		})
+		.catch((error) => {
```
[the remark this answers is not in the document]
martin-lysk: seems like a followup thing checked the original code - just copied

lix/packages/client/src/github/forkStatus.ts
```diff
@@ -0,0 +1,141 @@
...
+		githubClient,
+		// setErrors,
+		gitProxyUrl,
+	} = ctx
+
+	if (!gitUrl) {
+		throw new Error("Could not find repo url, only github supported for forkStatus at the moment")
+	}
+
+	// const repo = await this
```
martin-lysk, line 23: guess this is a leftover of the open repo beeing async - we can remove this comment

lix/packages/client/src/github/forkStatus.ts
```diff
@@ -0,0 +1,138 @@
...
+		isFork: boolean
+		parent: { url: string }
+	}
+
+	if (!isFork) {
+		return { error: "could not get fork upstream or repo not a fork" }
+	}
+
+	const forkFs = state.nodeishFs
+
```
martin-lysk, line 32: We could be more specific like this? We don't have to do this just that i got the api right here
```suggestion
	const { error, isFork, parent } = (await getMeta(ctx)) as {
		error: any
		isFork: boolean
		parent: { url: string }
	}

	if (error) {
		return { error: "could not get fork upstream" }
	}

	if (!isFork) {
		return { error: "repo is not a fork" }
	}
```

lix/packages/client/src/lix/checkoutPlaceholders.ts
```diff
@@ -0,0 +1,86 @@
...
+import isoGit from "../../vendored/isomorphic-git/index.js"
+import { doCheckout } from "../git/checkout.js"
+import { modeToFileType } from "../git/helpers.js"
+
+export async function checkOutPlaceholders(ctx: RepoContext, state: RepoState) {
+	const { rawFs, cache, dir, experimentalFeatures } = ctx
+	const { branchName, checkedOut, sparseFilter } = state
+
+	if (!experimentalFeatures.lazyClone) {
+		return
```
martin-lysk, line 11: I would rather throw in such a case - this function should not be called when the flag is not turned on?

lix/packages/client/src/git/commit.ts
janfjohannes: this will be optional when adding default behaviour

lix/packages/client/src/git/findRepoRoot.ts
janfjohannes: because openRepo expects a valid lix url  this is needed to pass the return value of findRepoRoot to open repo:

openRepository(await findRepoRoot ....


please explain why this was not clear as i am not sure what a helpful comment would be

lix/packages/client/src/lix/checkoutPlaceholders.ts
janfjohannes: i think we dont have to handle this at all anyomore.

lix/packages/client/src/github/forkStatus.ts
janfjohannes: sure

lix/packages/client/src/repoState.ts
```diff
@@ -0,0 +1,212 @@
...
+	}: {
+		execute: () => any
+		prop: any
+		argumentsList: any[]
+	}) {
+		const filename = argumentsList?.[0]?.replace(/^(\.)?(\/)?\//, "")
+		const pathParts = filename?.split("/") || []
+		const rootObject = pathParts[0]
+
+		// TODO: we need to delay readdir to prevent race condition for removing placeholders, remove when we have our own checkout implementation
```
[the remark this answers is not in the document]
janfjohannes: nope, we still need this for now and the comment is still correct

lix/packages/client/src/git/getOrigin.ts
```diff
@@ -0,0 +1,15 @@
+import { parseOrigin } from "../helpers.js"
+import { listRemotes } from "./listRemotes.js"
+import type { RepoContext, RepoState } from "../openRepository.js"
+
+/**
+ * Parses the origin from remotes.
```
janfjohannes, line 6:
```suggestion
 * Parses the origin from remotes...
```

janfjohannes:
@martin-lysk @NiklasBuchfink   please do manual check, for me everything works locally

NiklasBuchfink (approved):
I have checked use cases according to the changes made. Everything LGTM 👍

PR #2583, opened 2024-04-12
Introduce settings component in fink

NiklasBuchfink:
Bildschirm­foto 2024-04-12 um 18 37 27


samuelstroschein (commented):
why a popup? seems much better if you just open a page?

image


Mock page

CleanShot 2024-04-12 at 14 20 26@2x

NiklasBuchfink:
@samuelstroschein Revert a few commits and you will have the page. I liked the page as well, but:
- full page width doesn't make sense, because you have will have a 1120px width input for a two-character language tag -> make it 600px width and we almost have a modal
- the back button UX felt weird with a toggle and an x button would feel like a fake modal
- minor: both floats will conflict, so we need to hide one for the other, but it is the same pattern which is good
- minor: we unmount and mount the endless language list that results in a lag for large repos

Should I try to find a solution for this? I agree, that a page looks cleaner.

samuelstroschein:
@NiklasBuchfink ship what you deem right. 

the model seems wrong to me. i can be wrong though.

jldec:
just a few thoughts - nothing blocking first 🚢  + iterate

- personally i would also prefer a settings page. 
  - suggest limiting to a max-width with air on the margins left and right. (you're right, 1214 is too wide)
  - settings page would then become linkable (bookmarkable) and navigable in history 
  - (Q: could the settings scomponent do the max-width layout container?)

- `Attention, you have unsaved changes.` popup inside the popup feels "stressig"  
  - (again, maybe this is component responsibility?)
  - I'd prefer simple save/cancel buttons at the top right, not overlaid on wide views

- selecting (add/remove?) lint rules seems more important than adjusting warning levels
  - (i think this was discussed elsewhere already)

- it was unclear to me what the heading with the name of the message plugin "i18next" was indicating.
  - I'd prefer some text to explain: "this project uses the "xxx" plugin for message storage."
  - How about using visual indicator (bg color) to group those settings (path patterns etc.) which are specific to the plugin inside a plugin box.

- suggest linking to wikipedia or other reference for BCP-47
- style decision - but for me, input fields don't need to be full width, especiall for short values like the language code

NiklasBuchfink:
@jldec I fully agree with all your suggestions. Many of your points are settings component responsibilities. cc @NilsJacobsen
- some are addressed in upcoming milestones
- at some point we had the buttons always at the top, I remember that I shared a sketch of that
- Perhaps some dividing lines or a "module" prefix could make the different sections a little clearer
- I had the same reaction regarding the input field length, as the input length indicates the user's expected length of the text

I like to give the settings page a try again. A separate page would require all query parameters and the project and branch switching header again. My proposal is to overlay the settings component over the message list. Replacing the message list by mounting and unmounting 10.000 elements from the dom feels wrong.

This is what the page would look like. I need to find a better interaction for the settings button:
Bildschirm­foto 2024-04-14 um 14 10 45

I will work on it tomorrow, to get this out.

samuelstroschein:
> @NiklasBuchfink A separate page would require all query parameters and the project and branch switching header again. My proposal is to overlay the settings component over the message list.

I don't see an issue with that. 

Off-topic: The URL schema can surely be improved sometimes the future.

```diff
https://fink.inlang.com/github.com/opral/example?project=%2Fproject.inlang&lang=en&lang=de

https://fink.inlang.com/github.com/opral/example/SETTINGS?project=%2Fproject.inlang&lang=en&lang=de
```

CleanShot 2024-04-14 at 19 59 04@2x

NilsJacobsen:
@jldec 

> Attention, you have unsaved changes. popup inside the popup feels "stressig"

Yes, it is a component responsibility. We could make the text editable or removable if the app devs request that. I copied the wording from Discord.

> selecting (add/remove?) lint rules seems more important than adjusting warning levels

This will come in with this milestone: https://linear.app/opral/project/settings-component-64af33879e1b

> it was unclear to me what the heading with the name of the message plugin "i18next" was indicating.

- Gonna add a `module settings` title to make it clear that is about modules. See https://github.com/opral/inlang-message-ui-components/issues/23
- Visual indication is up to the app. They can style it as they want. I wanted to make the settings as raw as possible to fit every use case. 

> suggest linking to wikipedia or other reference for BCP-47

Links in description will come.

> style decision - but for me, input fields don't need to be full width, especiall for short values like the language code

I added a triage. Fell free to jump in: https://github.com/opral/inlang-message-ui-components/issues/37

NiklasBuchfink:
I've settled with the following:
https://www.loom.com/share/867cf4d9e20f4bc994f2c36dd9cd6f8f?sid=cf8d3f48-b9d8-40a1-8071-fb8391b7663d

TLDR; let's settle with this for now and improve from there:
localhost_4003_github com_NiklasBuchfink_example_branch=test-ninja project=%2Fproject inlang lang=en lang=de

NilsJacobsen:
When I want to login, I get redirected to `http://localhost:3000/de/services/auth/auth-callback` and get a 404. Is this a local issue? @NiklasBuchfink

NilsJacobsen:
Small thing:

You could put the back button to the left and make it's position static to avoid losing it from the view.

Like that:
image

NilsJacobsen:
Other small thing:
You might want to update filtereredLanguageTags when settings are updated.

inlang/source-code/editor/package.json
```diff
@@ -16,6 +16,8 @@
 		"node": ">=18.0.0"
 	},
 	"dependencies": {
+		"@inlang/settings-component": "workspace:*",
```
NilsJacobsen, line 19: I would pin the major version.

NilsJacobsen (approved):
Feels super integrated not alien. Love it.

Issue #2349, opened 2024-03-08
`createNewProject()` v1


samuelstroschein:
## Problem 

Apps can't create a project. 

## Proposal 

Expose a v1 of `createNewProject(path, { options })` in the inlang SDK. 

Until #1585 is implemented, we can create a new project with the inlang message format plugins as default, creating a folder in the inlang project to persist the messages. That unblocks @opral/inlang-fink and @opral/inlang-sherlock to start implementing onboarding flows. 

```
.
└── project.inlang
    ├── plugin.inlang.messageFormat/
    │   ├── de.json
    │   ├── en.json
    │   └── ... .json
    └── settings.json

```

```ts
const project = await createNewProject("blue_skydiver.inlang")

```

### Tasks

- [ ] validate that project ends with `*.inlang`
- [ ] pre-install common lint rules

## Additional information 

I propose to ship the most used lint rules by default. They can be uninstalled later

samuelstroschein:
Duplicate in linear https://linear.app/opral/issue/MESDK-31/createnewproject-v1

felixhaeberle:
@samuelstroschein why are we closing github issues which are connected to **not** done linear issues?

PR #2590, opened 2024-04-15
createNewProject()

jldec:
Resolves MESDK-31 (https://github.com/opral/monorepo/issues/2349)

```ts
createNewProject(args: {
	projectPath: string
	repo: Repository
	projectSettings: ProjectSettings
}): Promise<void>
```

- Creates a new project in the given directory.
- The directory must be an absolute path, must not exist, and must end with {name}.inlang
- Exports `sdk.defaultProjectSettings`  with values the same as paraglide init for now.
  - The default pathpattern `./messages/...` stores messages outside of the projject.
```
.
├── messages
│   ├── de.json
│   └── en.json
├── project.inlang
│   └── settings.json
```

### future direction
```
.
└── project.inlang
    ├── [internal-messages-directory tbd]
    │   ├── [internal messages file structure... tbd]
    └── settings.json
```


inlang/source-code/sdk/src/defaultProjectSettings.ts
```diff
@@ -0,0 +1,27 @@
...
+
+		// default to the message format plugin because it supports all features
+		"https://cdn.jsdelivr.net/npm/@inlang/plugin-message-format@latest/dist/index.js",
+
+		// the m function matcher should be installed by default in case Sherlock (VS Code extension) is adopted
+		"https://cdn.jsdelivr.net/npm/@inlang/plugin-m-function-matcher@latest/dist/index.js",
+	],
+	"plugin.inlang.messageFormat": {
+		pathPattern: "./messages/{languageTag}.json",
+	},
```
samuelstroschein, line 26: I recommend to use `./plugin.inlang.messageFormat/messages/{languageTag}.json` instead. 

- generally: namespace plugin "storage" to avoid clashes with features we will introduce
- in this case: we will very likely want to use `/messages` for persistency

samuelstroschein (commented):
@martin-lysk @jldec agree?

inlang/source-code/sdk/src/defaultProjectSettings.ts
martin-lysk: sounds good to me.

inlang/source-code/sdk/src/createNewProject.test.ts
```diff
@@ -0,0 +1,111 @@
...
+describe("createNewProject", () => {
+	it("should throw if a path does not end with .inlang", async () => {
+		try {
+			await createNewProject({
+				projectPath: "/boom/ba/da/bang",
+				repo: await mockRepo(),
+			})
+			throw new Error("Expected an error")
+		} catch (e) {
+			expect((e as Error).message).toMatch(
```
martin-lysk, line 21: I would use https://vitest.dev/api/expect#tothrowerror (tip, because i missed it: read till the end it suggest using .rejects for promises)

inlang/source-code/sdk/src/createNewProject.ts
```diff
@@ -0,0 +1,31 @@
...
+ * Uses defaultProjectSettings unless projectSettings are provided.
+ *
+ * @param projectPath - Absolute path to the [name].inlang directory
+ * @param repo - An instance of a lix repo as returned by `openRepository`
+ * @param projectSettings - Optional project settings to use for the new project.
+ */
+export async function createNewProject(args: {
+	projectPath: string
+	repo: Repository
+	projectSettings?: ProjectSettings
```
martin-lysk, line 18: I would suggest we make projectSettings non optional and export the default project settings for the consumer instead - In most cases the configuration will be allowed to be changed from default to specific. For example the language tags are most likely changed in all situations, Manage might allow you to choose the lints (with good defaults so) - To be able to change them one needs to access the defaults to make them visible - and the can pass them as an argument than.

inlang/source-code/sdk/src/defaultProjectSettings.ts
```diff
@@ -0,0 +1,27 @@
...
+		"https://cdn.jsdelivr.net/npm/@inlang/message-lint-rule-empty-pattern@latest/dist/index.js",
+		"https://cdn.jsdelivr.net/npm/@inlang/message-lint-rule-missing-translation@latest/dist/index.js",
+		"https://cdn.jsdelivr.net/npm/@inlang/message-lint-rule-without-source@latest/dist/index.js",
+		"https://cdn.jsdelivr.net/npm/@inlang/message-lint-rule-valid-js-identifier@latest/dist/index.js",
+
+		// default to the message format plugin because it supports all features
+		"https://cdn.jsdelivr.net/npm/@inlang/plugin-message-format@latest/dist/index.js",
+
+		// the m function matcher should be installed by default in case Sherlock (VS Code extension) is adopted
+		"https://cdn.jsdelivr.net/npm/@inlang/plugin-m-function-matcher@latest/dist/index.js",
```
martin-lysk, line 22: Not important for this PR -can be static for now - just a comment:

if we don't make the projectSettings parameter optional - we could make defaults dynamic as a helper function in the sdk `getSuggestedDefaults` that can check if the project is using paraglide or if it is fink that want's to create the project or paraglide and suggest specific defaults ... just a thought - as stated not needed in this pr

inlang/source-code/sdk/src/defaultProjectSettings.ts
martin-lysk: We should also consider adding a version identifier to the path - we discussed this earlier - we might need to migrate.

inlang/source-code/sdk/src/createNewProject.ts
jldec: > I would suggest we make projectSettings non optional 
 What's the benefit of making it non-optional - i can see use cases where no interaction before creating a project is desireable. 
 
 > export the default project settings for the consumer instead
 are you suggesting that we expose defaultProjectSettings on the sdk api,

inlang/source-code/sdk/src/defaultProjectSettings.ts
jldec: @martin-lysk do you have link to the earlier discussion?

Versioning directory paths is a bit too in-your-face for my taste (e.g. when paths show up in git logs, diffs, etc)
Why can't we rely on the schema refs inside the files?

inlang/source-code/sdk/src/createNewProject.test.ts
jldec: @martin-lysk I prefer the existing pattern even though it's a bit more verbose.

The suggested pattern using `async expect(() => fn()).rejects.toThrowError()` is more consise, but less readable because it requires internal knowledge of `async expect()` and `reject`, and forces the async function to be called without await. e.g.

```ts
it("should throw if a path does not end with .inlang", async () => {
	const repo = await mockRepo()
	const projectPath = "/test/name.inl"
	await expect(() => createNewProject({ projectPath, repo })).rejects.toThrowError(
		'Expected a path ending in "{name}.inlang" but received "/test/name.inl"'
	)
})
```

inlang/source-code/sdk/src/defaultProjectSettings.ts
samuelstroschein: > @jldec Are you saying we should default to ./plugin.inlang.messageFormat/{languageTag}.json or are you saying that the current default ./messages/{languageTag}.json is fine?

Change the path pattern to `./plugin.inlang.messageFormat/{languageTag}.json` to namespace plugin data and avoid conflicts. 

> Again, I'm a bit wary of adding long folder paths into the main tree of users' repos without a really good reason.

Dont forget that users won't see the inside of a `.inlang` file in the long run. I see a good reason here to establish a good practice of namespacing plugins to avoid conflicts.

jldec: @samuelstroschein I talked with @martin-lysk and I think I understand your comment:
- for this PR the messages should live at `./project.inlang/plugin.inlang.messageFormat/{languageTag}.json`
- for non-plugin based persistence, we should reserve paths starting with `./project.inlang/messages/...`

jldec: BTW - at the time of this writing the [default settings](https://github.com/opral/monorepo/blob/d573ab8c43ee27c4be14e6e2561a9eb6364f18af/inlang/source-code/sdk/src/defaultProjectSettings.ts#L25) had `pathPattern: "./messages/{languageTag}.json",` which results in a tree like this:

 ```
.
├── messages
│   ├── de.json
│   └── en.json
├── project.inlang
│   └── settings.json
```

jldec: options to resolve the issue with relative pathpattern pointing ☝️ 

1. leave /messages outside the project for this PR
2. use a magic prefix in the pathattern like `$/...`
3. use a different key like `projectpath`: instead of `pathpattern`: and use that to populate pathpattern at runtime
4. change the meaning of pathpattern and break/migrate all existing projects 😬 
5. write the project `{name}.inlang` into the pathpattern 😬 

@martin-lysk @samuelstroschein wdyt?

samuelstroschein: prob 1 then.

martin-lysk: I thought about this Option 1 makes sense anyway since this pr creates a default config with a plugin **not** with default persistence. Message plugins currently write to and read from a folder outside of the .inlang folder. When we settled on a format for the message during persistence and provide a persistence mechanism we don't migrate to the persistence we will load the message using the plugin (and we rename the method loadMessages - to import)

inlang/source-code/sdk/src/createNewProject.ts
martin-lysk: Default can be forgoton/ignored. If the default case is that the consumer of the api will pass a settings object i would enforce the consumer to pass it at compile time. Documentation of the function can guide how to fill it with a default value - but it is an explicit decision instead of an optional choice.

inlang/source-code/sdk/src/defaultProjectSettings.ts
jldec: Thanks for the feedback - will go with 1 then for this PR.

inlang/source-code/sdk/src/createNewProject.ts
jldec: 👍 
"explicit" is more readable too.

inlang/source-code/sdk/src/defaultProjectSettings.ts
jldec: makes sense - i like it.

inlang/source-code/sdk/src/defaultProjectSettings.ts
jldec: https://github.com/opral/inlang-message-sdk/issues/34 (MESDK-70)

PR #2596, opened 2024-04-16
fixes types in manage and adds comment to website - why any is ok

martin-lysk:
Fixes type issues in separate pr - discussed in call

PR #2562, opened 2024-04-10
Move browser auth to lix client

martin-lysk:
Moves auth to lix client and inject the required env vars to make remove the env dependency

PR #2597, opened 2024-04-16
moves or adds tsc --noEmit from test to build phase 

martin-lysk:
We missed type checks in some packages - most paraglide packages and manage. this lead to type not been caught by the ci. 

This pr moves all `tsc --noEmit` calls from test to build phase and adds them where it was missing to ensure we have type safeness in all packages. 

It was moved to the test phase in the past because of performance issues. I suggest to move it to build phase - since test is to slow during development. If we see problems with this approach we can change it later on.

This needs #2596 and #2595 land in main first to be green

From the chat:

> 
> @inlang/team  martin and me discovered today that not all packages were doing type checks in ci leading to missed issues that went to main. in order to fix this martin added a type check to all build commands of packages where the build did not naturally do type checks by using tsc compiler here: https://github.com/opral/monorepo/pull/2562
> 
> 
> before merging this, i would like to get team alignment this is the way to go or if we should instead have typechecking be part of the lint or test run or a new run that would be before build.
> 
> running before build has some advantages like being able to stop on type issues faster without being slowed down by parallel builds, but added complexity of having additional step and also not always checking types on local builds might lead to missing issues locally. i am neutral on this but just want everyone to be on the same page so please just confirm or comment if you find a good reason to do one or the other way
> 
> 
> (in future everyone please make sure to add typechecking for everything using esbuild, vite, rollup or webpack as these wont error on type issues)


martin-lysk:
@LorisSigrist paraglide doesn't seem to like the new build proccess - can i pass this PR over to you to have a look?

janfjohannes:
@martin-lysk  @LorisSigrist  i noticed that some types are only working after they have been build. this means that having typechecks in the same ci step as build might fail in unexpected ways. maybe this is related to some of the issues you are seeing, not sure. if yes we should move type checks out of build into its own pipeline step and run after build.

PR #2602, opened 2024-04-17
Fix flaky prettier

martin-lysk:
This PR sets the trailing comma rule in prettier to [es5](https://prettier.io/docs/en/options.html#trailing-commas) (the config that our project is currently formatted with) - but ensures cli and vs code extension format in the same way.

It also sets startRang and endRange so that the vscode extension always formats the whole file.

More details can be found in this loom:
https://www.loom.com/share/29fcf4d2cc5640d9923c2674744ca4b6


martin-lysk:
@LorisSigrist selected you as reviewer because your project is affected the most - should be only cosmetical changes so.

LorisSigrist (approved):
Cosmetical changes are great!

LorisSigrist:
Can I merge?

PR #2606, opened 2024-04-18
Nilsjacobsen/inweb 67 broken manageinlangcom page

NilsJacobsen:
see: https://github.com/opral/inlang.com/issues/47


inlang/source-code/manage/src/util/TailwindMixin.ts
NilsJacobsen, line 1: @martin-lysk needed to @ts-ignore it in order to render the page. We will get rid of this as soon as we remove manage.

PR #2609, opened 2024-04-18
Make lix setup work in sparse checkout

janfjohannes:
Make lix setup work in sparse checkout

---
* #2610
* __->__ #2609


lix/packages/server/eslint.config.js
```diff
@@ -0,0 +1,17 @@
+import lixStyle from "@lix-js/code-style"
+import globals from "globals"
+
+export default [
+	...lixStyle,
+	{
+		languageOptions: {
```
janfjohannes, line 7: add overrides for tests

janfjohannes: @martin-lysk  i saw we needed to ignore the node type issues for tests before too, so i will leave this until it becomes more annoying. lets try to keep node deps low anywyas, maybe make a helper module that has all the ignores or add this to the type setup

PR #2612, opened 2024-04-18
PARJS-18 Make Paraglide Adapters Standalone Packages

LorisSigrist:
This PR drops the `paraglide-js-adapter-x` naming convention in favor of the much shorter `paraglide-x`. This is a very breaking change, but it really needs to be done at some point. Might as-well be as soon as possible. 

It also updates all the guides to no longer use the word "Adapter".

As for the rollout:
- Old packages remain available but won't be updated
- Old packages are marked as depreciated on npm & link to the new packages
- Post updated Names on all our social channels


samuelstroschein (approved):
The Rollout makes sense. Haven't looked at all changes. Giving approval anticipating that everything will be OK. 

_PS it would help if you state in a PR what would be reviewed_

PR #2634, opened 2024-04-23
feat: add new paraglide-next header

NilsJacobsen:
see: https://github.com/opral/inlang-paraglide-js/issues/90

image


LorisSigrist (approved):
Looks great locally

PR #2644, opened 2024-04-24
reset nx to 17, fix website depl

janfjohannes:


janfjohannes:
@NilsJacobsen @LorisSigrist  cc

PR #2659, opened 2024-04-25
INCLI-2 inlang machine translate calls rpc only if message has missing translations

jldec:
Resolves INCLI-2 (https://github.com/opral/inlang-cli/issues/2)

- [x] Introduces a filter on the CLI client side to call the rpc server only when actually needed for missting translations.
(Before it would send all messages over the wire regardless)
- [x] Small fixes to sdk/load-test - can be ignored for PR review


felixhaeberle (approved):
lgtm 👍

PR #2655, opened 2024-04-25
MESDK-77 Requirements for Gender and Plurals [WORKING DRAFT]

LorisSigrist:
*please don't delete this branch - parts are still being extracted thanks - @jldec*

This PR updates the internal `Message` AST to include all the functionality we need to implement plurals, genders and other formatting functions.

It also adds the **experimental and not for public use** `icu2-message-format` plugin which is able to load (not save) messages from icu2 files. 

CI will probably freak out because the Message-Type was heavily edited, but that's alright for the moment 

**HOW TO TEST**
- Build `versioned-interfaces/message`
- Build `plugin/icu2-message-format`
- Build SDK
- Run `test` in `development-projects/icu2`

This should log the loaded messages. However, for some reason the SDK doesn't return any values or errors. The plugin does work (can also be seen in the logs).

PR #2700, opened 2024-05-03
MESDK-77 add @inlang/sdk/v2 with MessageBundle and related types

jldec:
Introduces v2 `MessageBundle` and related types, extracted from https://github.com/opral/monorepo/pull/2655
Resolves https://github.com/opral/inlang-message-sdk/issues/40

- [x] Export types from `@inlang/sdk/v2`
- [x] Remove local declarations
- [x] Add draft WIP v2 crud interfaces to `InlangProject`
  - e.g. project.messageBundles.get() and project.messages.getAll()
- [x] Inline LanguageTag from versioned-interfaces

### Usage

```ts
import type { Message, MessageBundle } from '@inlang/sdk/v2'

const myMessageBundle: MessageBundle = {
  // to be filled in
}
```

In case of type name conflicts with existing Message type

```ts
import type * as V2 from '@inlang/sdk/v2'

const myMessageBundle: V2.MessageBundle = {
  // to be filled in
}
```

A message bundle is a collection of Messages which are based on a subset of [ICU MessageFormat 2](https://github.com/unicode-org/message-format-wg).

A bundle has an id (and optional aliases) which can be used to reference messages from UI code. Each message in the bundle targets a different locale.

```ts
export type Message = Static<typeof Message>
export const Message = Type.Object({
	locale: Locale,
	declarations: Type.Array(Declaration),
	selectors: Type.Array(Expression),
	variants: Type.Array(Variant),
})

export type MessageBundle = Static<typeof MessageBundle>
export const MessageBundle = Type.Object({
	id: Type.String(),
	alias: Type.Record(Type.String(), Type.String()),
	messages: Type.Array(Message),
})
```


inlang/source-code/versioned-interfaces/message-bundle/package.json
```diff
@@ -0,0 +1,33 @@
+{
+	"name": "@inlang/message-bundle",
+	"type": "module",
+	"version": "0.0.1",
```
jldec, line 4: Do interfaces need to have a major version >0

LorisSigrist: No, a `0.x.x` version is fine

LorisSigrist (approved):

inlang/source-code/versioned-interfaces/message-bundle/README.md
```diff
@@ -0,0 +1,26 @@
...
+export type Message = Static<typeof Message>
+export const Message = Type.Object({
+	locale: Locale,
+	declarations: Type.Array(Declaration),
+	selectors: Type.Array(Expression),
+	variants: Type.Array(Variant),
+})
+
+export type MessageBundle = Static<typeof MessageBundle>
+export const MessageBundle = Type.Object({
```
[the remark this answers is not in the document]
LorisSigrist: +1 there should be a string-array called `inputOrder` on the MessageBundle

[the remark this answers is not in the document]
martin-lysk: Nope we removed it for now: 
https://linear.app/opral/issue/MESDK-77/requirements-for-gender-and-plurals#comment-8428e8de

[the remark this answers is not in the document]
NilsJacobsen: Where do we get the information about what is inserted in the paraglide function? How can translators add inputs that lead to a break in the ci? Maybe I'm missing something?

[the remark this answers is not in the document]
LorisSigrist: The actual "inputs" are defined as input-declarations on the messages (not the bundle) themselves. The `inputOrder` would simply have been a canonical _order_ to put the inputs in for platforms where the input-order matters.

inlang/source-code/versioned-interfaces/message-bundle/src/interface.ts
```diff
@@ -0,0 +1,115 @@
...
+/**
+ * An expression is a reference to a variable or a function.
+ *
+ * Think of expressions as elements that are rendered to a
+ * text value during runtime.
+ */
+export type Expression = Static<typeof Expression>
+export const Expression = Type.Object({
+	type: Type.Literal("expression"),
+	arg: Type.Union([Literal, VariableReference]),
```
samuelstroschein, line 47: please no abbreviations. what is `arg`? singular argument or plural arguments?

EDIT: i see that message format 2.0 uses `arg` okay... can stay

samuelstroschein, line 47: please no abbreviations. what is `arg`? singular argument or plural arguments?

EDIT: i see that message format 2.0 uses `arg` okay... can stay

inlang/source-code/versioned-interfaces/message-bundle/src/interface.ts
```diff
@@ -0,0 +1,115 @@
...
+	//TODO make this generic so that only Variable-Ref Expressions are allowed
+	value: Expression,
+})
+
+export type LocalDeclaration = Static<typeof InputDeclaration>
+export const LocalDeclaration = Type.Object({
+	type: Type.Literal("local"),
+	name: Type.String(),
+	value: Expression,
+})
```
samuelstroschein, line 94: Do we need local declarations for v1?

inlang/source-code/versioned-interfaces/message-bundle/src/interface.ts
```diff
@@ -0,0 +1,115 @@
...
+ */
+export type Text = Static<typeof Text>
+export const Text = Type.Object({
+	type: Type.Literal("text"),
+	value: Type.String(),
+})
+
+export type VariableReference = Static<typeof VariableReference>
+export const VariableReference = Type.Object({
+	type: Type.Literal("variable"),
```
samuelstroschein, line 21: Mismatch between type name and type string `VariableReference` vs `variable`. Recommendation, follow MessageFormat PascalCase for all types: 

```diff
-variable
+VariableRefernece
```

inlang/source-code/versioned-interfaces/message-bundle/src/interface.ts
```diff
@@ -0,0 +1,115 @@
...
+})
+
+/**
+ * A (text) element that is translatable and rendered to the UI.
+ */
+export type Text = Static<typeof Text>
+export const Text = Type.Object({
+	type: Type.Literal("text"),
+	value: Type.String(),
+})
```
samuelstroschein, line 17: MessageFormat 2.0 seems to have removed the `Text` type in favor of a pure string. What was their reasoning? 

Notes from me: 

- iterating through pattern elements seems much easier if every node has `node.type` property instead of using union matching e.g. `if (typeof node === string || node.type === XYZ)`

samuelstroschein (commented):
Wondering if releasing a v2 of `@inlang/message` isn't better. Apps using the SDK now have name collisions: 

```
import { Message } from "@inlang/sdk" 
```

What message is imported here? 

Recommendation: 

- Release V2 prerelease of `@inlang/message` instead of having a new package
- Suffix all types with V2 e.g. `MessageV2`, `PatternV2`, etc. to avoid name collisions 

(We will run into this problem again with a potential v3. Seems easier to then again suffix `MessageV3` instead of having a new package and 3 potential name collisions)

inlang/source-code/versioned-interfaces/message-bundle/src/interface.ts
LorisSigrist: The string `"variable"` is from the MessageFormat 2.0 Data-model recommendation
https://github.com/unicode-org/message-format-wg/blob/6d7b4ba213e686ff2d403d3025d38d76b42b75f7/spec/data-model/README.md#expressions

inlang/source-code/versioned-interfaces/message-bundle/README.md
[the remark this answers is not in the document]
NilsJacobsen: Sorry you are right :D

inlang/source-code/versioned-interfaces/message-bundle/src/interface.ts
samuelstroschein: Oh wow, they deviated from the previous pattern then. I let @jldec decide. I prefer written out types it's a naming thing though which no real impact

jldec:
> Wondering if releasing a v2 of @inlang/message isn't better. Apps using the SDK now have name collisions:
> 
> `import { Message } from "@inlang/sdk"`
> 
> What message is imported here?
>
> Recommendation:
>- Release V2 prerelease of @inlang/message instead of having a new package
>- Suffix all types with V2 e.g. MessageV2, PatternV2, etc. to avoid name collisions
>
> (We will run into this problem again with a potential v3. Seems easier to then again suffix MessageV3 instead of having a new package and 3 potential name collisions)

@samuelstroschein that' s a good catch!

I was hoping that the new `Message` type would automatically be namespaced under the `MessageBundle` module  but this is not the pattern that @inlang/sdk uses [here](https://github.com/opral/monorepo/blob/d1879fda6d736a778410d68ee88b98649386c7f7/inlang/source-code/sdk/src/index.ts#L31) 

#### ideas:
1. follow your suggestion of new suffixed types like `MessageV2` etc.
2. export new types like Message and MessageBundle ONLY as part of the `project.query2` api, as proposed in [MESDK-83](https://linear.app/opral/issue/MESDK-83/sdk-api-for-messagebundle-with-variants#comment-69373f50)      
3. differentiate `Message` from `{namespace}.Message` as proposed [originally](https://github.com/opral/monorepo/pull/2655/files#diff-a42f4005a5fc991ab0d4fc39b301e7b78b0646962c11fbc2d77053a9002dc8b8R2) with `AST.Message` as the prefix. 
4. deprecate the @inlang/sdk root level types, and import new types only from versioned-interface packages E.g. `import { Message } from @inlang/message-bundle`
5. namespace under `sdk/*` e.g. `import { Message } from @inlang/sdk/message-bundle`

samuelstroschein:
@jldec 

1. i propose to suffix the new interfaces with `V2` and remove the suffix once we release the SDK in a new major version similar to https://github.com/opral/inlang-message-sdk/issues/45#issuecomment-2096282401 
2. https://linear.app/opral/issue/MESDK-84/get-rid-of-versioned-interface-packages triggered by this discussion

inlang/source-code/versioned-interfaces/message-bundle/src/interface.ts
jldec: Yes, we want to be able to import MF2 using this, right?

jldec:
@martin-lysk any opinion on this?

jldec:
@samuelstroschein ok - i'll add MessageBundleV2 and MessageV2 to sdk.
(no need for a new versioned-interface package, right)

inlang/source-code/versioned-interfaces/message-bundle/src/interface.ts
samuelstroschein: Why do we want to import MF2 for the MVP? 

- nobody uses MF2
- let's ship only what we need to unblock apps namely (variants) 
- ship in a way to make the data structure future proof

MF2 is still changing. anything we don't need i wouldn't introduce now

samuelstroschein:
> @jldec (no need for a new versioned-interface package, right)

correct.

inlang/source-code/versioned-interfaces/message-bundle/src/interface.ts
samuelstroschein: Here still seems to be some unalignment. Let's achieve consensus @jldec @LorisSigrist @martin-lysk. 

I wouldn't risk adding features from MF2 that apps don't need right now to support variants (pluralization, gendering) as they still change. I don't know why the need to import MF2 is recurringly coming up. We don't need to import MF2 just yet. Nobody uses MF2.

jldec: Editing nested functions without locals also feels harder from UI/UX perspective, that doing it with locals. Do you agree @NilsJacobsen ?

samuelstroschein: I would go step by step to enable pluralization first. Markup, for example, is unrelated to pluralization. But @jldec this will be your decision now. Being compatible with full MF2 from the get-go is in wrong IMO. The spec will change and we have different requirements because we have apps. 

What is required for pluralization? 

- MessageBundle
- Variants
- Pluralization function 

Seems not to be required: 

- Markup
- Other functions (like date formatting, gendering etc.)
- local declarations

samuelstroschein: Sync notes https://linear.app/opral/issue/MESDK-77/requirements-for-gender-and-plurals#comment-0dd412be. 

It seems like local declarations can be removed if we go the bottoms up approach. cc @martin-lysk

jldec: 👍 Removed for now.

jldec:
@samuelstroschein 
I have moved the new types into @inlang/sdk/v2 but kept the type names unchanged.
To avoid conflicts, it's easy to alias the imported type name individually, or `import * as V2`
The types are not exported at the root of the sdk.

inlang/source-code/sdk/src/v2/types.ts
```diff
@@ -0,0 +1,122 @@
+import { LanguageTag as Locale } from "@inlang/language-tag"
```
samuelstroschein, line 1: You can pull the Locale type into the types here. 

- language tag has been it's own package to allow the community to use it 
-- this is not happening
- eases getting rid of versioned-interfaces

samuelstroschein (approved):

PR #2734, opened 2024-05-08
MESDK-87 Add delegate pattern for MessagesQuery<->LintReportQuery communication

martin-lysk:
resolves https://github.com/opral/inlang-message-sdk/issues/49


inlang/source-code/sdk/src/createMessageLintReportsQuery.ts
```diff
@@ -42,81 +38,65 @@ export function createMessageLintReportsQuery(
...
+		onMessageDelete: (messageId: string) => {
+			index.delete(messageId)
+		},
+	}
+
+	messagesQuery.setDelegate(messageQueryChangeDelegate)
 
 	return {
 		getAll: async () => {
+			// TODO reintroduce reactivity here - wont be the bottleneck
```
jldec, line 92: I wouldn't spend more time on this TODO - let's remove it.

inlang/source-code/sdk/src/createMessageLintReportsQuery.ts
```diff
@@ -42,81 +38,65 @@ export function createMessageLintReportsQuery(
...
 			await sleep(0) // evaluate on next tick to allow for out-of-order effects
 			return structuredClone(
 				[...index.values()].flat().length === 0 ? [] : [...index.values()].flat()
 			)
 		},
 		get: async (args: Parameters<MessageLintReportsQueryApi["get"]>[0]) => {
+			// TODO reintroduce reactivity here
```
jldec, line 99: same - no need to re-introduce reactivity.

inlang/source-code/sdk/src/createMessagesQuery.ts
```diff
@@ -62,6 +62,13 @@ export function createMessagesQuery({
 	// filepath for the lock folder
 	const messageLockDirPath = projectPath + "/messagelock"
 
+	// TODO check if we should use a weak ref instead
```
jldec, line 65: I'd prefer not to complicate the code with weak refs unless really needed.

jldec (approved):
The surprise (bad) is that with a 10k load-test without translation but with watch turned on which keeps the process alive (`pnpm test 10000 0  1 1 1`) the process eventually dies with an out of memory crash 💥. 

I don't understand why the delegate needs so much more memory when we're no longer creating nested effects. 
 
Once I introduced the fix from #2738, the process no longer crashes, and total memory usage drops to 380MB which is much better (before it would stay pegged at about 2GB). I was even able to load 20k messages, and then translate them.

I pushed a commit to add the sleep during load.

I also filed https://github.com/opral/inlang-message-sdk/issues/54 and https://github.com/opral/inlang-message-sdk/issues/53 for important issues found during the investigation.

jldec:
See also https://github.com/opral/inlang-message-sdk/issues/57

PR #2738, opened 2024-05-08
Fix solid memory leak

martin-lysk:
This change schedules message loading to the next tick on every 500th message to give solid a tick to free resources.

https://github.com/solidjs-community/solid-primitives/blob/9ca76a47ffa2172770e075a90695cf933da0ff48/packages/trigger/src/index.ts#L64


inlang/source-code/sdk/src/createMessagesQuery.ts
```diff
@@ -339,6 +344,7 @@ async function loadMessagesViaPlugin(
 				messages.set(loadedMessageClone.id, loadedMessageClone)
 				// NOTE could use hash instead of the whole object JSON to save memory...
 				messageState.messageLoadHash[loadedMessageClone.id] = importedEnecoded
+				loadedMessageCount++
```
jldec, line 347: You're bumping the count in 2 code paths but only testing for `loadedMessageCount % messagesPerTick === 0`  in one of those paths. This makes it hard to know if the check will actually be hit.

Why not test for > maxMessagesPerTick and then reset the count?

jldec (changes requested):
see comment.

inlang/source-code/sdk/src/createMessagesQuery.ts
```diff
@@ -365,6 +371,13 @@ async function loadMessagesViaPlugin(
 				// we don't have to check - done before hand if (messages.has(loadedMessageClone.id)) return false
 				messages.set(loadedMessageClone.id, loadedMessageClone)
 				messageState.messageLoadHash[loadedMessageClone.id] = importedEnecoded
+				loadedMessageCount++
+			}
+			if (loadedMessageCount % messagesPerTick === 0) {
+				// move loading of the next messages to the next ticks to allow solid to cleanup resources
+				// solid needs some time to settle and clean up
+				// https://github.com/solidjs-community/solid-primitives/blob/9ca76a47ffa2172770e075a90695cf933da0ff48/packages/trigger/src/index.ts#L64
+				await 0
```
jldec, line 380: I think I'd prefer to use a setTimeout here.

jldec: I verified that the behavior of sleep(0) with `setTimeout(...,0)` is not the same as `await 0`

jldec:
Closing this PR - code is included in #2734

PR #2742, opened 2024-05-10
Open in Fink link does not include branch | Sherlock 🕵️‍♂️

felixhaeberle:
fix SHERL-25

cc @janfjohannes can you please check why `test-symlink` isn't recognized as return value for `repo.getCurrentBranch()` in https://github.com/opral/monorepo/blob/f6b654bec37dd443fd13c2edfe2702330e52cec6/inlang/source-code/ide-extension/src/utilities/settings/getCurrentBranch.test.ts#L36


janfjohannes:
@felixhaeberle     seems to be unrelated to the branch thing but a problem with finding the repo root, let me know if that helps or i should debug/ fix the pr directly

```
stderr | src/utilities/settings/getCurrentBranch.test.ts > getCurrentBranch > should return undefined when no repository root is found
  Failed to find repository root.
```

felixhaeberle:
@janfjohannes I basically copied the test setup from the test you wrote so it would be cool if you could look into it with the mocked repo state.

janfjohannes:
@felixhaeberle  looks like just your mocking setup was bad, please try to avoid mocks like hell.

PR #2751, opened 2024-05-12
OPS-22 Use latest vite, vitest, @vitest/coverage in all packages

jldec:
This is an experiment to see how easily we can update vite and vitest.
The versions of those packages have to set to the latest available.

This branch does not currently build because of incompatibilities in @inlang/manage.
I did not pursue this further for now (perhaps we can retire the manage package and save ourselvs the effort)


jldec:
Closing because it wasn't straightforward for me to bump vite and vitest.

PR #2754, opened 2024-05-13
PARJS-100 Support for compiling variants

LorisSigrist:
This PR adds support for compiling the new V2 Message AST.

Types aren't yet generated since that depends on the Registry discussion


samuelstroschein:
Close in favor of #3057 ?

PR #2758, opened 2024-05-13
Change load-test to write to temp file and then rename

jldec:
See https://github.com/opral/inlang-message-sdk/issues/53 (MESDK-91)

To repro, build the sdk, then go to the load test directory and run `pnpm test 10 0 1 0 1`. This generates 10 messages, no translation, subscribed to messages.getAll() events, no lint reports, and with watch mode on to prevent the project from exiting. 

```
cd /inlang/source-code/sdk/load-test
pnpm test 10 0 1 0 1
```

Normally using an editor to manually overwrite the messages file in `project.inlang/messages.json` will trigger additional message events (marked with pink below) in the load-test console output on each change. 

With the changes in this PR, those events are not being triggered.

Screenshot 2024-05-13 at 08 45 06

### DIFF
--- a/inlang/source-code/sdk/load-test/.gitignore
+++ b/inlang/source-code/sdk/load-test/.gitignore
@@ -2,3 +2,4 @@ node_modules
 locales
 project.inlang/messages
 project.inlang/messages.json
+project.inlang/temp.json
--- a/inlang/source-code/sdk/load-test/load-test.ts
+++ b/inlang/source-code/sdk/load-test/load-test.ts
@@ -127,6 +127,7 @@ export async function runLoadTest(
 // inlang message format
 async function generateMessageFile(messageCount: number) {
 	if (isExperimentalPersistence) {
+		const tempFile = join(__dirname, "project.inlang", "temp.json")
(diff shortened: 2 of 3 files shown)


lucidNTR:
@jldec  fs.rename is not part of the nodeishfs subset that lix supports and cannot be used. please remove and create instead.

jldec:
@lucidNTR the load-test data generator is not really a lix app, just a tool to generate test data files in the file system.

Does this mean that lix fs cannot watch files owned (and potentially renamed) by 3rd party, non-lix-aware, applications like shell scripts?

janfjohannes:
@jldec  in those cases it is just the node.js fs implementation, not the memoryfs and lix has nothing to do with the watcher, it behaves however nodejs is working and nodejs is highly inconsistent between nodejs versions and also between different platforms. there are known issues with rename, remove, subfolders and many others. this will only fully be fixed when we swtich to lix.fs  which completely replaces the watch implementation with a full folder scan/hashing diff algorythim that is predictable and behaves the same in all environments

jldec:
OK - thanks @janfjohannes - i'll find a different strategy to write the test data.

PR #2773, opened 2024-05-14
Add a `createMessageBundle` utility function

LorisSigrist:
This PR adds a `createMessageBundle` utility function to the SDK.

Example usage
```ts

const bundle = createMessageBundle({
  id: "happy_elephant",

  //optionally define aliases
  alias: {
    "default": "hello_world"
  },
  
  // define a map of messages for all the languages you want. 
  // ICU MF 2 strings are used
  // Throws error if the ICU string cannot be represented using our AST
  messages: {
    en: ".input {$name} {{Hello {$name}}}",
    de: ".input {$name} .match {$name} World {{Hallo Welt}} * {{Hallo {$name}}}",
  },
})
```

Exported from `@inlang/sdk/test-utilities`

### DIFF
--- a/inlang/source-code/sdk/package.json
+++ b/inlang/source-code/sdk/package.json
@@ -50,6 +50,7 @@
 		"debug": "^4.3.4",
 		"dedent": "1.5.1",
 		"deepmerge-ts": "^5.1.0",
+		"messageformat": "4.0.0-7",
 		"murmurhash3js": "^3.0.1",
 		"solid-js": "1.6.12",
 		"throttle-debounce": "^5.0.0"
--- a/inlang/source-code/sdk/src/test-utilities/createMessageBundle.ts
+++ b/inlang/source-code/sdk/src/test-utilities/createMessageBundle.ts
@@ -1,0 +1,117 @@
+import * as MF from "messageformat"
+import type * as AST from "../v2/types.js"
+
+/**
+ * Utility function to generate message bundles for tests
+ *
+ * @throws if the given source-strings aren't valid or not representable
+ */
+export function createMessageBundle(bundleOptions: {
+	id: string
+	aliases?: Record<string, string>
+	/**
+	 * A map of languages to ICU MessageFormat 2 message strings
+	 */
+	messages: Record<string, string>
+}): AST.MessageBundle {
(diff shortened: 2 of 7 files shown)


inlang/source-code/sdk/src/test-utilities/createMessageBundle.ts
```diff
@@ -0,0 +1,117 @@
+import * as MF from "messageformat"
+import type * as AST from "../v2/types.js"
+
+/**
+ * Utility function to generate message bundles for tests
+ *
+ * @throws if the given source-strings aren't valid or not representable
```
jldec, line 7: Could you add a test for the throws-on-error case?

inlang/source-code/sdk/package.json
```diff
@@ -50,6 +50,7 @@
 		"debug": "^4.3.4",
 		"dedent": "1.5.1",
 		"deepmerge-ts": "^5.1.0",
+		"messageformat": "4.0.0-7",
```
jldec, line 53: should we pin to this version or use a ^ ?

jldec (approved):
LGTM - 1 additional test for error case requested, and question about dep. version.

inlang/source-code/sdk/package.json
LorisSigrist: Let's leave it pinned for now. We don't really care about future updates, we aren't trying to replicate MF2 1-to-1. We just want to express our AST more concisely. 
Unless we find a serious bug in the package I don't see us ever needing to update.

jldec (changes requested):
hold pending resolution of [internal discussion](https://discord.com/channels/897438559458430986/1239884337172844564/1239933431681519688)

jldec (changes requested):
@LorisSigrist 
Instead of introducing a dependency on the mf2 "messageformat" package, let's add some v2/Message type ("AST") mocks and a v2 `createMessage()` function into sdk/v2/test-utilities (v2 mirror of sdk/test-utilities)

inlang/source-code/sdk/package.json
jldec: better to remove dependency entirely.

PR #2793, opened 2024-05-15
create mock asts for message bundle component | Message Component

felixhaeberle:
- fix INLMC-88


felixhaeberle:
@jldec It would be nice if you could have a look at the mock message format to verify if we are going in the right direction here and add a additional mock message `mockMessageBundleMinimal` with the minimal number of properties.

jldec:
Thanks @felixhaeberle - your mock looks good to me.
I just added you to also review https://github.com/opral/monorepo/pull/2795 which should help create simple mocks more easily. Using the new structure in a test project is on the roadmap ([MESDK-104](https://github.com/opral/inlang-message-sdk/issues/65)), but we first have to be able to persist messages using the new structure.

jldec:
@felixhaeberle 
I think we should put these mocks into `sdk/v2/mocks` instead, so that they can be re-used for testing etc across apps.
Please see https://github.com/opral/monorepo/pull/2795#issuecomment-2115803266

jldec:
@felixhaeberle - I merged a slightly simplified and corrected [plural example](https://github.com/opral/monorepo/blob/main/inlang/source-code/sdk/src/v2/mocks/plural/bundle.ts) message bundle ast with #2795

NilsJacobsen:
@jldec you can start making the review.

inlang/source-code/sdk/src/v2/mocks/multipleMatcher/bundle.ts
```diff
@@ -0,0 +1,188 @@
...
+						},
+					},
+				},
+			],
+			selectors: [
+				{
+					type: "expression",
+					arg: {
+						type: "variable",
+						name: "count: plural",
```
jldec, line 29: The plural function is an annotation to the variable expression
```js
{
  type : "expression",
  arg: { type : "variable", name: "count"},
  annotation: { type: "function", name: "plural", options: [] }
}
```

inlang/source-code/sdk/src/v2/mocks/multipleMatcher/bundle.ts
```diff
@@ -0,0 +1,188 @@
...
```
jldec, line 83: should be annotation - see above.

inlang/source-code/sdk/src/v2/mocks/multipleMatcher/bundle.ts
```diff
@@ -0,0 +1,188 @@
...
+					},
+				},
+				{
+					type: "expression",
+					arg: {
+						type: "variable",
+						name: "formal: bool",
```
jldec, line 90: bool should be an annotation (function) as well right?
Do you know how this function would be defined ?
I could not find it in https://github.com/unicode-org/message-format-wg/blob/main/spec/registry.md
cc: @martin-lysk @LorisSigrist for clarification of how the "formal" selector is expected to work.

jldec: This might also affect how the formal selectors are expressed in other parts of the AST.

jldec (changes requested):
changes needed for function annotations and clarification needed for "formal".

inlang/source-code/sdk/src/v2/mocks/multipleMatcher/bundle.ts
jldec: from here
https://github.com/opral/monorepo/pull/2795#pullrequestreview-2062588832

NilsJacobsen: Done

inlang/source-code/sdk/src/v2/mocks/multipleMatcher/bundle.ts
NilsJacobsen: Done

inlang/source-code/sdk/src/v2/mocks/multipleMatcher/bundle.ts
NilsJacobsen: I did it like the plural. Can we make a separate discussion about it, so we can merge this PR.

jldec (approved):
LGTM thanks @NilsJacobsen

PR #2795, opened 2024-05-15
MESDK-103 - v2 createMessageBundle() and example mocks

jldec:
This is an alternative implementation of #2773 (closed) which used MF2 string format internally.

It includes only simple placeholder pattern parsing so that we can use to migrate existing projects while we work on the v2 persistence design.

Resolves https://github.com/opral/inlang-message-sdk/issues/64

- [x] createMessageBundle()
- [x] example mocks
- [x] createMessage() with either text or pattern
- [x] basic parsing of variable placeholders in text (same as plugins)
- [x] automatic creation on input declarations based on the union of variable placeholders   

NOTE: Input variable declarations and variables in patterns are now wrapped inside an [Expression](https://github.com/opral/monorepo/blob/main/inlang/source-code/sdk/src/v2/types.ts#L64) object which adds more nesting to the ast structure. (see test for example)

**Example**
```ts
createMessageBundle({
  id: "hello_world_2",
  messages: [
    createMessage({ locale: "en", pattern: "Hello, {name}!" }),
    createMessage({ locale: "de", pattern: "Hallo, {name}!" }),
  ],
})

// pass text to CreateMessage for messages with no {variable} placeholders
createMessageBundle({
  id: "hello_world",
  messages: [createMessage({ locale: "en", text: "Hello, World!" })],
})
```


inlang/source-code/sdk/src/v2/createMessageBundle.ts
```diff
@@ -0,0 +1,121 @@
...
+	Expression,
+} from "./types.js"
+
+export function createMessageBundle(
+	id: string,
+	localeText: Record<LanguageTag, string>,
+	options?: {
+		alias?: MessageBundle["alias"]
+		variableReferencePattern?: string[]
+	}
```
jldec, line 17: Q: should we switch to passing an args object instead of regular function params?

samuelstroschein:
Why not pass messages to `createMessageBundle`? 

instead of 

```
const bundle: unknown = createMessageBundle(
  "hello_name",
  {
    // pass multiple messages using locale as the key here. 
    en: "Hello, {name}!",
  },
  { variableReferencePattern: ["{", "}"] }
)
```

have this

```
const bundle: unknown = createMessageBundle({
  id: "happy_human_elephant",
  locales: [
    createMessage({ locale: "en", pattern: "Hello, {name}!" })
    createMessage({ locale: "de", pattern: "Hallo, {name}!" })
  ]
})
```

LorisSigrist:
Why are we suddenly using a **serialized pattern** here? We dismissed using ICU MF2 to avoid forcing devs to learn a serialized format.  
- Which Format are we using in `createMessage`? 
- How is this different / better than ICU2?

felixhaeberle:
@jldec Could you address the above question in a LOOM? I agree with Loris&Samuel here on the observations.

inlang/source-code/sdk/src/v2/createMessageBundle.ts
jldec: 👍 changed.

jldec:
@samuelstroschein @LorisSigrist 

This is not intended to provide an alternative string format for messages. 

My intent was to help us work on v2 persistence with existing project data (like the monorepo website), and to start experimenting with utility functions to manipulate v2 MessageBundles without having to operate directly on the AST.

jldec:
@samuelstroschein 

> Why not pass messages to createMessageBundle?

🙏 i like that better. I have modified the signatures to use an args object and pass an array Message objects into createMessageBundle as you suggested.

> variableReferencePattern is redundant

I was thinking we might use that for migrating messages from existing i18next projects, but we can revisit that later. For now only {var} is supported.

> [args] heuristic: more than two arguments

👍 changed

jldec:
@felixhaeberle 
> Could you address the above question in a LOOM

I'll do a loom as soon I can actually demo something useful 

The goal is supprt v2 MessageBundles with persistence in a way which is similar  to what I demo'ed in https://www.loom.com/share/9f6e47997a784728a492c17e00ee030b (NOTE that video still shows persitence with the legacy internal Message ast)

samuelstroschein:
> @LorisSigrist  Why are we suddenly using a serialized pattern in createMessage?

I was wondering about this, too, and thought you and @jldec dismissed my arguments against a serialized format. 

I am asking again if a serialized format is needed now/if the utilities are even needed. Isn't it OK if you manually type some messages for your tests for now @LorisSigrist until we get clarity on what "good" utilities will look like and if we even need them?

LorisSigrist:
> I was wondering about this, too, and thought you and @jldec dismissed my arguments against a serialized format.

I'm on your side here, I was also surprised to see a serialized format. 

>  Isn't it OK if you manually type some messages for your tests for now

Yes, writing the AST _fine_ for the moment. 

In the long term it will become a maintenance burden IMO. It's a bit like writing half your tests in byte code. It's the wrong level of abstraction for a lot of the tests. Many of paraglide's tests are "take this message, run it through the complier & see if it renders correctly". Introducing an abstraction here will eventually be necessary, but can wait until we have more experience. 

For now writing the AST is _fine_.

martin-lysk:
I can’t follow the reasoning here.

martin-lysk:
Wow wrong button during scrolling on mobile - sorry for the ping. 

I am already afk -  I can have a Quick Look at the ast tomorrow. 

Reagarding the discussion: 
I think I made my points for an api using parsed strings based on an existing bases (icu2) already. 
I understand the upsides of having a textual format in favour of an unreadable json inside of test you @LorisSigrist and you @jldec referring. 
But introducing our own parsing and format as input seems like the wrong direction. 

If we want an interims string based api - we should use current state of icu2. If not it’s an object based typed api - that was at least the two path we discussed (?) @samuelstroschein ?

samuelstroschein:
> If we want an interims string based api - we should use current state of icu2. If not it’s an object based typed api - that was at least the two path we discussed (?) @samuelstroschein ?

Yes. I wouldn't cook our own serialized format and I am hesitant to introduce unstable icu2 now. Hence, delay all those utility functions until either or: 

1. we can't live without utilities anymore
2. icu2 is stable

jldec:
ok, parsePattern is gone - I'll get by without the feature for now.

inlang/source-code/sdk/src/v2/mocks/messageBundles.ts
```diff
@@ -0,0 +1,152 @@
...
+						},
+					},
+				},
+			],
+			selectors: [
+				{
+					type: "expression",
+					arg: {
+						type: "variable",
+						name: "count: plural",
```
LorisSigrist, line 29: I don't think this expression is what you meant, this literally reads as "a variable called `count: plural`"

I think you wanted to apply the `plural` function to the variable `count`. 

The way to express this is by adding a function annotation to the expression:

```ts
{
  type : "expression",
  arg: { type : "variable", name: "count"},
  annotation: { type: "function", name: "plural", options: [] }
}
```

inlang/source-code/sdk/src/v2/mocks/messageBundles.ts
```diff
@@ -0,0 +1,152 @@
...
+					},
+				},
+			],
+			variants: [
+				{
+					match: ["one"],
+					pattern: [
+						{
+							type: "text",
+							value: "Show {count} message.",
```
LorisSigrist, line 39: The `value`: `"Show {count} message."` doesn't treat "{count}" as a variable, but as plain text. 

The way to mix in variables into the AST is by adding `expression`s to the pattern.

```ts
pattern: [
  { type: "text", value: "Show "},
  { type: "expression", arg: { type: "variable", name: "count" }},
  { type: "text", value: " message."}
]
```

martin-lysk: +1

inlang/source-code/sdk/src/v2/mocks/messageBundles.ts
felixhaeberle: That was my fault 😅 I naively created a mock message (which @jldec copied) without respecting the correct structure.

inlang/source-code/sdk/src/v2/mocks/messageBundles.ts
felixhaeberle: Same ⤴️

felixhaeberle:
I'll take myself out of the review loop here. Nothing to add besides @samuelstroschein & @LorisSigrist already valuable comments.

samuelstroschein:
I am also taking myself out of this discussion now. @LorisSigrist @jldec @martin-lysk you'll make the right choice

jldec:
Thanks for the feedback and suggestions.
To move things along, I will merge this now with a simplified example bundle in sdk/src/v2/mocks/plural

@felixhaeberle, @LorisSigrist, @martin-lysk please feel free to add additional examples using a similar directory naming pattern.

PR #2797, opened 2024-05-16
Adds a utility function to debug pack responses coming from git server

martin-lysk:


lucidNTR:
@martin-lysk  does not work and responseBody: any is not really a helpful type to know how this can be used 🗡️

lucidNTR:
update: i got it to work. what is missing is detection if something even is a pack file, as this has to be added to any response body, we need to know how to know if something is a pack file body

lucidNTR:
also please add a bit of explanation to: what is nak, what is ack, what is shallows what is packfile fifo, why is there progress and what does it mean,

lucidNTR:
update: i use valid extraction of packfile oid as detecion which seems to work well enough for now.

lucidNTR:
update: finally got everything running i needed and answered my own questions!

lucidNTR (approved):

PR #2800, opened 2024-05-17
Trigger signals and callbacks for messages not present in messages provided by loadMessages plugin

martin-lysk:
fixes Reactivity for messages that are deleted
adds a test for deleted and created messages


martin-lysk:
@jldec the fix was rather small - we just need to go through the messages loaded already and substract the new messages coming from the plugin. The messages left are those that are deleted. 

I remember that we discussed this when we worked on the new persistence - where messages that are not present in the set of imported messages does not mean a message was deleted - it was just not in the source used by the importer. 
Until we deprecate loadMessagesPlugins in favor of importer this assumption doesn't hold and its clearly a bug. 

I also added a test for message creation that was missing as well.

inlang/source-code/sdk/src/loadProject.test.ts
```diff
@@ -1145,11 +1162,25 @@ describe("functionality", () => {
...
 			await new Promise((resolve) => setTimeout(resolve, 200)) // file event will lock a file and be handled sequentially - give it time to pickup the change
 
 			expect(counter).toBe(3)
+
+			// change file - add a message
+			messages.data.push(newMessage)
+			await fs.writeFile("./messages.json", JSON.stringify(messages))
+			await new Promise((resolve) => setTimeout(resolve, 200)) // file event will lock a file and be handled sequentially - give it time to pickup the change
+
+			expect(counter).toBe(4)
```
jldec, line 1176: In this case, instead of just checking the event counter, I think the test should check the number of messages returned from `inproject.query.messages.getAll.subscribe(()`.

jldec (changes requested):
LGTM - just one thing needed in the test.

LorisSigrist (approved):
Thanks!

inlang/source-code/sdk/src/loadProject.test.ts
martin-lysk: test added

martin-lysk:
I just found a bug in fsWatch cleanup - fixed it and also added the test to probe for inserts and updates

inlang/source-code/sdk/src/loadProject.test.ts
jldec: much better -  thank you.

PR #2801, opened 2024-05-18
MESDK-88 Minimal v2 persistence

jldec:
This is step 1 toward [full persistence and apis for v2 MessageBundles](https://github.com/opral/inlang-message-sdk/issues/70) and resolves https://github.com/opral/inlang-message-sdk/issues/50.

The (bare-bones) [store](https://github.com/opral/monorepo/pull/2801/files#diff-490b75f6e01e8c5e24c7dde1427e5b929c59fc0296072d6da90f4c20d93ecb95) dumps messageBundles into a single JSON file.  

The goal of this PR is to be able to run the load-test and multi-project-test with v2 MessageBundles persisted to disk using the experimental persistence feature flag.

- open a message store via loadProject
- separately run `inlang machine translate`, adapted to a new CRUD api.
- check the results, also using a new CRUD api.

Only the minimum read/write CRUD api necessary for the above will be included in this PR. This api should be considered internal / temporary.

- [x] load/store JSON v2 MessageBundles under experimental persistence feature flag
- [x] hook into loadProject to return a `project.store.messageBundles` [CRUD api](https://github.com/opral/monorepo/blob/v2-persistence/inlang/source-code/sdk/src/persistence/storeApi.ts).
- [x] stub out the existing messages `project.query.*` api
- [x] implement minimal store api for v2 MessageBundles
- [x] minimal store api tests
- [x] reintroduce helper from [#2655](https://github.com/opral/monorepo/pull/2655/files#diff-9e26e34447efe65d95d3323b2a8620c27f9d4e7fb83829a80070d9a0a622c739) to convert between v1 Message and v2 MessageBundle
- [x] adapt `inlang machine translate` to call the store api behind experimental persistence flag.
- [x] adapt load-test to support experimental persistence 
- [x] review comments
- [x] batch writes when there are large numbers of updates
  - [x] durable update api: use batchedIO to await until fs write completes
  - [x] test for batchedIO 
- [x] load-test with 10k messages 
- [x] experimental json with slots and newlines for improved merge conflict avoidance
- [x] create issues for TODOs
- [x] changeset

#### to test
$ `DEBUG=sdk:store,sdk:fileLock,sdk:batchedIO pnpm --filter @inlang/sdk-load-test test`

#### not included in this PR (but planned in future PRs)
- file per MessageBundle  - https://github.com/opral/inlang-message-sdk/issues/84
- lazy reads - https://github.com/opral/inlang-message-sdk/issues/83
- query builder api - https://github.com/opral/inlang-message-sdk/issues/82
- subscribe api - https://github.com/opral/inlang-message-sdk/issues/81

> [!NOTE] 
> Because the existing messagesQueryApi is stubbed out, existing apps will no longer work in projects with the experimental persistence feature flag turned on.


jldec:
Tests are now green, however dumping all messages into a single JSON file _on every update_ is not a tenable design even for this first PR. When there a multiple writers, the large-file save lock-contention eventually produces errors (see below).

### options
1. use mutliple files e.g per message bundle or by language
2. batch writes (like we do in v1)

cc: @janfjohannes @martin-lysk @samuelstroschein 
---
```
$ pnpm --filter @inlang/sdk-load-test test

  load-test load-test start - using experimental persistence +0ms
  load-test generating 1000 messages +0ms
  load-test opening repo and loading project +2ms
Using existing cloned repo
  load-test subscribing to project.errors +6s
  load-test loaded 1000 v2 MessageBundles +0ms
  load-test translating messages with inlang cli +0ms
Using existing cloned repo

 ERROR   save exceeded maximum retries (10) to acquire lockfile 11                                            2:31:13 PM

  at acquireFileLock (/Users/jldec/opral/monorepo/inlang/source-code/cli/dist/main.js:74839:11)
  at Timeout._onTimeout (/Users/jldec/opral/monorepo/inlang/source-code/cli/dist/main.js:74904:33)

  load-test load-test done - exiting +9s
```

janfjohannes:
@martin-lysk  @jldec  my last status was that you use many small files with a locking mechanism for persistnece. what changed since then?

samuelstroschein:
go for option 1: split by bundle (we have to do this anyways)

inlang/source-code/sdk/src/loadProject.ts
```diff
@@ -92,9 +101,13 @@ export async function loadProject(args: {
...
-				.then((settings) => setSettings(settings))
+				.then((settings) => {
+					setSettings(settings)
+					markSettingsAsLoaded()
+				})
 				.catch((err) => {
 					markInitAsFailed(err)
+					markSettingsAsFailed(err)
 				})
 		})
```
jldec, line 112: @martin-lysk can you confirm that this effect is not really doing anything and could be removed, allowing the loadSettings to be awaited?

inlang/source-code/sdk/src/loadProject.ts
```diff
@@ -106,11 +119,7 @@ export async function loadProject(args: {
...
 			try {
 				const validatedSettings = parseSettings(settings)
-				if (validatedSettings.experimental?.persistence) {
-					settings["plugin.sdk.persistence"] = {
-						pathPattern: projectPath + "/messages.json",
-					}
-				}
+				v2Persistence = !!validatedSettings.experimental?.persistence
 
 				batch(() => {
```
jldec, line 124: @martin-lysk could you clarify the intent of wrapping the next two calls inside a batch()?

jldec:
> go for option 1: split by bundle (we have to do this anyways)

@samuelstroschein could you clarify the reason why you think we have to do this?
I understand in the short term it would be nice to use separate files to help git to avoid merge conflicts, but once we have sqlite support, lix will have to do this differently right? 

I will look at both options to see which will unblock us to ship this PR soonest. File-per-message has more complexity which may not be justified given our immediate goal of shipping the v2 persistence api for apps asap. 

> my last status was that you use many small files with a locking mechanism for persistnece. what changed since then?

@janfjohannes, the POC for new sdk persistence with a file-per-message was removed from [PR 2108](https://github.com/opral/monorepo/pull/2108) in order to limit the scope of that PR and address data-loss issues first. The plan to build [bidirectional data-sync](https://discord.com/channels/897438559458430986/1200092725341597797/1200393781975982153) between plugins and sdk files was [simplified](https://github.com/opral/monorepo/pull/2108#issuecomment-1912284172) to ship the new persistence separately, behind a feature flag, which is what we're doing in this PR.

The locking mechanism currently in place is a global (project-level) lock. We never implemented per-file locking. 
@janfjohannes - b.t.w. it would be nice to have a simpler way to do atomic file writes through lix fs - e.g. by writing to a temp file and renaming.. That would allow us to avoid at least some of the complextity of the current lockfile with stale-lock detection.

inlang/source-code/sdk/src/loadProject.ts
martin-lysk:
```
batch(() => {
	// reset the resolved modules first - since they are no longer valid at that point
	setResolvedModules(undefined)
	_setSettings(validatedSettings)
})
```

when setSettings is called with a new settings object the call to _setSettings will trigger an effect - modules that depending on settings and will load the modules asynchrouns will refelect the old state at this point.

If something like  createMessagesQuery has an effect on settings AND on resolve modules it would get triggered with an invalid state. We call _setSettings and setResolvedModules in a batch to make sure the state is reflected in the effect. 

The execution call with the batch looks like this:

1. setSettings is called and the new settings object is set and the modules are reset (not loaded yet) and the module load is triggered https://github.com/opral/monorepo/blob/ae47203e54f1e0f41d0c25bff7aafc3cf90a1cc5/inlang/source-code/sdk/src/loadProject.ts#L145
2. an effect in messageReportQuery is triggered for the settings. it will return early since the modules are not ready yet
3. the modules get loaded and an effect on resolvedModules is triggerd
4. another effect in messageReportQuery is triggered - this time with settings and resolved modules

inlang/source-code/sdk/src/loadProject.ts
martin-lysk: good catch - this effect doesn't have a reference to a signal.
Context: i guess this one was meant to watch on an effect produced by a file watcher on the settings file in the TODO the line below.

martin-lysk: Shall we keep the effect and introduce settings file watching instead or introduce settings reactivity with the new sdk?

janfjohannes:
i think a good batching algorithm is needed in any case and would always be my first step before going into the other optimizations.

inlang/source-code/sdk/src/loadProject.ts
jldec: no, let's not fix this now - it will adds more complexity to watch files and differentiate between file-sourced vs. api-sourced changes.

#### notes from our 1:1 discussion:
- the planned intent (as mentioned in the TODO) was to watch the fs for settings changes but this was never implemented.
- awaiting and immediately throwing an error on load if the settings are invalid would make the code simpler. @jldec prefers this approach because it would simplify the startup logic 
- we need to handle runtime changes (after initial load) for settings as well, including error cases.

inlang/source-code/sdk/src/persistence/storeApi.ts
```diff
@@ -0,0 +1,16 @@
...
+	messageBundles: Query<V2.MessageBundle>
+}
+
+/**
+ * WIP template for async V2 crud interfaces
+ * E.g. `await project.messageBundles.get({ id: "..." })`
+ **/
+export interface Query<T> {
+	get: (args: { id: string }) => Promise<T | undefined>
+	set: (args: { data: T }) => Promise<void>
```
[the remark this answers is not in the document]
martin-lysk: The api of rxdb looks pretty clean and carefully drafted [see](https://rxdb.info/rx-query.html#update) I use this as an inspiration

inlang/source-code/sdk/src/persistence/filelock/acquireFileLock.ts
```diff
@@ -21,16 +21,17 @@ export async function acquireFileLock(
 	try {
 		debug(lockOrigin + " tries to acquire a lockfile Retry Nr.: " + tryCount)
 		await fs.mkdir(lockDirPath)
-		const stats = await fs.stat(lockDirPath)
+		const stats = await fs.stat(lockDirPath) // not atomic with mkdir, but good enough
```
martin-lysk, line 24:
```suggestion
		// NOTE: fs.stat does not need to be atomic since mkdir would crash atomically - if we are here its save to consider the lock held by this process
		const stats = await fs.stat(lockDirPath) 
```
I agree a comment on this is needed to clearify the reasoning and i added a suggestion. e should understand why this is good enough - if the comment doesn't help let me know so we can be more precise

inlang/source-code/sdk/src/persistence/filelock/acquireFileLock.ts
```diff
@@ -21,16 +21,17 @@ export async function acquireFileLock(
...
 		debug(lockOrigin + " acquired a lockfile Retry Nr.: " + tryCount)
+
 		return stats.mtimeMs
 	} catch (error: any) {
 		if (error.code !== "EEXIST") {
```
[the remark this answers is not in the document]
martin-lysk: I reintroduce a comment here - hope this is more clear

martin-lysk (commented):
I just had a look into the PR to clarify questions. 

With a look at the [query api interface](https://github.com/opral/monorepo/pull/2801/files#diff-56f4d066e7f40ac8405608a284f52791bf2eee57fecb327bd2f74a68892a01afR11) and with the discussion @jldec and me had via around I feel  like we are not 100% alinged about the steps forward and the subscribe functionality exposed by the SDK.

jldec:
A simple throttle to the save did the trick (for now). The `set()` api now returns immediately (doesn't wait for the file write). I was able to generate, load, and translate, 30,000 messages - no reactivity yet - so the load-test does not re-load the translated files. 

```ts
$ DEBUG=sdk:store p load-test 30000 1 1 1 1

  load-test load-test start - watchMode on, ctrl C to exit - using experimental persistence +0ms
  load-test generating 30000 messages +0ms
  load-test opening repo and loading project +78ms
Using existing cloned repo
  sdk:store loadAll /Users/jldec/opral/monorepo/inlang/source-code/sdk/load-test/project.inlang/messages.json +0ms
  load-test subscribing to project.errors +413ms
  load-test loaded 30000 v2 MessageBundles +1ms
  load-test translating messages with inlang cli +0ms
Using existing cloned repo
  sdk:store loadAll /Users/jldec/opral/monorepo/inlang/source-code/sdk/load-test/project.inlang/messages.json +0ms
✔ Machine translate complete.                                                                                                             7:28:02 PM
  sdk:store saveall /Users/jldec/opral/monorepo/inlang/source-code/sdk/load-test/project.inlang/messages.json +2s
  sdk:store saveall /Users/jldec/opral/monorepo/inlang/source-code/sdk/load-test/project.inlang/messages.json +2s
  sdk:store saveall /Users/jldec/opral/monorepo/inlang/source-code/sdk/load-test/project.inlang/messages.json +1s
  load-test load-test done - watching for events +7s
```

samuelstroschein:
> A simple throttle to the save did the trick (for now). The set() api now returns immediately (doesn't wait for the file write).

This will lead to bugs. Is this an interim implementation?

Apps don't know when set now executed and the process can exit. We have a bunch of "sleep" workarounds because APIs don't await IO.

inlang/source-code/sdk/src/persistence/storeApi.ts
```diff
@@ -0,0 +1,16 @@
...
+export type StoreApi = {
+	messageBundles: Query<V2.MessageBundle>
+}
+
+/**
+ * WIP template for async V2 crud interfaces
+ * E.g. `await project.messageBundles.get({ id: "..." })`
+ **/
+export interface Query<T> {
+	get: (args: { id: string }) => Promise<T | undefined>
```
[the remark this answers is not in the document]
samuelstroschein: Why on T and not a separate subscribe method?

query.get("id")
query.subscribe("id")

inlang/source-code/sdk/src/persistence/storeApi.ts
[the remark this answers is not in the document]
samuelstroschein: Agree. I would just copy the mango query builder and be done with discussions how our query Api should look like.

samuelstroschein:
> @samuelstroschein could you clarify the reason why you think we have to do this?
I understand in the short term it would be nice to use separate files to help git to avoid merge conflicts, but once we have sqlite support, lix will have to do this differently right?

Optimize for plain git storage now. Sqlite support is years away (even if supported in lix, it won't be supported in git nor github). 

> I will look at both options to see which will unblock us to ship this PR soonest. File-per-message has more complexity which may not be justified given our immediate goal of shipping the v2 persistence api for apps asap.

1 week more or less doesn't matter. I assume that we will need split per bundle files to minimize merge conflicts within the next 12 months [as sqlite is at least 1-2 years away from production]. If you do it now, you avoid a migration and potentially breaking change.

> I think we can ship this PR without implementing file watching - but thinking ahead:

File watching is likely a must. Sherlock has bugs because the UI is constantly out of sync when the user pulls from git (an external file change).

inlang/source-code/sdk/src/persistence/storeApi.ts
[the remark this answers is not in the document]
samuelstroschein: if the argument is that a mango query builder would take 1-2 weeks longer to implement, i would invest those 1-2 weeks:

1. having a basic query builder now will ship the SDK faster but slow down apps (aka no speed advantage)
2. breaking change in the future because of argument 1

inlang/source-code/sdk/src/loadProject.ts
jldec: ok, thanks for clarifying the flow.

inlang/source-code/sdk/src/persistence/storeApi.ts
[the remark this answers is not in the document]
jldec: Thanks, that's helpful guidance.

The query builder api needs to live in a layer above this single entity, low-level store api, so that it can include other entities like lint reports. I will rename this api to Store<T> (instead of Query<T>) to make this clearer.

jldec:
> @jldec did i clarify all questions on your end?
yes, thanks @samuelstroschein 

#### durability
You are right, I agree it would be better for the store to make this explicit in the api.
Awaiting an update should wait for the write to the file system. (will do in this PR)
(b.t.w. we did remove the sleeps in the code after fixing the locking and read-write coordination in PR - as long as apps today don't forcefully exit the updates will be saved and data won't be lost because of load-overwrites)   

#### file-per-MessageBundle for git
- OK, the sdk will store each message bundle in a separate plain text file (future PR) 
- use dir/file names that users can recognize (matching human id ?)
- use multi-line indented JSON format for mvp, but open to improvements for git/human readability on that side too

#### file watching
- agree 💯 that we need this
- I have some ideas for how to improve this for sherlock - we're depending on node's fs.watch which is problematic. 

#### query builder and subscribe (reactivity)
- Also fully onboard that the SDK needs to provide this
- I was planning to implement these in another layer above the store api

inlang/source-code/sdk/src/persistence/storeApi.ts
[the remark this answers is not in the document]
samuelstroschein: Ah, okay Store instead of Query makes sense.

> the implementation can be non-performance optimized. as long as the API is set, apps have an easy time querying now and we avoid a breaking change

Addding to this remark: The query API doesn't even need to be complete. Functionality like filters (`$gte`) can be added incrementally.

samuelstroschein:
Perfect. Your remarks makes sense, seems like we are algined. @martin-lysk can you confirm alignment?

> file-per-MessageBundle for git

> use dir/file names that users can recognize (matching human id ?)

Not required. Optimizng the git storage doesn't;t mean it needs to be human readable. That said, @martin-lysk's implementation was using human ids and splitting that intro directories. Crucial: No more than 1000 nodes. Otherwise, github is choking. Ask him for more info.

jldec:
Thanks for the feedback @martin-lysk, 
I'm looking forward to seeing your thoughts about a "git compatible file format".

> I guess [this](https://github.com/opral/monorepo/pull/2801/files#diff-5824350d932fc3b3c17fdd8e61387a5c169243e5c405e2105e1cd5523ba8df40R213) is where we are not aligned yet and where i don't understand the direction of the PR specially about the intent to keep it separate from the current implementation.

The description in https://github.com/opral/inlang-message-sdk/issues/70 outlines the roadmap beginning with the low-level store for MessageBundles. This PR is the first step, and has already yielded good learnings.

The reasons to keep this work separate instead of using the existing "query" code are the following

1. We don't require backward compatibiliity now, and we may decide that it is not required at all.
2. We do need a simpler, more reliable store layer which we can reason about and keep evolving.
3. The code in createMessageQuery.ts and createMessageLintReportsQuery.ts has useful parts, but it is also complex and hard to understand, with hidden solidjs side effects and compromises, because of solid's treatment of async effects. I would honestly much rather invest time in building new apis without solidjs. 

All that said, I do like the direction you took with the delegate pattern for triggering the evaluation of lint rules, and the api changes we came up with to make getAll() behave reasonably, and I don't believe any of your efforts have been wasted time.

cc: @samuelstroschein

samuelstroschein:
@martin-lysk i understand that @jldec (correct me if i am wrong) want's to unblock the message component (cc @NilsJacobsen) by shipping persistency in this PR without backwards compatibility faster, then think about backwards compatibility afterwards (if it is even needed)

```
1. have the AST types in place
2. minimal/non-backwards compatible persistency to unblock message component
3. (if required) backwards compatibility 
```

PR #2813, opened 2024-05-23
fix fink login

janfjohannes:
fix fink login

---
* #2820
* #2610
* #2787
* __->__ #2813

### DIFF
--- a/inlang/source-code/website/src/services/local-storage/src/LocalStorageProvider.tsx
+++ b/inlang/source-code/website/src/services/local-storage/src/LocalStorageProvider.tsx
@@ -56,8 +56,6 @@ export function LocalStorageProvider(props: { children: JSXElement }) {
 
 	/** custom setStore to trigger localStorage.setItem on change */
 	const setStore: typeof setOriginStore = (...args: any) => {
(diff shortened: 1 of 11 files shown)


inlang/source-code/website/src/services/local-storage/src/LocalStorageProvider.tsx
```diff
@@ -146,6 +147,8 @@ export function LocalStorageProvider(props: { children: JSXElement }) {
 		}
 	}
 
+	window.addEventListener("storage", onStorageSetByOtherWindow)
```
NiklasBuchfink, line 150: I'm not sure if it is this part of the code, but something in the LocalStorageProvider triggers now a "window is not defined" error on the website.

janfjohannes: @NiklasBuchfink  is there ssr?

janfjohannes: i need a screenshot where and how that happens

PR #2840, opened 2024-05-29
feat: move the website project into the website package

NilsJacobsen:
see: https://github.com/opral/inlang.com/issues/74


NilsJacobsen:
@NiklasBuchfink can you remove paraglide from the editor package? There is a dependency on the website now that we don't want to have. It also causes problems on build time.

I would say as long as we do not dogfood the editor we should get completely rid of it in the package. Don't want to mess up the editor. Can you take over?

NiklasBuchfink:
@NilsJacobsen Sadly, it is not done in a minute. I tried it before, but I can take over.

NilsJacobsen:
No worries there is no extreme urge.

NiklasBuchfink:
@NilsJacobsen Paraglide is removed from Fink ✅

inlang/source-code/editor/src/interface/components/Card.tsx
```diff
@@ -6,7 +6,7 @@ import { showToast } from "./Toast.jsx"
 import { rpc } from "@inlang/rpc"
 import { Button } from "#src/interface/components/Button.jsx"
 import Link from "#src/renderer/Link.jsx"
-import * as m from "#src/paraglide/messages.js"
```
NilsJacobsen, line 9: Can we get rid of those? (Just use the hardcoded ones)

inlang/source-code/editor/src/interface/components/NewsletterForm.tsx
```diff
@@ -1,7 +1,7 @@
 import { createSignal } from "solid-js"
 import { showToast } from "./Toast.jsx"
 import { rpc } from "@inlang/rpc"
-import * as m from "#src/paraglide/messages.js"
```
NilsJacobsen, line 4: Same here.

inlang/source-code/editor/src/interface/components/SearchBar.tsx
```diff
@@ -2,7 +2,7 @@ import IconSearch from "~icons/material-symbols/search-rounded"
 import { currentPageContext } from "#src/renderer/state.js"
 import { createSignal } from "solid-js"
 import { navigate } from "vike/client/router"
-import * as m from "#src/paraglide/messages.js"
```
NilsJacobsen, line 5: Same here.

inlang/source-code/editor/src/renderer/Link.tsx
```diff
@@ -1,4 +1,4 @@
-import { languageTag, sourceLanguageTag } from "#src/paraglide/runtime.js"
```
NilsJacobsen, line 1: We should get rid of the runtime import here too.

NilsJacobsen (commented):
I think there are still some paraglide imports in here:)

NilsJacobsen:
Thanks a lot @NiklasBuchfink let's merge this on Monday. I expect some missing translations on the website so we shouldn't push it out without time (weekend).

PR #2863, opened 2024-06-03
extend library support with new matchers | Sherlock 🕵️‍♂️ [skip preview]

felixhaeberle:
- fix SHERL-65, see issue for targeted libraries


felixhaeberle:
cc @NiklasBuchfink why is ninja error'ing here?

NiklasBuchfink:
I have it on the list. It looks like the fetched messageLintReports are undefined in `sdk-load/project.inlang` and it results in the error. I'll add a check to prevent this. 

image

PR #2878, opened 2024-06-04
MESDK-130 -  fix Ninja undefined lintreports getAll.settled() in stubQueryApi

jldec:


jldec:
@NiklasBuchfink Is there any way for you to test this fix?
The PR is failing the ninja action, i guess because ninja still has the old version, not the PR build

PR #2881, opened 2024-06-04
fix: make PushException in Fink consistent

NiklasBuchfink:
@janfjohannes Just to double-check, does this correct the issues we had with the error handling, or did I miss something?

Related: https://linear.app/opral/issue/FINK-48/make-pushexception-in-fink-consistent


janfjohannes (approved):
should do the trick! push will always error before the pull after it, so probably not required to handle that case, if pull errors without push error we have a weird situation and it will throw , that should be ok.

PR #2885, opened 2024-06-05
turning off inline previews also turns off hover for messages | Sherlock 🕵️‍♂️

felixhaeberle:
- fix SHERL-76
cc @KraXen72


KraXen72:
it took 1200 lines??!? lol
thanks for the quick fix!

felixhaeberle:
> it took 1200 lines??!? lol
No. Git calculates it wrong. The change was about 10 lines.

PR #2886, opened 2024-06-05
t function matcher extract message should offer no brackets version | Sherlock 🕵️‍♂️

felixhaeberle:
- fix SHERL-64
cc @KraXen72

### DIFF
--- a/.changeset/wet-donkeys-rescue.md
+++ b/.changeset/wet-donkeys-rescue.md
@@ -1,0 +1,5 @@
+---
+"@inlang/plugin-t-function-matcher": patch
+---
+
+fix t-function matcher extract message should offer no-brace version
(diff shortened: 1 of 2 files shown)


KraXen72:
thanks! might i suggest wrapping the last option in a set of double quotes? because after extraction, the entire thing, including quotes gets replaced, so with the last option, it would mean you have to re-add the quotes manually

felixhaeberle:
@KraXen72 Can you make an exmaple where this would be helpful? The last option is only to have the messageId ready to copy if you use another format.

KraXen72:
<Text text="whatever" />
1. extract it, type message_id
2. change text attribute to tx as per my custom components

current:
<Text tx=message_id /> (invalid js, have to add " around it manually

proposed:
<Text tx="message_id" /> (valid js, ready to go without more changes)

felixhaeberle:
Then just extract `whatever ` without quotes?

PR #2897, opened 2024-06-06
support fink repo permisision detections

janfjohannes:
support fink repo permisision detections

### DIFF
--- a/lix/packages/client/src/github/client.ts
+++ b/lix/packages/client/src/github/client.ts
@@ -37,6 +37,30 @@ export function makeGithubClient({ gitHubProxyUrl }: { gitHubProxyUrl?: string }
 				return { error: newError }
 			})
 
+	type getAvailableReposType = (
+		installation_id: number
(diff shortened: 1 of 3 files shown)


janfjohannes:
await repo.getMeta() will allow you to detect if:

- forking is allowed, if not hide the fork feature
- user has admin rights  permissions.admin, if not and app not installed, disable push and tell user to have admin install app
- if pull permission is false and allow forking is true only show the forking workflow

janfjohannes:
update:
@NiklasBuchfink  you now also have an isInstalled boolean in the getmeta object to see if the app is installed. this should give you all information you need!

NiklasBuchfink (approved):
Should we change "allow_forking" to "allowForking"?

Everything else looks good, I will implement it tomorrow 👍

janfjohannes:
@NiklasBuchfink  i changed it. also i dont think logout on permission change is needed anymore, i hope i just did  a mistake while testing but lets keep an eye for inconsistent behaviour

PR #2901, opened 2024-06-07
SDK V2 Prerelease - (Slot based storage)

martin-lysk:
## Iteration 1 (Prerelease 15.7.)

Meta TODO's:
- PR description
- Ticket writing

TODO's
- Collect requirements
  - Fink
  - Sherlock

### @nils.jacobsen 

MessageBundle Editor
- Pattern
- Make the Component stylable (check where we still have issues) 
- Pass in installedLintRules to render translatable
- Minimal Docs in Storybook
- Clean helper functions (Move to SDK, write tests)

Settings Component 
- Adjustments for new settings schema
- Adjust docs

### @loris.sigrist 
- Finalize lint rule interface
  - API Types
  - Typescript gymnastic (finalize run method)
  - ✅ Group results by bundle and hash groups (offload work to worker)
  - Tests
- IPC FS - needed by sherlock and Parrot
  - ✅ integrate into sample web app for testing
  - test in vs code
  - Tests
- Review SDK code
  - Support with tests

### @martin.lysk1 

- Finalize SlotStorage
  - work in Jan's feedback
  - tests
  - Report property in bundle
  - custom conflict handler
- Merge state property in bundle
- Integrate lix (currently iso git)
- Finalize sdk2 api and loadproject2 (Martin)
  - load tests
  - Tests
  
## Iteration 2 (Experimental release 30.7.)
- ? Filter Bar 
- Import
- Export
- Conflict handling
- Migration strategy
  - settings shims?
  @loris.sigrist collects requirements (utility functions for upgrades)
- Quill/Tiptap/FBEditor based BundleEditor
- Hosted Component Docs
- Lint Rule Directives

...

## Iteration 3

 - Version history


jldec:
### notes from initial slotfile storage exploration

Slotfiles are an on-disk text file format designed to play nicely with git. They persist inlang message data into discreet slots in files, thereby avoiding collisions and minimizing merge conflicts.

### High level design

- a SlotFile contains an array with a predefined number of slots
- each slot may contain one record, aka "doc", with an id, or the slot may be empty
- the slot index is determined using a hash of the doc id
- this results in a very low probability of new docs occupying the same slot
- when the slot for a new doc is occupied in the first file, it lands on the same slot index in another slotfile (I think)  
- each slot is serialized as a line in the slotfile, separated from other slots by blank lines.

### Questions

- what does "transient" mean, particularly "transient slot file"?
- why is file read (load) mixed with write (save) e.g. in saveChangesToDisk
- what does "YDEEEAS" mean?
- why do we need a forceReload flag
- why this duplication `...loadResults.updated, ...loadResults.updated` (plan for tombstones?)

### Nits

- would love to see more passing tests
- format code, and fix or disable ts checks and lints before committing.
- wrap jsdoc comment lines where they are really long (e.g. types/SlotFile.ts)
- avoid hard-to-differentiate function names like loadSlotFileFromFs and loadSlotFilesFromFs
- avoid super long functions like saveChangesToDisk (>200 lines)

martin-lysk:
> ### notes from initial slotfile storage exploration
> Slotfiles are an on-disk text file format designed to play nicely with git. They persist inlang message data into discreet slots in files, thereby avoiding collisions and minimizing merge conflicts.
> 

> ### High level design
> * a SlotFile contains an array with a predefined number of slots
👍
> * each slot may contain one record, aka "doc", with an id, or the slot may be empty
👍
> * the slot index is determined using a hash of the doc id
👍
> * this results in a very low probability of new docs occupying the same slot
👍
> * when the slot for a new doc is occupied in the first file, it lands on the same slot index in another slotfile (I think)

It checks for a slot in the transient file / files if one excists first (transient files are about to be created - and slots used in those can only be occupied by the current process so conflicts can not occur). If no transient file exists yet (the normal case for non batch operations) it will check the files existing files by the nearest files by name first for a free slot. If all slots in existing files are occupied it will create a transient file.
When it’s slot is occupied it uses the name of the file it would create if there 
> * each slot is serialized as a line in the slotfile, separated from other slots by blank lines.
👍
> 
> ### Questions
> * what does "transient" mean, particularly "transient slot file"?

To understand this the following properties are important:
* New Records that are not yet persist are hold in memory and kept as transient records until persisted (successful fs.write file)
* Changes made on records that occupy a slot in a slot file are represented in the corresponding slotfiles changedRecords property 

During a saving to disc routine, the transient records tries to find a slot in existing files (to reduce file creation to a minimum) if no free slot is found a transient file is created (for the lifetime of a save operation). A transientFile is also taken into account to find free slots for other transient records (compare slot search from comment above)

> * why is file read (load) mixed with write (save) e.g. in saveChangesToDisk
This is an extra roundtrip to collect changes on the disc that may happen right before the save.  This is not a guarantee to catch all possible changes from other processes but since we will (TODO in place) block incoming event processing during the save an extra load was placed. Definitely a need to clearify and Optimize with the introduction of the lock. 

> * what does "YDEEEAS" mean?
it should be „ASDasd“ but I guess my fingers gone crazy when I needed a log I can look for during debugging 🤣
> * why do we need a forceReload flag
🕵️‍♂️ don’t know without looking into the code Will check tomorrow 
> * why this duplication `...loadResults.updated, ...loadResults.updated` (plan for tombstones?)
🕵️‍♂️ Will check - as well
> 
> ### Nits
> * would love to see more passing tests
Sure - I started writing some todos in the .test.ts file please add everything you would like to see 
> * format code, and fix or disable ts checks and lints before committing.
Disagree - I commit WIP and build only what I need in my focus, everything that kicks me out of the flow - but now it’s time to clean up - on it!  
> * wrap jsdoc comment lines where they are really long (e.g. types/SlotFile.ts)
👍
> * avoid hard-to-differentiate function names like loadSlotFileFromFs and loadSlotFilesFromFs
👍
> * avoid super long functions like saveChangesToDisk (>200 lines)
🤓👍

janfjohannes:
- could you split up the pr into a pr only with the slot file handling and the rest?
- transient should be replaced with something understandable, why not "unsaved"?
- its not safe to rely on separate normalize, json.stringify and string hashing for hashing, you need to use a hashing algo that does all in one step, there are a couple of examples easy to google
- lets try to remove the use of connect / disconnect (the naming also obfuscate that this just starts and stops watching). this lib would make little sense without fs and for testing we can use memoryfs.
- could you move to using function args with object destructuring,  its currently very hard to read eg. createSlotStorage("storage-1", 16, 3) >> createSlotStorage("storage-1", { numberSlots: 16, whateverThatIs: 3 })
- what is "storage name" and why is this needed?  i expected a folder path for the slot store here
- can we avoid storing the id hash in the file? it is only needed to find the slot and has to be regenerated anyways

martin-lysk:
> * could you split up the pr into a pr only with the slot file handling and the rest?

👍 Will do that

> * transient should be replaced with something understandable, why not "unsaved"?

Transient was the most natural to me and i know now why :D being grown up with hibernate... https://en.wikipedia.org/wiki/Transient_(computer_programming)#cite_ref-2 

> * its not safe to rely on separate normalize, json.stringify and string hashing for hashing, its better to use a hashing algo that does all in one step, there are a couple of examples easy to google, the main point is that json.stringify is platform and usage dependent

can you link those examples plz? #service 

> * lets try to remove the use of connect / disconnect (the naming also obfuscate that this just starts and stops watching). this lib would make little sense without fs and for testing we can use memoryfs.

👍 agree - will check if i see any reasoning for it or if this is a left over

> * could you move to using function args with object destructuring,  its currently very hard to read eg. createSlotStorage("storage-1", 16, 3) >> createSlotStorage("storage-1", { numberSlots: 16, whateverThatIs: 3 })

👍 Will do that

> * what is "storage name" and why is this needed?  i expected a folder path for the slot store here

Just added this recently for better log output. if you have multiple instances running you need a way to distinguish them.

> * can we avoid storing the id hash in the file? it is only needed to find the slot and has to be regenerated anyways

The Id and the respective hash stays the same - since id is immutable. Storing it is just a performance optimization to deal with thousands of objects. On 10k  records calculating the becomes a bottleneck alreadyd.

martin-lysk:
@janfjohannes what would you need in the separate pr, only the slotstorage folder and the cli I guess

janfjohannes:
> @janfjohannes what would you need in the separate pr, only the slotstorage folder and the cli I guess

the cli did not work and im not a fan of it anyways, i would only need the slotstorage folder + tests

janfjohannes:
> Transient was the most natural to me and i know now why :D being grown up with hibernate... https://en.wikipedia.org/wiki/Transient_(computer_programming)#cite_ref-2
maybe this does not need a prefix at all, obviously any in memory object is transient like anything that is in memory, the fact that writes are batched does not need to be reflected in every place that the object is used?


> can you link those examples plz? #service
most look like this: https://github.com/puleos/object-hash/blob/master/index.js    just that we dont need all the js stuff  that is not part of json



> Just added this recently for better log output. if you have multiple instances running you need a way to distinguish them.
multiple instances on different slot stores can use the folder name for distinguishing, multiple instances on the same slot store can use something more obviously made for debugging that does not impact the rest of the code

> The Id and the respective hash stays the same - since id is immutable. Storing it is just a performance optimization to deal with thousands of objects. On 10k records calculating the becomes a bottleneck already.
>
> premature optimization. its hard to se where this would be relevant, the file will have unneeded data, the code will be harder to reason about. sha-1 is also good enough. inlang has never a requirement to generate a massive amount of new entries at once except for machine translation where the api for that will be slower.

martin-lysk:
> > Transient was the most natural to me and i know now why :D being grown up with hibernate... https://en.wikipedia.org/wiki/Transient_(computer_programming)#cite_ref-2
> 
> maybe this does not need a prefix at all, obviously any in memory object is transient like anything that is in memory, the fact that writes are batched does not need to be reflected in every place that the object is used?

- will consider this


> > can you link those examples plz? #service
> 
> most look like this: https://github.com/puleos/object-hash/blob/master/index.js just that we dont need all the js stuff that is not part of json
> 

- thanks

> > Just added this recently for better log output. if you have multiple instances running you need a way to distinguish them.
> 
> multiple instances on different slot stores can use the folder name for distinguishing, multiple instances on the same slot store can use something more obviously made for debugging that does not impact the rest of the code
> 

- great idea - this makes sense combined with attaching it to a folder on creation (the folder was only available on connect before - which we change anyway.

> > The Id and the respective hash stays the same - since id is immutable. Storing it is just a performance optimization to deal with thousands of objects. On 10k records calculating the becomes a bottleneck already.
> 
> premature optimization. its hard to se where this would be relevant, the file will have unneeded data, the code will be harder to reason about. sha-1 is also good enough. inlang has never a requirement to generate a massive amount of new entries at once except for machine translation where the api for that will be slower.

Seems like you are right - and this might even be a relict from hash based slotentry distribution - where i needed the hash of the entry beyond the initial save -> will doublecheck. thanks for the push to simplification here!

janfjohannes:
@martin-lysk  one last thing before next round of reviews it would be good if you simplify the types , for me its not readable to see things like this:

```js
createSlotStorage<DocumentExample>(...)
````

its also not clear why the type would not be inferred and this is neceassary at all

**Comment by felixhaeberle** [comment]
@martin-lysk can we rename and update the PR description to reflect what all those changes in this PR are about?

Really helpful for the community tracking our progress on new functionality and see how they can benefit & also a great documentation for us later in the process.

File: inlang/source-code/sdk/src/persistence/slotfiles/createSlotWriter.test.ts [hunk 1]
```diff
@@ -0,0 +1,145 @@
...
+// TODO test that data witten by another process reaches the process via fs events
+
+describe("Disconnected slot storage", () => {
+	it("insert should lead to a transient record", async () => {
+		const memoryFs = createNodeishMemoryFs()
+
+		const slotStorage = await createSlotWriter<DocumentExample>({
+			fs: memoryFs,
+			path: "/slot/",
+			fileNameCharacters: 3,
```
janfjohannes, line 22: use a sane default, i expect this will rarely ever need to be touched.

inlang/source-code/sdk/src/persistence/slotfiles/createSlotWriter.test.ts
```diff
@@ -0,0 +1,145 @@
...
+			slotsPerFile: 16,
+			watch: false,
+		})
+		const insertedDocument: DocumentExample = {
+			id: "1",
+			content: "first document",
+		}
+		await slotStorage.insert(insertedDocument, false)
```
janfjohannes, line 30: use object instead of boolean

inlang/source-code/sdk/src/persistence/slotfiles/createSlotWriter.test.ts
```diff
@@ -0,0 +1,145 @@
...
+
+		expect(slotStorage._writerInternals.transientSlotEntries.size).eq(1)
+		const docs1 = await slotStorage.findDocumentsById([insertedDocument.id])
```
janfjohannes, line 33: what else would you find here? getAll() or findById should be more readable and precise enough

inlang/source-code/sdk/src/persistence/slotfiles/createSlotWriter.test.ts
```diff
@@ -0,0 +1,145 @@
...
+		expect(docs1.length).eq(1)
+		// expect(Object.isFrozen(docs1[0])).eq(true, "Records returned by slotStorage should be frozen")
+		// eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- we check the length before
+		expect(docs1[0]!.data.content).eq(insertedDocument.content)
+		const updateDocument = structuredClone(insertedDocument)
```
janfjohannes, line 38: why is clone needed? if storage can break without cloning, this has to move into the library

inlang/source-code/sdk/src/persistence/slotfiles/createSlotWriter.test.ts
```diff
@@ -0,0 +1,145 @@
...
+			id: "1",
+			content: "first document",
+		}
+
+		await slotStorage.insert(insertedDocument, false)
+
+		expect(slotStorage._writerInternals.transientSlotEntries.size).eq(1)
+
+		await slotStorage.save()
+		expect(slotStorage._writerInternals.transientSlotEntries.size).eq(0)
```
janfjohannes, line 88: aahhhhh so this is really a transient storage, not just a bad name for a memory layer on top of the file?! this is the first time i understand it is really working like that., maybe i misread this before?  if you just updated the memory representation of the file directly an then reconcile when saving you could reuse the same conflict logic as for writes by multiple threads and the whole impl would be simpler. i dont fully understand the need for this 3 layer design.

inlang/source-code/sdk/src/persistence/slotfiles/createSlotWriter.test.ts
```diff
@@ -0,0 +1,145 @@
...
+			id: "1",
+			content: "first document",
+		}
+		await slotStorage1.insert(insertedDocument, false)
+		expect(slotStorage1._writerInternals.transientSlotEntries.size).eq(1)
+		await slotStorage1.save()
+		expect(slotStorage1._writerInternals.transientSlotEntries.size).eq(0)
+
+		// force reload to avoid waiting for fs event
+		await slotStorage2.loadSlotFilesFromWorkingCopy(true)
```
janfjohannes, line 123: use object

inlang/source-code/sdk/src/persistence/slotfiles/createSlotWriter.test.ts
```diff
@@ -0,0 +1,145 @@
...
+		updateInSlot1.content = "updated in slot1"
+		await slotStorage1.update(updateInSlot1, false)
+
+		const updateInSlot2 = structuredClone(insertedDocument)
+		updateInSlot2.content = "updated in slot2"
+		await slotStorage2.update(updateInSlot2, false)
+
+		await slotStorage1.save()
+		await slotStorage2.save()
+		const conflictingRecord = slotStorage2.findDocumentsById([insertedDocument.id])[0]
```
janfjohannes, line 141: isnt locking supposed to prevent this case? we need locking in any case to now destroy the file write on concurrent threads, so it should be simple to just prevent any writes from other threads concurrently.

janfjohannes (changes requested):

inlang/source-code/sdk/src/persistence/slotfiles/createSlotWriter.test.ts
martin-lysk: done

samuelstroschein:
Closing because outdated. Inlang SDK v2 is rebuild on sqlite

PR #2902, opened 2024-06-07
Niklasbuchfink/fink-53-parsing-workflow-yaml-for-detecting-ninja-should-not-throw

NiklasBuchfink:
- extend GitHubActionEvent type
- remove console.error and early return after failed type check
- update tests to include newest workflow structure

### DIFF
--- a/inlang/source-code/cross-sell/cross-sell-ninja/src/index.ts
+++ b/inlang/source-code/cross-sell/cross-sell-ninja/src/index.ts
@@ -51,8 +51,6 @@ export async function isAdopted(args: { fs: NodeishFilesystem }): Promise<boolea
 					}
 				} else {
 					// Ignore invalid YAML files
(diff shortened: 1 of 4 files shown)


inlang/source-code/cross-sell/cross-sell-ninja/src/index.ts
```diff
@@ -51,8 +51,6 @@ export async function isAdopted(args: { fs: NodeishFilesystem }): Promise<boolea
 					}
 				} else {
 					// Ignore invalid YAML files
-					console.error("Failed to parse YAML", [...Value.Errors(GitHubActionsWorkflow, workflow)])
-					return false // Malformed YAML should result in a negative check
```
felixhaeberle, line 55: remove else?

NiklasBuchfink: Done ✅

PR #2905, opened 2024-06-07
Improve workflow conditions based on repo meta infos

NiklasBuchfink:


NiklasBuchfink:
Wait for:
- [ ] repo.getmeta.permissions.admin should return if the user is a owner of the repo/orga
- [ ] add permissions should directly open the install page for the current repo via the target id

NiklasBuchfink:
@janfjohannes This should help all users with setting up the repo and managing the permissions.

Only the owner status is missing to prevent users from data loss. I did it now using the admin state, and the logic is working, but the state is not representative of the user's ability to install apps or update permissions, as we discussed. Let's get this merged as soon as possible, as this will prevent plenty of user requests.

The direct link to update the permissions, including the repos target_id, can be changed afterward, as it is part of Lix.

samuelstroschein:
close because of fink2 which uses the new lix on sqlite?

PR #2913, opened 2024-06-10
v2 subscribe api 

jldec:
Parts 3 of https://github.com/opral/inlang-message-sdk/issues/70 - `subscribe` api which does not depend on solidjs.

- [x] RxJS Observable investigation
- [x] hand-written Observable ([experiment](https://github.com/jldec/watchfs))
- [ ] introduce RxJS subscribe reactivity for Store API 
- [ ] changeset


samuelstroschein:
@jldec awesome write up, thank you. I deem that you, @martin-lysk and I are aligned that RxJS observables is the basis for `*.subscribe()`? reasoning: 

- only major (?) production reactive system that is not framework-bound
- adapters to framework reactivity is trivial
- standardization processes around observables and signals are still years away (sad)

cc @janfjohannes who has experience with rxjs

janfjohannes:
@jldec  the important missing info is why hand written observable is not playing nice with the operators. what exactly are the issues?  
as i said i recoend adopting minimal surface area, if we get away with minimal internal subscribable totally fine. especially in the early days when rxjs 4 was quite slow companies like netflix used a minimal internal obeservable implementation with support for external operators and interop, so this was a common practice leading mostly to issues as the spec moved and there were incompatible apis leading to them using apis prefixed with classicObservable.   as recent rxjs has much better performance and does not bundle all operators immediately the need for this pattern was reduced, but can still make sense if it works and is compatible enough to be able to replace without api change.

jldec:
@janfjohannes - you were right, with a little bit of extra work ([subscribed state management](https://github.com/jldec/watchfs/blob/main/watchFs.ts#L25)) and the introduction of a `@@observable` symbol, and the use of [rxjs.from()](https://rxjs.dev/api/index/function/from), the hand-written observable does play nicely with the RxJS operators like [pipe](https://rxjs.dev/api/index/function/pipe) and [share](https://rxjs.dev/api/index/function/share).

I can maybe simplify things even more using `from(asyncIterator)`, but here is the [native RxJS watchFs](https://github.com/jldec/watchfs/blob/main/watchFs-rxjs.ts) observable and here is the [hand-written watchFs](https://github.com/jldec/watchfs/blob/main/watchFs.ts) observable. 

Example hand-written watcher usage from [the test here](https://github.com/jldec/watchfs/blob/main/watchFs.test.ts#L135-L138)
```ts
const watcher = watchFs({ nodeishFs, baseDir }) as unknown as rxjs.ObservableInput<string>

const unshared = rxjs.from(watcher)
const shared = rxjs.from(watcher).pipe(rxjs.share())
```

the equivalent with a native RxJS observable watcher looks like [this](https://github.com/jldec/watchfs/blob/main/watchFs.test.ts#L135-L138)
```ts
const watcher = watchFs({ nodeishFs, baseDir })

const unshared = watcher
const shared = watcher.pipe(share())
```

I'm going to move forward with the introduction of the RxJS dependency in this PR while we continue our evaluation of RxDB.

samuelstroschein:
closing because outdated

PR #2936, opened 2024-06-17
ship default human ids in sherlock already | Sherlock 🕵️‍♂️

felixhaeberle:
- fix SHERL-79

cc @samuelstroschein

### DIFF
--- a/inlang/source-code/ide-extension/README.md
+++ b/inlang/source-code/ide-extension/README.md
@@ -233,7 +233,7 @@ You can disable inline annotations by setting the following property to `false` 
 You can disable the automatic generation of human IDs by setting the following property to `false` in your VS Code `settings.json` file. The default value is `true`.
 
 ```json
-"sherlock.extract.autoHumanId": false
+"sherlock.extract.autoHumanId.enabled": false
 ```
 
 #### Troubleshooting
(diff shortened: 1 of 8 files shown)


inlang/source-code/ide-extension/README.md
```diff
@@ -228,6 +228,14 @@ You can disable inline annotations by setting the following property to `false`
...
 ```
 
+### Disable Auto Human ID Generation
+
+You can disable the automatic generation of human IDs by setting the following property to `false` in your VS Code `settings.json` file. The default value is `true`.
+
+```json
+"sherlock.extract.autoHumanId.enabled": false
+```
+
```
samuelstroschein, line 238: I would remove this. Wait for user complains. If non appear you have to maintain a config prop that nobody uses

felixhaeberle: I think this is a great middle-ground that will lead to a smooth transition. Since the effort to maintain other config properties is currently 0, the argument is not valid.

It's under my responsibility to let Sherlock thrive and I'm pretty confident this is the right thing to do.

samuelstroschein: > It's under my responsibility to let Sherlock thrive and I'm pretty confident this is the right thing to do.

Haha baba comment. OK, agree! :D 

> Since the effort to maintain other config properties is currently 0, the argument is not valid.

You just extended the docs of sherlock by another paragraph, making sherlock harder to understand ;)

PR #2940, opened 2024-06-17
SDK - Module Caching

LorisSigrist:
This PR adds a network-first module cache to the SDK. This enables devs to keep working offline & in case JSDelivr goes down.

In case the network isn't available, or incredibly slow, it can be annoying to have to wait for the request timeout. To speed this up the SDK now fetches modules in parallel instead of sequentially.


LorisSigrist:
The remaining failing tests are for persistence & seem unrelated. Locally they pass

LorisSigrist:
Questions for Reviewers:
- Is "Network first" the right strategy?
- How should we purge the cache of removed modules (it currently doesn't). IMO we can do this later since modules are rarely removed & barely use any disk space
- How can we test offline usage?

jldec:
Thanks for working on this @LorisSigrist! The implementation appears to work as intended. 👍 

I have an offline compile working in a test paraglide-sveltekit project, by running the build once, and then disconnecting :tada:  

I haven't looked at the code in detail yet, but here are some initial comments after testing:

1. how about generating the .gitignore also during `paraglide-js init` and in `createNewProject` to avoid the subsequent additional commit.  
1. the cache filenames are rather long (180+ characters) - any chance of reducing that.
1. I see typescript errors in the load-test project after the first run because it doesn't exclude project.inlang - any ideas for how to avoid this in other projects.
 (you can repro by running the following command twice: `pnpm run --filter @inlang/sdk-load-test test`)
1. I agree that we can defer the cache removal.
1. I didn't observe much of delay from the "network first" approech when running offline, so I think that's ok, to keep the modules fresh if users are using @latest. I do see a long pause at the end, after compiling, but I think that maybe coming from our telemetry or something else.

samuelstroschein (approved):
Network-first avoids cache invalidation problems. Hence, good choice to ship this fast. Can be improved later. 

Minor comments. Implement them or not. Your choice.

inlang/source-code/sdk/src/resolve-modules/cache.ts
```diff
@@ -0,0 +1,55 @@
...
+	return escaped
+}
+
+async function readModuleFromCache(
+	moduleURI: string,
+	projectPath: string,
+	readFile: NodeishFilesystemSubset["readFile"]
+): Promise<Result<string, Error>> {
+	const moduleHash = escape(moduleURI)
+	const filePath = projectPath + `/cache/${moduleHash}.js`
```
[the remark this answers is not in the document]
LorisSigrist: We can't use the module ID in the file-name since we need to be able to know the file-name before loading the module. The module ID cannot always be determined from the URL.

Currently the "hash" is just a hashed version of the URL to make sure it's FS compatible. It doesn't have anything to do with the content.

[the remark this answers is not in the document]
samuelstroschein: > We can't use the module ID in the file-name since we need to be able to know the file-name before loading the module. The module ID cannot always be determined from the URL.

makes sense.

Change the path from `/cache/*` to `/cache/modules/*` to enable `/cache/lint-reports/*` and other caches in the future?

samuelstroschein (approved):

inlang/source-code/sdk/src/migrations/maybeAddModuleCache.ts
```diff
@@ -0,0 +1,55 @@
...
+}): Promise<void> {
+	if (args.repo === undefined) return
+
+	const projectExists = await directoryExists(args.projectPath, args.repo.nodeishFs)
+	if (!projectExists) return
+
+	// This is how paths are handled in the other migrations.
+	// Is this reliable on windows?
+	const gitignorePath = args.projectPath + "/.gitignore"
+	const cachePath = args.projectPath + "/cache"
```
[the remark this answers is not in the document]
LorisSigrist: I tested it on the Shadow PC and it seems to work. Lix FS works with posix paths on windows

inlang/source-code/sdk/src/resolve-modules/cache.ts
```diff
@@ -2,10 +2,15 @@ import type { NodeishFilesystemSubset } from "@inlang/plugin"
...
 function escape(url: string) {
-	// collect the bytes of the UTF-8 representation & hex encode
+	// collect the bytes of the UTF-8 representation
 	const bytes = new TextEncoder().encode(url)
-	const escaped = [...bytes].map((byte) => byte.toString(16).padStart(2, "0")).join("")
-	return escaped
+
+	// 32-bit FNV1a hash to make the file-names shorter
+	// https://en.wikipedia.org/wiki/FNV-1a
+	const hash = bytes.reduce((hash, byte) => (hash * 0x01000193) ^ byte, 0x811c9dc5)
```
jldec, line 10: I'm not a mathematician - If you validate this say against https://github.com/sindresorhus/fnv1a to make sure it captures the same algorithm, then we can merge it.

inlang/source-code/cross-sell/cross-sell-sherlock/src/recommendation/utils/exists.test.ts
```diff
@@ -45,8 +45,10 @@ const mockFiles = (files: memfs.NestedDirectoryJSON) => {
 			// @ts-ignore - memfs has the same interface as node:fs/promises
 			vi.spyOn(_memfs.promises, prop).mockImplementation(lixFs[prop])
 		} else {
-			// @ts-ignore - memfs has the same interface as node:fs/promises
-			vi.spyOn(_memfs.promises, prop)
+			if (prop in _memfs.promises) {
```
jldec, line 48: was this change intentional?

inlang/source-code/paraglide/paraglide-js/src/cli/commands/init/command.test.ts
```diff
@@ -461,7 +468,9 @@ describe("maybeAddSherlock", () => {
 		])
 		await maybeAddSherlock({ project, logger, repo })
 		expect(consola.prompt).toHaveBeenCalledOnce()
-		expect(fs.writeFile).not.toHaveBeenCalled()
+		await expect(
+			async () => await fs.readFile("/.vscode/extensions.json", { encoding: "utf-8" })
+		).rejects.toThrow()
```
jldec, line 473: was this change intentional?

jldec (approved):
LGTM - see comment regarding checking the FNV implementation

fyi - i'd be fine with slightly longer file names (10-40 chars) if that gave us better collision avoidance or makes the filenames nicer e.g. if you want to use 52 bit from https://github.com/tjwebb/fnv-plus

inlang/source-code/cross-sell/cross-sell-sherlock/src/recommendation/utils/exists.test.ts
LorisSigrist: Yes.

On newer versions of `node` there is a `"glob"` property on `fs/promises` which is not present on memory FS. This caused the test setup to fail. I modified it to only spy on the subset of `fs` that is supported by `memfs`

inlang/source-code/paraglide/paraglide-js/src/cli/commands/init/command.test.ts
LorisSigrist: Yes, the caching was throwing off the "writeFile" count so I switched to a more reliable check. It might be even better to call `Sherlock.isAdopted` to check if it has **not** been adopted, but this works for now.

inlang/source-code/sdk/src/resolve-modules/import.ts
```diff
@@ -53,11 +44,90 @@ async function $import(
...
+	if (!isValidUrl(uri))
+		throw new ModuleImportError({ module: uri, cause: new Error("Malformed URL") })
+
+	const result = await tryCatch(async () => await fetch(uri))
+	if (result.error) {
+		throw new ModuleImportError({
+			module: uri,
+			cause: result.error,
+		})
+	}
```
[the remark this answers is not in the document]
LorisSigrist: Let's leave this as a separate feature for now

inlang/source-code/sdk/src/resolve-modules/cache.ts
LorisSigrist: I upgraded to a 64 bit version for more hash space. 

The implementations actually weren't equivalent. Mine didn't discard the bits past 32 bits, so there was some extra junk in the binary representation. The upgraded 64bit one is equivalent.

PR #2944, opened 2024-06-18
Expose Error types that are part of the external api

LorisSigrist:
See: https://linear.app/opral/issue/MESDK-140/expose-error-types-that-are-part-of-the-external-api

Reviewers: If this change is ok for you you can merge this

### DIFF
--- a/.changeset/pink-windows-divide.md
+++ b/.changeset/pink-windows-divide.md
@@ -1,0 +1,5 @@
+---
+"@inlang/sdk": patch
+---
+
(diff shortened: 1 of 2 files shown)


jldec (approved):
LGTM - thanks @LorisSigrist 
Small nit, I would prefer to list the exported error classes one by one in the index file instead of using `*` for readability and to avoid inadvertent exports in future. ( I can do it for you if you're busy with the react summit 😜 )

jldec:
I will merge when CI is happy

PR #2953, opened 2024-06-19
MESDK - Better settings reactivity

LorisSigrist:
Closes https://github.com/opral/inlang-message-sdk/issues/77
Closes https://github.com/opral/inlang-message-sdk/issues/35

This PR reworks how Settings reactivity is handled & simplifies a lot of code. The `settings.json` file is now watched & the SDK will react to changes in it, as if `project.setSettings()` had been called. This matters when the file is changed by git

I hesitated to just straight up watch the `project.inlang` directory since that now also includes the cache.


inlang/source-code/sdk/src/createNodeishFsWithWatcher.ts
```diff
@@ -8,67 +8,75 @@ import type { NodeishFilesystem } from "@lix-js/fs"
...
+			throw err
+		}
 	}
 
-	const readFileAndExtractPath = (path: string, options: { encoding: "utf-8" | "binary" }) => {
-		if (!pathList.includes(path)) {
-			makeWatcher(path)
-			pathList.push(path)
+	/**
+	 * Creates watchers on-the-fly for any file or directory that is not yet watched.
```
lucidNTR, line 53: @LorisSigrist  please add a comment that the only reason we have to do this is that node 18 on linux does not support recursive file watching, otherwise we would add a recursive watch at the root and then just filter out the events with the list of used files.

lucidNTR (commented):
looks like a good step forward but of course can be muuuch better when we drop node 18 support and have a lix fs api. helped me look at this getting the requirements, so please always ping me when there are things touching watcher of fs wrappers.

felixhaeberle:
@jannesblobel can we please remove the double @janfjohannes / @lucidNTR ?

cc @janfjohannes I suppose this is your real account? Block @lucidNTR from accidentally commenting on opral repos?

janfjohannes:
@felixhaeberle  this is a public repo, there is nothing to block or remove. githubs user switcher is really bad this sometimes happens on pr comments and i think it does not really impoact you

felixhaeberle:
There is no value in you using two accounts for your work here, so I blocked the user you are accidentally using from time to time.

Bildschirmfoto 2024-06-21 um 14 05 08

Feel free to change this if there is a real reason why you should work on two accounts.

PR #2987, opened 2024-07-01
create paraglidejs plugin to ease communication

felixhaeberle:
- fix SHERL-89


felixhaeberle:
cc @LorisSigrist just a small insight from the "bundle multiple plugins in one" try-out:

open questions:
- how does plugin id + plugin specific settings like `settings["plugin.inlang.messageFormat"] -> settings["plugin.inlang.paraglideJs"]` get inherited?

---

more thoughts, but mostly inapplicable:

- is it possible to use an old configuration when switching like use `settings["plugin.inlang.messageFormat"]` only?
- is it possible to use two plugin settings at the same time `settings["plugin.inlang.messageFormat"] + settings["plugin.inlang.paraglideJs"]`, even multiple properties from multiple plugin settings. (arg 1 from settings1, arg2 from settings2 = {arg1, arg2} from settingsCombined)

samuelstroschein:
close because stale?

PR #3015, opened 2024-07-09
exposes readBlob via lix client

martin-lysk:
This PR exposes a readBlob function to via lix client to allow to read a content of a blob via oid directly.


janfjohannes (approved):

samuelstroschein:
Close because outdated? Lix is being re-build on sqlite

PR #3017, opened 2024-07-11
fix: Ninja recommend and add from subdir

NiklasBuchfink:
- always check and add Ninja workflow in the root of the repo using `findRepoRoot` to create the relativePath (helper function)
- update paraglide init to not prompt the user if Ninja is not recommend
- add test case for `shouldRecommend` and `add`
- refactor tests to use `createNodeishMemoryFs`

### DIFF
--- a/.changeset/slow-panthers-greet.md
+++ b/.changeset/slow-panthers-greet.md
@@ -1,0 +1,6 @@
+---
+"@inlang/recommend-ninja": patch
+"@inlang/paraglide-js": patch
+---
+
+Fix Ninja recommendation and adoption if working directory is not the repo root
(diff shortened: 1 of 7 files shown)


NiklasBuchfink:
@felixhaeberle recommend-sherlock has the same issue.

If paraglide init is executed in `inland/source-code/editor,` the `.vs-code` directory is added to the same directory instead of the root of the repo.

You can check how I solved it for Ninja and reuse the helper function I used to calculate the relative path. I also recommend refactoring the tests.

NiklasBuchfink:
@LorisSigrist There is a timeout in the paraglide e2e test. Probably because of this:
https://github.com/opral/monorepo/blob/2f19bf0d2ece8c9596bdc9c72bdcae91f09341f3/inlang/source-code/end-to-end-tests/paraglide-js/src/e2e.test.ts#L56

The prompt does not appear anymore, if the action is detected and it probably detects the one of our monorepo as there sits the repo root

LorisSigrist (approved):

PR #3024, opened 2024-07-22
SDK v2

martin-lysk:


inlang/source-code/sdk-v2/src/types/plugin.ts
```diff
@@ -117,17 +137,9 @@ export type ResolvePlugins2Function = (args: {
...
-	// }) => Promise<Message[]> | Message[]
-	// TODO SDK-V2 check how to setup saveMessages
-	// saveMessages: (args: {
-	// 	settings: ProjectSettings2
-	// 	messages: Message[]
-	// 	nodeishFs: NodeishFilesystemSubset
-	// }) => Promise<void> | void
+	toBeImportedFiles: Array<Plugin2["toBeImportedFiles"]>
+	importFiles: Array<Plugin2["importFiles"]>
+	exportFiles: Array<Plugin2["exportFiles"]>
```
felixhaeberle, line 142: @martin-lysk @samuelstroschein how should the resolved plugin API look like? do we support multiple exporter / importer?

inlang/source-code/sdk-v2/src/types/schema.ts
```diff
@@ -1,113 +1,127 @@
...
+	id: Type.String(),
+	alias: Type.Any(), // TODO: Use appropriate Typebox type for JSONColumnType<AliasMap>
+})
+
+export const MessageTable = Type.Object({
+	id: Type.String(),
+	bundleId: Type.String(),
+	locale: Type.String(),
+	declarations: Type.Any(), // TODO: Use appropriate Typebox type for JSONColumnType<Declaration[]>
+	selectors: Type.Any(), // TODO: Use appropriate Typebox type for JSONColumnType<Expression[]>
```
felixhaeberle, line 61: @samuelstroschein / @martin-lysk what are we going to do with the kysely types in typebox?

martin-lysk: @felixhaeberle for now we will just expose the sql db and do no validation on data we pass into the db - this should change and everything you update/insert should go through a runtime validation check.

inlang/source-code/sdk-v2/src/types/plugin.ts
samuelstroschein: We don't need resolved plugin api anymore as we don't merge APIs together.

samuelstroschein:
Closing because outdated. SDK v2 will use sqlite instead of RxDB

PR #3025, opened 2024-07-24
feat(plugins/i18next): Use path to decide default namespace

MrTwixxy:
_Disclaimer: I haven't tested it yet, since Im unable to build myself_

This fixes the default behavior of using the first namespace when there's none defined.
Useful when you have a monorepo and only want 1 sherlock project

### DIFF
--- a/inlang/source-code/ide-extension/src/decorations/messagePreview.ts
+++ b/inlang/source-code/ide-extension/src/decorations/messagePreview.ts
@@ -66,6 +66,7 @@ export async function messagePreview(args: { context: vscode.ExtensionContext })
 		const wrappedDecorations = (messageReferenceMatchers ?? []).map(async (matcher) => {
 			const messages = await matcher({
 				documentText: activeTextEditor.document.getText(),
+				documentPath: activeTextEditor.document.uri.path,
 			})
 
 			return messages.map(async (message) => {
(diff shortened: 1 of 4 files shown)


felixhaeberle:
@MrTwixxy can you sign the CLA?

MrTwixxy:

felixhaeberle:
@MrTwixxy I've approved the pipeline so we can check if the changes leads to errors.

MrTwixxy:
I've fixed the tests and added a new one. 
The behavior isn't 'defined' somewhere, but the default behavior isn't either. 
I don't expect things to break for the following reasons:
 - Nothing changed if namespaces are used how they normally are
 - If you aren't using them how they normally are, it's just using always using the default, which still happens if the path doesnt match any namespace

felixhaeberle:
hey, we have to discuss this change internally as this would mean that every plugin now needs some kind of strategy on how to do `documentPath`'s, see https://github.com/MrTwixxy/inlang-monorepo-fork/blob/7ec9773714d9875bab0b3778a7567b4b103f1aed/inlang/source-code/versioned-interfaces/plugin/src/customApis/app.inlang.ideExtension.ts#L43

MrTwixxy:
Having the argument passed into the function doesn't necessarily mean that it has to be used right?

samuelstroschein:
@felixhaeberle i tend to agree with @MrTwixxy. just because the argument exists, doesn't mean that a plugin needs to use `documentPath`.

samuelstroschein:
i propose to close this pr given that we will re-implement the custom API https://github.com/opral/inlang-sdk/issues/125

PR #3028, opened 2024-07-24
copy suggestions

samuelstroschein:
See the commits and messages for changes. 

(We would love to see an image diff here, wouldn't we? :D )


NilsJacobsen (approved):
Looks good. What do you mean with image diff?

PR #3041, opened 2024-08-05
add: providePlugin api

samuelstroschein:
## Changes

Lix can be opened with `providePlugin: [myplugin]`. 

```ts
const mockPlugin: LixPlugin = {
  key: "mock-plugin",
  glob: "*",
  diff: {},
}
const lix = await openLixInMemory({ blob: await newLixFile(), providePlugin: [mockPlugin] })

```

## Why

It avoids separating the inlang plugin code from the app (SDK) code. 

- no dual packaging
- no bundling needed

## Additional information 

I have concerns about the portability of the change. Any inlang file must now be loaded through the inlang SDK for lix to work. The plugin is not saved in lix. We can revisit this decision but for now we can iterate faster.

### DIFF
--- a/lix/packages/sdk/src/open/openLix.ts
+++ b/lix/packages/sdk/src/open/openLix.ts
@@ -22,7 +22,7 @@ export async function openLix(args: {
 	 * @example
 	 *   const lix = await openLixInMemory({ blob: await newLixFile(), providePlugin: [myPlugin] })
 	 */
-	providePlugin?: LixPlugin[]
+	providePlugins?: LixPlugin[]
 }) {
 	const db = new Kysely<LixDatabase>({
 		dialect: createDialect({ database: args.database }),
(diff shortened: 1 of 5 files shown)


lix/packages/sdk/src/open/openLix.ts
```diff
@@ -8,13 +8,31 @@ import { contentFromDatabase, createDialect, type SqliteDatabase } from "sqlite-
...
+	 *
+	 * (+) avoids separating app code from plugin code and
+	 *     resulting bundling logic.
+	 * (-) such a file format must always be opened with the
+	 *     file format sdk. the file is not portable
+	 *
+	 * @example
+	 *   const lix = await openLixInMemory({ blob: await newLixFile(), providePlugin: [myPlugin] })
+	 */
+	providePlugin?: LixPlugin[]
```
janfjohannes, line 25: why is it not called plugins?  pvovide seems confusing.

janfjohannes (approved):

lix/packages/sdk/src/open/openLix.ts
samuelstroschein: will change it!

PR #3043, opened 2024-08-06
refactor: add interpretation for deletion

samuelstroschein:
### Changes

Lix was not able to account for a deletion. Support is added by

1. allowing a `diffReport.neu` value to be undefined (the value doesn't exist anymore)

2. change the diff report api to provide the old and/or new value, dependent on the operation, which simplifies working with diff reports but is also required. take a `deletion` as an example. if the old value is not reported, the id information of the type is lost.

Side benefits: 

- dealing with diffs is simplified because querying old changes is oftentimes not required (the diff provides the previous value)
- the API aligns more with SQL old, neu triggers which should simplify the mental model.

https://www.loom.com/share/cec409f4f3194931aa1af4cb479aabcc?sid=34de3198-2898-421e-9558-364ddcf2396d

### Additional information 

- the functionality of lix has not been tested as the inlang plugin was not ready and the csv prototype not available either
- Types behave as expected

CleanShot 2024-08-05 at 22 53 23@2x


samuelstroschein:
As a follow-up step, I would add the `operation` property to a change too based on the diff report.

samuelstroschein:
Yep, returning `old` and `neu` is the right move. @NilsJacobsen got into problems displaying diffs because he has to query the old changes

lix/packages/sdk/src/open/openLix.ts
```diff
@@ -142,10 +142,13 @@ async function handleFileChange(args: {
 			neu: args.neuData,
 		})
 		for (const diff of diffs ?? []) {
+			// assume an insert or update operation as the default
+			// if diff.neu is not present, it's a delete operation
+			const value = diff.neu ?? diff.old
```
janfjohannes, line 147: this is only safe to use for things that cannot change, i would do const id = diff.neu?.id ?? diff.old?.id
to prevent accidentally using the value in unexpected ways later on.

janfjohannes (approved):

lix/packages/sdk/src/open/openLix.ts
samuelstroschein: IMplemented. Thanks

samuelstroschein:
Erghhhhhhh we should start writing unit tests :D 

https://www.loom.com/share/96e7fbce5b5c465892f92aea4e2a3d77?sid=4ed04b09-f382-4a81-aae7-91815cc91e7c

PR #3044, opened 2024-08-06
implement loadProjectFromDirectory plugin.loadMessages logic

martin-lysk:
Context

We want to support v1 plugins in inlang dev tools (paraglide, sherlock).

Todo

Implement importing messages via the legacy plugin.loadMessages function.


inlang/source-code/sdk2/src/lix-plugin/inlangLixPluginV1.ts
```diff
@@ -22,6 +22,8 @@ export const inlangLixPluginV1: LixPlugin<{
 	diff: {
 		// TODO does not account for deletions
 		file: async ({ old, neu, path }) => {
+			const diffRunId = new Date().getTime() + "_" + Math.random();
+			console.time(diffRunId + " diff:");
```
samuelstroschein, line 26: Scheduling issues?

inlang/source-code/sdk2/src/project/util/insertBundleNested.ts
```diff
@@ -0,0 +1,47 @@
...
+		throw new Error("Project is undefined");
+	}
+	await project.db
+		.insertInto("bundle")
+		.values({
+			id: bundle.id,
+			alias: json(bundle.alias),
+		})
+		.returning("id")
+		.execute();
```
samuelstroschein, line 19: Suggested changes to make the function generalizable/available for inlang apps:

1. I would move the function to `query-utilities`
2. I would not let the function .execute() and define what should be returned (leave it to apps)
3. I would use `db` as the argument instead of `project` (you don't need project)
4. Why can `project`/`db` be undefined?

inlang/source-code/sdk2/src/project/loadProject.ts
```diff
@@ -55,7 +55,9 @@ export async function loadProject(args: {
 		db,
 		plugins: {
 			get: () =>
-				structuredClone(reactiveState.plugins$.getValue()) as InlangPlugin[],
+				// TODO loadMessages - we cant use structuredClone since we have a function
+				// structuredClone(reactiveState.plugins$.getValue()) as InlangPlugin[],
+				reactiveState.plugins$.getValue() as InlangPlugin[],
```
samuelstroschein, line 60: Good remark. I would use the Typescript `Readonly` type in the api definition to leave the plugins as mutable but consumers from mutating because Typescript would complain.

inlang/source-code/sdk2/src/project/loadProjectFromDirectory.ts
```diff
@@ -5,6 +5,9 @@ import { uuidv4, type Lix } from "@lix-js/sdk";
 import type fs from "node:fs/promises";
 // eslint-disable-next-line no-restricted-imports
 import nodePath from "node:path";
+import type { InlangPlugin } from "../plugin/schema.js";
+import { insertNestedBundle } from "./util/insertBundleNested.js";
+import { fromMessageV1 } from "./util/fromMessageV1.js";
 
 /**
  * Loads a project from a directory.
```
samuelstroschein, line 13: (Not very important) Rename `loadProjectFromDirectory` to `loadProjectFromDirectoryInMemory` to clarify that this function loads a project in memory? Later we can add `loadProjectFromDirectoryWithTempFile()` or similar functions. Yes, long name but it's only called once.

inlang/source-code/sdk2/src/project/loadProjectFromDirectory.ts
```diff
@@ -18,15 +21,56 @@ export async function loadProjectFromDirectory(
...
 	>
 ) {
-	const project = await loadProjectInMemory({
+	const tempProject = await loadProjectInMemory({
 		// pass common arguments to loadProjectInMemory
 		...args,
 		blob: await newProject(),
 	});
-	await copyFiles({ fs: args.fs, path: args.path, lix: project.lix });
+	await copyFiles({ fs: args.fs, path: args.path, lix: tempProject.lix });
```
samuelstroschein, line 29: Why is the `tempProject` needed?

inlang/source-code/sdk2/src/project/loadProjectFromDirectory.ts
```diff
@@ -18,15 +21,56 @@ export async function loadProjectFromDirectory(
...
+			});
+		}
+	}
+
 	return project;
 }
 
+async function loadLegacyMessages(args: {
+	project: Awaited<ReturnType<typeof loadProjectInMemory>>;
+	loadMessagesFn: Required<InlangPlugin>["loadMessages"];
```
samuelstroschein, line 56: Try this

```suggestion
	loadMessagesFn: Required<InlangPlugin>["loadMessages"]; 
  loadMessagesFn: NonNullable<InlangPlugin["loadMessages"]>;
```

samuelstroschein (commented):
Looks OK. 

I can extract the query utility after merging. Just need clarification why the temp project is needed.

inlang/source-code/sdk2/src/project/loadProjectFromDirectory.ts
martin-lysk: see todo https://github.com/opral/monorepo/blob/38da990bc2567e19d276065c88a4d30bd30f5e0e/inlang/source-code/sdk2/src/project/loadProjectFromDirectory.ts#L31

inlang/source-code/sdk2/src/lix-plugin/inlangLixPluginV1.ts
martin-lysk: nope - leftover from debugging of commit increment - will remove it again

inlang/source-code/sdk2/src/project/loadProjectFromDirectory.ts
martin-lysk: Required actually worked - missed to drop the todo

inlang/source-code/sdk2/src/project/loadProjectFromDirectory.ts
samuelstroschein: roger

inlang/source-code/sdk2/src/project/util/insertBundleNested.ts
martin-lysk: 1. greate
2. we have multiple executes in the function, one per bundle/message/variant... we could wrap them into a transaction or what would you return?
3. just copied from fink2 - which was copied from the original prototype :D - guess thats a leftover - fixed in next commit
4. see 3, done

samuelstroschein:
@martin-lysk merging this as discussed on the call to generalize the helper and avoid running out of sync

PR #3045, opened 2024-08-06
refactor: expose lix file

samuelstroschein:
## Context

While writing the inlang lix plugin, I ran into the problem that I needed more information than the Blob of the file. Specifically, the information of the `path` was crucial as the inlang lix plugin diffs various files (settings.json, db.sqlite, ...). 

I added a workaround by exposing `path` to the differ. 

```ts
diff: {
  file: async ({ old, neu, path })
}
```

## Proposal 

Expose `LixFile` as `old` `neu` arguments for the differ.

```diff
diff: {
-  file?: (args: { old?: LixFile['data']; neu?: LixFile['data'] }) 
+  file?: (args: { old?: LixFile; neu?: LixFile }) 
}

```

The workaround works but it felt like exposing the lixfile type would be the way to go. Eventually, a differ might need all data in a lix file. Instead of adding more and more arguments to `diff.file`, it seems easier to expose the lix file type. 

- (+) solves a requirement for the inlang lix plugin 
- (+) avoids adding more and more arguments to the differ that are contained in the lix file itself (like path)
- (-) leaks the `LixFile` to plugins. a breaking change would require `diff.fileV2` that seems to be good practice anyways

### DIFF
--- a/inlang/source-code/sdk2/src/lix-plugin/inlangLixPluginV1.test.ts
+++ b/inlang/source-code/sdk2/src/lix-plugin/inlangLixPluginV1.test.ts
@@ -18,10 +18,14 @@ describe("plugin.diff.file", () => {
 				alias: JSON.stringify({}),
 			})
 			.execute();
+		const path = "/db.sqlite";
 		const diffReports = await inlangLixPluginV1.diff.file!({
 			old: undefined,
-			neu: contentFromDatabase(neuProject._sqlite),
-			path: "/db.sqlite",
+			neu: {
+				id: "uuid",
+				path,
+				data: contentFromDatabase(neuProject._sqlite),
+			},
 		});
 		expect(diffReports).toEqual([
 			{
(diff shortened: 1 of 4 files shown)


janfjohannes:
@samuelstroschein  aaaaaahh, please dont call that lix file :D a lix file is the sqlite lix databse for me, i was super confused until i realized you mean the file type definition from lix.

janfjohannes (approved):
@samuelstroschein  i understand you first felt this way, but if we want thefile id to really be used as file identifier which is our plan, this is absolutely the right way to go!

samuelstroschein:
> @samuelstroschein aaaaaahh, please dont call that lix file :D a lix file is the sqlite lix databse for me, i was super confused until i realized you mean the file type definition from lix.

Oh, i see confusing indeed. Do you have a better alternative for "files in a lix"? Using just `File` won't work as `File` is a global type in TypeScript node envs that would overshadow our file type (led to several bugs in inlang when we used `Plugin` as a type instead of `InlangPlugin`)

janfjohannes:
@samuelstroschein  no i just mean when talking, its fine as a type definition. but you just said lix file without mention you mea the "file type from lix"

PR #3046, opened 2024-08-06
add: `change.operation`

samuelstroschein:
## Context

Apps can't know if a change was an insert, update, or a delete. This is a follow up of https://github.com/opral/monorepo/pull/3043. 

## Proposal 

Expose the `operation` information that differs provide in a change. Apps will be able to display if a change was/is an insert, update, or deletion.  

```diff

export type Change = {
	id: string
	file_id: LixFile["id"]
	commit_id?: Commit["id"]
	plugin_key: LixPlugin["key"]
+	operation: "insert" | "update" | "delete"
	type: string
	value?: 
	meta?: string // JSONB
}


```

## Additional information

Testing with finkv2 indicates that everything works as expected. However, the written unit test is failing because the query `await lix.selectFrom("changes")` doesn't resolve. Don't know why. Works in fink.

### DIFF
--- a/inlang/source-code/sdk2/src/lix-plugin/inlangLixPluginV1.test.ts
+++ b/inlang/source-code/sdk2/src/lix-plugin/inlangLixPluginV1.test.ts
@@ -30,7 +30,7 @@ describe("plugin.diff.file", () => {
 		expect(diffReports).toEqual([
 			{
 				type: "bundle",
-				operation: "insert",
+				operation: "create",
 				old: undefined,
 				neu: { id: "1", alias: {} },
 			} satisfies DiffReport,
@@ -121,7 +121,7 @@ describe("plugin.diff.file", () => {
 		expect(diffReports).toEqual([
 			{
 				type: "message",
-				operation: "insert",
+				operation: "create",
 				old: undefined,
 				neu: {
 					id: "1",
@@ -238,7 +238,7 @@ describe("plugin.diff.file", () => {
 		expect(diffReports).toEqual([
 			{
 				type: "variant",
-				operation: "insert",
+				operation: "create",
 				old: undefined,
 				neu: {
 					id: "1",
@@ -372,7 +372,7 @@ describe("plugin.diff.variant", () => {
 		};
 		const diff = await inlangLixPluginV1.diff.variant({ old, neu });
 		expect(diff).toEqual([
-			{ operation: "insert", type: "variant", neu, old } satisfies DiffReport,
+			{ operation: "create", type: "variant", neu, old } satisfies DiffReport,
 		]);
 	});
 });
--- a/inlang/source-code/sdk2/src/lix-plugin/inlangLixPluginV1.ts
+++ b/inlang/source-code/sdk2/src/lix-plugin/inlangLixPluginV1.ts
@@ -111,7 +111,7 @@ function jsonStringifyComparison({
 	type: "bundle" | "message" | "variant";
 }): DiffReport[] {
 	if (old === undefined && neu) {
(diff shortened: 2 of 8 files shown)


janfjohannes (changes requested):
i expect this is what the meta field is for, no need to add this as new field. especially as the workding insert update delete is sql lingo and only makes sense for inlang as that uses sqlite, but other file formats might have different operations.

samuelstroschein:
@janfjohannes Can you give examples of more operations that are needed to model change conflicts? 

I think you refer to collisions that should happen based on two update "operations" e.g. if the drill size changed, and someone else changed the drill size, that should raise a change conflict. Those are still two "update" operations, though. Modeling the drill size changed change conflict would happen as so: 

```ts
if (old.drillSize !== neu.drillSize){
  changeConflicts.push("Drill size differed. Choose one.", fixes: [])
}
```

There is no need to express "drill size changed" as part of the operation [given that we store snapshots]. 

**Reasons for the operation first class prop**

- the operations 'insert', 'update','delete' and contained types seems universal across any file given that we store snapshots. 
- lix (via the host) and lix apps need awareness if something has been inserted, updated, or deleted to render the correct diffs
- without the operation information, working with `change.value` is cumbersome because it's always potentially undefined. a check with `if (change.operation !== "deletion")` to derive that `change.value` exists is not possible.


Examples of universality:

- inlang: insertion, update, deletion of bundle, message, variant, settings
- source code: insertion, update, deletion of a line 
- semantic soure code: insertion, update, deletion of a function, variable, class, etc.
- markdown: insertion, update, deletion of a paragraph, word, ...
- cad: insertion, update, deletion of component
- figma: insertion, update, deletion of layer, frame, component, vector, ...

janfjohannes:
@samuelstroschein  insert is allready an example that makes no sense outside sql, you inssert into a table but you would otherwise create an id or a document etc. other formats might have move, rename, change size etc. 

can you give a reason why this cannot be put into meta?

samuelstroschein:
I'll build a prototype that contains the operation field in `meta` to see how good/bad that works. Will report later today.  

> can you give a reason why this cannot be put into meta?

Hmm. I assumed that lix itself, the latest with the host, needs to know if something is a create, update, or delete operation. Otherwise, diffing becomes complicated due to the implicit meaning of create, update, and delete, which might run out of sync with a plugin's understanding of create, update, delete. Lix current implicit definition: 

- `create` = if (change.parent_id === undefined && value)
- `update` =  if (change.parent_id && value)
- `delete` =  if (change.parent_id && value === undefined) 

Communicating this will lead to complicated docs, and surely to downstream bugs because the implicitness will lead to a situation where lix treats something as X while the plugin had a different intent:

```ts
change = {
  parent_id: "53239jsfsfs",
  value: undefined, // <- undefined is implicitly treated by lix as deletion (will lead to a bug)
  meta: {
    "operation": "create"
  }
}

```

Furthermore, working with changes becomes complicated because 

> you inssert into a table but you would otherwise create

It doesn't matter if we name it `create, update, delete` or `insert, update, delete`. I can rename it to `create` instead of `insert`.

> other formats might have move, rename, change size etc.

This seems to be all updates of the provided snapshot. The semantic meaning "move," "rename," or "change size" might be meta information that a plugin/app saves to avoid a diff, but the data was updated regardless of whether it was moved, renamed, or whatever. Most importantly, the semantic understanding can be reconstructed from the snapshots. 

```ts
function wasRenamed(old, neu){
 return old.name !== neu.name
}
```

```ts
function sizeChanges(old, neu){
  if (old.size !== neu.size){}
}
```

janfjohannes (approved):

PR #3047, opened 2024-08-07
Fix a typo in localised-routing.md

azezsan:
I'm not native but but I think this is the right way to write it?


azezsan:

samuelstroschein:
@azezsan i think you're right :D thanks!

PR #3048, opened 2024-08-07
implement loadProjectFromDirectory plugin.loadMessages logic

martin-lysk:


inlang/source-code/sdk2/src/project/loadProjectFromDirectory.ts
```diff
@@ -28,27 +28,73 @@ export async function loadProjectFromDirectoryInMemory(
 	});
 	await copyFiles({ fs: args.fs, path: args.path, lix: tempProject.lix });
 
-	// TODO loadMessages - call tempProject.lix.settled() to wait for the new settings file
-
-	// TODO loadMessages - since settings are not reactiv yet - just reload the project
+	// TODO call tempProject.lix.settled() to wait for the new settings file, and remove reload of the proejct as soon as reactive settings has landed
+	// NOTE: we need to ensure two things:
+	// 1. settled needs to include the changes from the copyFiles call
+	// 2. the changes created from the copyFiles call need to be realized and lead to a signal on the settings
```
martin-lysk, line 34: I described the issue we are facing with the current setup and left a todo to fix this as soon as settled and reactivty for settings based on changes is implemented. This should not block us from closing https://linear.app/opral/issue/MESDK-163/implement-loadprojectfromdirectory-pluginloadmessages-logic so

inlang/source-code/sdk2/src/project/loadProjectFromDirectory.ts
```diff
@@ -28,27 +28,73 @@ export async function loadProjectFromDirectoryInMemory(
...
-		if (plugin.loadMessages !== undefined) {
-			await loadLegacyMessages({
-				project,
-				fs: args.fs,
-				loadMessagesFn: plugin.loadMessages,
-			});
+	// TODO i guess we should move this validation logic into sdk2/src/project/loadProject.ts
+	// Two scenarios could arise:
+	// 1. set settings is called from an app - it should detect and reject the setting of settings -> app need to be able to validate before calling set
+	// 2. the settings file loaded from disc here is corrupted -> user has to fix the file on disc
```
martin-lysk, line 44: @samuelstroschein I placed the validation - including corresponding tests - in loadProjectFromDirectory for now. I guess we will move it to the setters of settings together with the changes regarding reactivity. So I would keep that out of scope for [this ticket](https://linear.app/opral/issue/MESDK-163/implement-loadprojectfromdirectory-pluginloadmessages-logic) and think we can merge and close MESDK-163

inlang/source-code/sdk2/src/project/loadProjectFromDirectory.ts
```diff
@@ -28,27 +28,73 @@ export async function loadProjectFromDirectoryInMemory(
...
 
-			// TODO check user id and description (where will this one appear?)
```
martin-lysk, line 49: I left this todo not this issue so - to cleanup after the sdk2 v1

inlang/source-code/sdk2/src/bundle-id/bundle-id.ts
```diff
@@ -11,20 +11,20 @@ export function generateBundleId() {
 }
 
 // TODO needed for what?
-// export function humanIdHash(value: string, offset: number = 0) {
-// 	// Seed value can be any arbitrary value
-// 	const seed = 42
+export function humanIdHash(value: string, offset: number = 0) {
```
samuelstroschein, line 14: Rename to `stableBundleId` to increase the understanding of what the function is for?

```suggestion
export function humanIdHash(value: string, offset: number = 0) {
export function stableBundleId(value: string, offset: number = 0){
```

inlang/source-code/sdk2/src/project/util/toMessageV1.test.ts
```diff
@@ -32,21 +32,25 @@ const messageV1: MessageV1 = {
...
 };
 
-const bundle: BundleNested = {
-	alias: {},
-	id: "hello_world",
+const humanReadableId = "awful_lamb_mend_smooth";
+
+const bundle = {
+	alias: {
+		default: "hello_world",
```
samuelstroschein, line 39: Shall we remove the concept of `default` alias? 

The pattern of `default` will break latest with multiple importers. I think it's best to use the plugin key as alias key `{ 'i18next': 'hello_world' }`. There are really only two scenarios: 

1. Multiple aliases exist (default doesn't make sense. Worst of all, default might be overwritten or change conflicts are generated)
2. One alias exists (no default needed. app can derive the alias via `alias[Object.keys(alias)[0]]`)

--- 

For context, we introduced `alias.default` to avoid a breaking change in Paraglide JS (cc @LorisSigrist). Paraglide JS would be able to fallback to the default alias https://github.com/opral/monorepo/commit/68edada6fc45dbaaaf4d2ad13505396c73abe790. As shown though, the concept of a default alias doesn't need to exist for Paraglide to fallback to an alias. 

I even think Paraglide doesn't need a concept of default alias at all to avoid breaking changes. Any alias should be compiled as long as it is a valid JS identifier. Why only compile the `default` alias? Imagine that someone wants to migrate from i18next to Paraglide JS. If Paraglide JS compiles the `i18next` aliases (that are a valid js identifier), the migration is as simple as 

```diff
- t("my_header_button")
+ m.my_header_button()
```

In a second step, Paraglide JS would signal that the bundle id should be used directly. But, a dev can in the first step just use the i18next (or any other JS valid) alias. 

```diff
-m.my_header_button()
+m.blue_moon_shine_world()
```

inlang/source-code/sdk2/src/project/loadProjectFromDirectory.test.ts
```diff
@@ -8,6 +8,78 @@ import type { Text } from "../schema/schemaV2.js";
...
+						],
+					} satisfies ProjectSettings),
+				}).promises as any,
+				path: "./project.inlang",
+				_mockPlugins: mockPluginMapping,
+			});
+		})()
+	).rejects.toThrowError();
+
+	console.log('test')
```
samuelstroschein, line 79: remove the console.log

```suggestion
```

inlang/source-code/sdk2/src/project/loadProjectFromDirectory.ts
samuelstroschein: Given that the "one only load and save messages" can be defined is legacay stuff, only used in `loadProjectFromDirectory`, I would keep it in here. 

Other things in a plugin do not need to be validated since every plugin can define any API they need to (no mergnig happens) and we have TypeScript. Validating a few static props of a plugin can be done with TypeScript. Dynamic function can't be validated anyways.

samuelstroschein (approved):
Can be merged! 

Seeking clarification on bundle.default. Can be done in a separate issue/PR.

inlang/source-code/sdk2/src/project/util/toMessageV1.test.ts
samuelstroschein: Open a dedicated issue for that to not block this PR? 

CleanShot 2024-08-07 at 12 34 29@2x

Also, removing `alias.default` is yet another thing we will not have to explain then. The question came up from someone in an issue already (can't find the link)

CleanShot 2024-08-07 at 12 24 30@2x

martin-lysk: 100% agree 
- ~~lint~~ **validation**rules for the rescue - they will ensure duplicates independent of the alias. 
- if aliases of different exporters/importers match paraglide dedubs them if they differe it makes sense to have for each entry a message function.

samuelstroschein:
manually merged in https://github.com/opral/monorepo/commit/274d27f774ad9eec0c10f0e41b3a2a5b87b4ed45 because rebasing screwed up

PR #3050, opened 2024-08-08
minimal change parent id and csv prototype support for lix sdk

janfjohannes:


lix/packages/sdk/src/open/openLix.ts
```diff
@@ -262,6 +343,8 @@ async function handleFileInsert(args: {
 			neu: args.neu,
 		})
 		for (const diff of diffs ?? []) {
+			const value = diff.neu ?? diff.old
```
janfjohannes, line 346: @samuelstroschein  this was the second bug!

lix/packages/sdk/src/open/openLix.ts
```diff
@@ -149,22 +149,90 @@ async function loadPlugins(db: Kysely<LixDatabase>) {
...
+			.selectFrom("change")
+			.selectAll()
+			.where("commit_id", "=", commit.id)
+			.where((eb) => eb.ref("value", "->>").key("id"), "=", atomId)
+			.where("plugin_key", "=", pluginKey)
+			.where("file_id", "=", fileId)
+			.where("type", "=", diffType)
+			.executeTakeFirst()
+
+		console.log(commit, firstChange)
```
samuelstroschein, line 198: i assume the console.log was not intended

lix/packages/sdk/src/open/openLixInMemory.ts
```diff
@@ -5,14 +5,19 @@ import { openLix } from "./openLix.js"
  *
  */
 export async function openLixInMemory(
-	args: { blob: Blob } & Omit<Parameters<typeof openLix>[0], "database">
+	args: { blob?: Blob; arrayBuffer?: ArrayBuffer } & Omit<Parameters<typeof openLix>[0], "database">
```
samuelstroschein, line 8: Not important now. Before a release we should choose either one. 

Dependent on the perspective `blob` makes more sense than `arrayBuffer` given that files are "blobs". We could also do `openLixInMemory(buffer: Blob | ArrayBuffer, options)`

lix/packages/sdk/src/commit.test.ts
```diff
@@ -0,0 +1,174 @@
...
+		},
+	}
+	const lix = await openLixInMemory({ blob: await newLixFile(), providePlugins: [mockPlugin] })
+
+	const firstRef = await lix.db
+		.selectFrom("ref")
+		.selectAll()
+		.where("name", "=", "current")
+		.executeTakeFirstOrThrow()
+	expect(firstRef.commit_id).toBe("0000-000000-0000")
```
samuelstroschein, line 46: I assume that's a special "first commit" id that adheres to uuid format?

lix/packages/sdk/src/open/openLixInMemory.ts
janfjohannes: @samuelstroschein  its simpler to make a blob into array buffer than other way round as every blob has a blob.arrayBuffer , i would prefer array buffer, as most apis seem to use that we use

lix/packages/sdk/src/open/openLix.ts
janfjohannes: @samuelstroschein  ahh i was looking for that one !

lix/packages/sdk/src/newLix.ts
```diff
@@ -34,15 +40,19 @@ export async function newLixFile(): Promise<Blob> {
...
+        conflict TEXT
+      ) strict;
         
       CREATE TABLE 'commit' (
         id TEXT PRIMARY KEY,
         user_id TEXT NOT NULL,
+        parent_id TEXT NOT NULL,
         description TEXT NOT NULL,
-        zoned_date_time TEXT NOT NULL
+        created TEXT DEFAULT CURRENT_TIMESTAMP NOT NULL
```
samuelstroschein, line 52: Why created as name instead of timestamp?

lix/packages/sdk/src/newLix.ts
```diff
@@ -19,11 +20,16 @@ export async function newLixFile(): Promise<Blob> {
 
 	try {
 		await sql`
+      CREATE TABLE ref (
+        name TEXT PRIMARY KEY,
+        commit_id TEXT
+      );
+
```
samuelstroschein, line 27: Unsure what ref is for. Ignoring it for now to get you into the monorepo

lix/packages/sdk/src/newLix.ts
```diff
@@ -34,15 +40,19 @@ export async function newLixFile(): Promise<Blob> {
         operation TEXT NOT NULL,
         value TEXT,
         meta TEXT,
-        commit_id TEXT
-        ) strict;
+        commit_id TEXT,
+        conflict TEXT
```
samuelstroschein, line 44: Conflict is a reference to another change id right? 

If so, I would expect `conflict_with_change_id` or `conflict_change_id`

lix/packages/sdk/src/schema.ts
```diff
@@ -67,6 +74,8 @@ export type Change = {
 	value?: Record<string, any> & {
 		id: string
 	} // JSONB
+
+	conflict?: any[] | null
```
samuelstroschein, line 78: I recommend using undefined for types (which you accidentally did?). Using `null` opens 3 possible states for a value `value | undefined | null` where `null` might or might not convey different meaning than `undefined`. 

Also unsure how conflict could be an array. If conflict points to a conflict_id, it can only be a string. If conflict is an array of changes, I would be confused. 

```suggestion
	conflict?: any
```

samuelstroschein (approved):

lix/packages/sdk/src/newLix.ts
janfjohannes: @samuelstroschein  not sure what you mean, a name is suppossed to explain what the value is not its datatype. a timestamp could be created modified or any other event of a lifetime of a record.

lix/packages/sdk/src/newLix.ts
janfjohannes: @samuelstroschein    ok!   ref is needed to know where we are, using timestamps is not possible in a production environment for distributed systems the use of timestamp until now was only a prototype hack. ping me if you want to discuss this more.

lix/packages/sdk/src/commit.test.ts
janfjohannes: good point, i make this a valid uuid just to be sure, even though we dont enfornce id formats anywhere at the moment

lix/packages/sdk/src/newLix.ts
janfjohannes: removed the conflict until i the pr for merging.

lix/packages/sdk/src/schema.ts
janfjohannes: @samuelstroschein  i removed the conflict until i the pr for merging. but still unsure about how to type for sql. in sql null is used for empty cells and this is also how kysely works.

lix/packages/sdk/src/newLix.ts
samuelstroschein: I thought we are going to use the `parent_id` property to derive where we are from?

janfjohannes:
@samuelstroschein  i pushed this to lix-integration as we are now working together having your name in the branch seems wrong. eg. i would never merge something into someones private branch

samuelstroschein:
> i pushed this to lix-integration as we are now working together having your name in the branch seems wrong

ergh now everyone needs to switch branches. can we use `samuel/lix-integration` to avoid chaos PRs?

samuelstroschein:
nvm pinged loris and nils but that seemed unnecessary :D

PR #3059, opened 2024-08-12
fix: old neu file passing

samuelstroschein:
## Context 

Closes https://github.com/opral/lix-sdk/issues/33 and likely also closes https://github.com/opral/lix-sdk/issues/34. 

## Changes

1. Fix the argument order of `handle_file_change` 
2. Workaround for awaiting written changes to SQlite. See the changes to `inithandleChangesToLix`. 
3. Deactivate query `getParent()` as the query crashes and is not important for now. 

## Additional info

This is a workaround to unblock Fink2. Should be refactored in the future to handle inlang sdk changes to lix without a setTimeout.


samuelstroschein:
Confirmed. Fixes both bugs. 

@janfjohannes feel free to merge

cc @NilsJacobsen @NiklasBuchfink 


https://github.com/user-attachments/assets/2a6cd19d-822e-45c8-927c-941a9f3f62d9

inlang/source-code/sdk2/src/project/logic/initHandleSaveToLixOnChange.ts
```diff
@@ -16,12 +16,16 @@ export async function initHandleSaveToLixOnChange(args: {
...
+			(async () => {
+				await new Promise((resolve) => setTimeout(resolve, 100));
+				const data = contentFromDatabase(args.sqlite);
+				await args.lix.db
+					.updateTable("file")
+					.set("data", data)
+					.where("path", "=", "/db.sqlite")
+					.execute();
+			})();
+			return;
```
samuelstroschein, line 28: I am confused why this workaround is needed. The triggers are AFTER insert, not BEFORE.

inlang/source-code/sdk2/src/project/logic/initHandleSaveToLixOnChange.ts
```diff
@@ -16,12 +16,16 @@ export async function initHandleSaveToLixOnChange(args: {
...
 		// @ts-expect-error - dynamic function
 		xFunc: () => {
-			const data = contentFromDatabase(args.sqlite);
-			args.lix.db
-				.updateTable("file")
-				.set("data", data)
-				.where("path", "=", "/db.sqlite")
-				.execute();
+			(async () => {
+				await new Promise((resolve) => setTimeout(resolve, 100));
```
janfjohannes, line 20: @samuelstroschein  could you comment why this timeout is required?

lix/packages/sdk/src/open/openLix.ts
```diff
@@ -69,7 +69,7 @@ export async function openLix(args: {
 	await sql`
 	CREATE TEMP TRIGGER file_modified AFTER UPDATE ON file
 	BEGIN
-	  SELECT handle_file_change(NEW.id, NEW.path, NEW.data, OLD.id, OLD.path, OLD.data);
+	  SELECT handle_file_change(OLD.id, OLD.path, OLD.data, NEW.id, NEW.path, NEW.data);
```
samuelstroschein, line 72: Fixes https://github.com/opral/lix-sdk/issues/34

lix/packages/sdk/src/open/openLix.ts
```diff
@@ -249,16 +249,16 @@ async function handleFileChange(args: {
...
-					await getChangeHistory({
-						atomId: value.id,
-						depth: 1,
-						db: args.db,
-						fileId,
-						pluginKey: plugin.key,
-						diffType: diff.type,
-					})
-				)[0];
+				// const parent = (
```
janfjohannes, line 252: this is important, if it crashes it needs to be fixed not disabled. what exactly is the issue eror, so i can look into it?

janfjohannes (changes requested):

inlang/source-code/sdk2/src/project/logic/initHandleSaveToLixOnChange.ts
samuelstroschein: See https://github.com/opral/monorepo/pull/3059#discussion_r1714438603. I don't know why. It seems like the AFTER triggers are triggering before the transaction went through, leading to no diffs.

lix/packages/sdk/src/open/openLix.ts
samuelstroschein: You can look into it. But please unblock @NiklasBuchfink and @NilsJacobsen tmr with Fink2. 

The issue is that there was no change for the query that is executed here. I saw that you intend to create a "0 commit" for a change but that does not account for all sorts of changes a plugin might safe. In other words, I don't think you can create a 0 commit during creation of lix, only during runtime if you encounter a change is has no parent id. At that point though, why bother about a 0 commit? We could define `parent_id === undefined` as "initial change"/start of the history.

janfjohannes: @samuelstroschein  i fixed this with nils and niklas this morning. just for completeness, i dont intend to make a 0 commit, the 0 commit id is only needed to know a repo currently has no commits and to not have to make the parent id of commit optional just for the initial commit. for changes this is not an issue as changes without a parent happen all the time and should be a normal case.

janfjohannes (approved):

lix/packages/sdk/src/open/openLix.ts
samuelstroschein: > the 0 commit id is only needed to know a repo currently has no commits

Why do you need a commit for that and don't query via SQL?

PR #3064, opened 2024-08-16
Add Importer/Exporter plugins

LorisSigrist:
This PR adds support for importer/exporter plugins and adds an example `icu-messageformat-1` importer. 
There is a development-project set up with it where you can try it. 

`loadProject` now has an `importFiles` and `exportFiles` function that can be called. 
`loadProjectFromDirectory` eagerly calls `importFiles`.


LorisSigrist:
My B, wrong branch

PR #3065, opened 2024-08-16
Add Importer/Exporter plugins to the SDK

LorisSigrist:
This PR adds support for importer/exporter plugins and adds an example icu-messageformat-1 importer.
There is a development-project set up with it where you can try it.

loadProject now has an importFiles and exportFiles function that can be called.
loadProjectFromDirectory eagerly calls importFiles.


inlang/source-code/plugins/icu1-importer/package.json
```diff
@@ -0,0 +1,30 @@
+{
+	"name": "@inlang/icu1-importer",
```
samuelstroschein, line 2:
```suggestion
	"name": "@inlang/plugin-icu1",
```

inlang/source-code/plugins/icu1-importer/src/importer.ts
```diff
@@ -0,0 +1,64 @@
+import type { BundleNested, InlangPlugin, ResourceFile } from "@inlang/sdk2"
+import { PluginSettings } from "./settings.js"
+import { createMessage } from "./parse.js"
+
+const importerKey = "importer.inlang.icu-messageformat-1"
+export const importer: InlangPlugin<{
```
samuelstroschein, line 6:
```suggestion
export const plugin: InlangPlugin<{
```

inlang/source-code/plugins/icu1-importer/src/importer.ts
```diff
@@ -0,0 +1,64 @@
+import type { BundleNested, InlangPlugin, ResourceFile } from "@inlang/sdk2"
+import { PluginSettings } from "./settings.js"
+import { createMessage } from "./parse.js"
+
+const importerKey = "importer.inlang.icu-messageformat-1"
```
samuelstroschein, line 5:
```suggestion
const pluginKey = "plugin.inlang.icu-messageformat-1"
```

inlang/source-code/plugins/icu1-importer/src/importer.ts
```diff
@@ -0,0 +1,64 @@
...
+export const importer: InlangPlugin<{
+	[importerKey]: PluginSettings
+}> = {
+	key: importerKey,
+	settingsSchema: PluginSettings,
+
+	toBeImportedFiles: async ({ settings, nodeFs }) => {
+		const files: ResourceFile[] = []
+
+		const folderPath = settings["importer.inlang.icu-messageformat-1"].messagesFolderPath
```
samuelstroschein, line 15:
```suggestion
		const folderPath = settings["plugin.inlang.icu-messageformat-1"].messagesFolderPath
```

inlang/source-code/plugins/icu1-importer/src/settings.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export type PluginSettings = Static<typeof PluginSettings>
+export const PluginSettings = Type.Object({
+	messagesFolderPath: Type.String({
+		pattern: ".*",
+		examples: ["./messages", "./i18n"],
+		title: "Path to folder containing language files",
+		description:
+			"The relative path to the folder containing the language files in your repository. It must contain files named `{locale}.json`.",
+	}),
```
samuelstroschein, line 11: Why don't you use the pathPattern concept like other plugins?

- you would support different patterns like folders, files, namespaces
- aligns with other plugins

```
./messages/{locale}.json
./{locale}/file.json
```

inlang/source-code/plugins/icu1-importer/src/parse.ts
```diff
@@ -0,0 +1,221 @@
...
+	}
+
+	return branches
+}
+
+export function createMessage(
+	messageSource: string,
+	bundleId: string,
+	locale: string
+): MessageNested {
```
samuelstroschein, line 155: (minor, can be ignored)

I recommend using the object as argument pattern to: 

- increase readability
- adding properties in the future without bein dependent on the argument order

```ts
createMessage({
  source: string, 
  bundleId: string, 
  locale: string
})

```

samuelstroschein (changes requested):
Reads good. Needed changes before merging: 

1. change `importer.inlang.*` to `plugin.inlang.*`. we don't differentiate between importer/exporter or other plugins. a plugin is a plugin with certain capabilities like import and/or export.

2. can you add the export function to the icu1 format? roundtrips for sherlock are impossible and in roundtrip tests we might stumble on a few bugs 

3. can you use [pathPattern](https://inlang.com/m/3i8bor92/plugin-inlang-i18next) in the settings of the plugin to increase flexibility and align with other plugins?

 **what i learned**

- good enough stable ids can be derived from the message key (for the bundle, messages, and variants via bundle id = key, message = key + locale, and variant = key + locale + match) 
- the icu1 format seems complicated with the branching logic 
- exposing `project.plugins.get()` instead of a `project.installed.plugins` is easy, enables everything, and therefore good enough

inlang/source-code/plugins/icu1-importer/src/settings.ts
LorisSigrist: Until I changed the API to also pass the `settings` into `importFiles` there would have been no way to tell which file was associated with which language. Now that we have the settings I can revert back to `pathPattern`

samuelstroschein (approved):
Let's get this merged to unblock paraglide. An exporter function for ICU1 can be added with a follow up pr

PR #3067, opened 2024-08-19
Samuel/merge inlang

samuelstroschein:
## Changes

This PR introduces an alternative combine/merge logic, which:

- does not require commits to increase the UX
- does not use heuristics to determine what a conflict is (let's a plugin report instead)
- simplifies the logic for a plugin by splitting "merge/combine" into three functions for a plugin `applyChanges()`, `reportConflicts()`, and (not yet implemented and also optional) `tryResolveConflict()` 

[Here](https://app.excalidraw.com/s/1RmnkzJA3Ph/9GH49NRl7bh) is a flowchart of the logic 

CleanShot 2024-08-19 at 03 05 38@2x

## Learnings

- we likely need a "sequence"/branching concept that determines the order of changes because all changes, even with conflicts, are stored in the change table. 

- move as much logic into plugins for now to see how things generalize. for example, i moved `findParent` etc all into the plugin, which can be extracted over time

- we should build the csv plugin with a patch based approach to see how lix generalizes (https://github.com/opral/lix-sdk/issues/42). @janfjohannes the csv plugin is mentally different than the inlang plugin because no stable ids exist. I suspect that the difference in both plugins is a cause for differently perceived requirements 

- `change.operation` was not needed and is therefore likely not needed at all https://github.com/opral/lix-sdk/issues/41

## Uncertainties   

- no id for a conflict given that a conflict between change A and change B should not be re-raised on re-merging. 

- if "combine" is a better term than merge. should be tested with users. for now, i chose "combine" to avoid potential confusion with git-like merging.


samuelstroschein:
@janfjohannes please review the PR, then let's do a session (if needed & when I am awake again :D) to see if we can further simplify things. 

if you happen to review the PR hours before i am awake, taking a look at a sequence/branching concept would be helpful. given that every change is now stored in the change table, a file/app needs to define the sequence of changes (branch) it should display. this can be manually achieved atm by filtering changes in the history that have conflicts, so no hard requirement. i wonder if it is easy to implement though. 

From https://app.excalidraw.com/s/1RmnkzJA3Ph/9GH49NRl7bh
   
CleanShot 2024-08-19 at 03 22 29@2x

 Good night!

inlang/source-code/sdk2/src/lix-plugin/applyChanges.ts
```diff
@@ -0,0 +1,56 @@
...
+		}
+		// upsert the value
+		else {
+			await db
+				.insertInto(lastChange.type as "bundle" | "message" | "variant")
+				.values(lastChange.value as any)
+				.onConflict((c) => c.column("id").doUpdateSet(lastChange.value as any))
+				.execute();
+		}
+	}
```
samuelstroschein, line 54: This code is the reason why `change.operation` is likely not required. For file formats like inlang where stable ids exist, we can simply upsert.

inlang/source-code/sdk2/src/lix-plugin/reportConflicts.test.ts
```diff
@@ -0,0 +1,183 @@
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
+import { test, expect } from "vitest";
+import { inlangLixPluginV1 } from "./inlangLixPluginV1.js";
+import { newLixFile, openLixInMemory, type Change } from "@lix-js/sdk";
+
+test("a create operation should not report a conflict given that the change does not exist in target", async () => {
+	const targetLix = await openLixInMemory({ blob: await newLixFile() });
+	const sourceLix = await openLixInMemory({ blob: await newLixFile() });
```
samuelstroschein, line 8: I used `target` and `source` as terms in the hope that `target` is signalling "aha we are copy changes into the target". 

@janfjohannes what do you think about the terms?

inlang/source-code/sdk2/src/lix-plugin/reportConflicts.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export const reportConflicts: LixPlugin["reportConflicts"] = async ({
+	sourceLix,
+	targetLix,
+}) => {
+	const result: Conflict[] = [];
+	const changesNotInTarget = await getChangesNotInTarget({
+		sourceLix,
+		targetLix,
+	});
```
samuelstroschein, line 14: Letting a plugin deal with all logic on how to derive conflicts likely reduced the risk atm of building wrong abstractions that don't generalize. 

See https://github.com/opral/lix-sdk/issues/42 as an example. A CSV plugin would need to capture dependencies between changes and retrieve / dynamically compute snapshots based on the dependencies.

lix/packages/sdk/src/newLix.ts
```diff
@@ -58,6 +58,14 @@ export async function newLixFile(): Promise<Blob> {
         meta TEXT,
         commit_id TEXT
       ) strict;
+
+      CREATE TABLE conflict (
+        change_id TEXT NOT NULL,
+        conflicts_with_change_id TEXT NOT NULL,
+        reason TEXT,
+        meta TEXT,
+        PRIMARY KEY (change_id, conflicts_with_change_id)
```
samuelstroschein, line 67: Unsure about the primary key here and if a conflict shouldn't have an ID. The problem with an ID would be re-generation of a conflict that is already stored in the conflict table.

inlang/source-code/sdk2/src/lix-plugin/reportConflicts.test.ts
janfjohannes: sounds clear and good. we need similar good terms instead of old and neu asap

lix/packages/sdk/src/combine/combine-changes.ts
```diff
@@ -0,0 +1,106 @@
...
+			throw new Error("Plugin does not support applying changes");
+		}
+
+		const { fileData } = await plugin.applyChanges({
+			changes,
+			file,
+			lix: args.target,
+		});
+
+		await args.target.db
```
janfjohannes, line 100: this will   trigger the change generation logic again, if a change is used to generate the file instead of the other way around the internal_file table needs to be used

lix/packages/sdk/src/combine/combine-changes.ts
```diff
@@ -0,0 +1,106 @@
+import type { LixPlugin } from "../plugin.js";
+import type { Change } from "../schema.js";
+import type { Lix } from "../types.js";
+
+/**
+ * Combines the changes of two lixes into the target lix.
+ */
+export async function combineChanges(args: { target: Lix; source: Lix }): Promise<void> {
+	const toBeAppliedChanges: Record<LixPlugin["key"], Array<Change>> = {};
```
janfjohannes, line 9: this needs to be in a transaction , otherwise the db will be in inconsistent state in case of errors

lix/packages/sdk/src/combine/combine-changes.ts
```diff
@@ -0,0 +1,106 @@
...
+
+	// TODO ideally a right join via sqlite attaching works
+	//      instead of fetching all changes from source
```
janfjohannes, line 12: what happens if source allready has conflicts?

lix/packages/sdk/src/combine/combine-changes.ts
```diff
@@ -0,0 +1,106 @@
...
+			.select("id")
+			.where("id", "=", change.id)
+			.executeTakeFirst();
+
+		// change already exists, skip
+		if (targetChange !== undefined) {
+			continue;
+		}
+		// copy change to target
+		await args.target.db
```
janfjohannes, line 31: this will lead to missing commits from source db in target db.

inlang/source-code/sdk2/src/lix-plugin/reportConflicts.test.ts
samuelstroschein: I opened an issue for old/neu https://github.com/opral/lix-sdk/issues/44

lix/packages/sdk/src/combine/combine-changes.ts
janfjohannes: also this will only work with confirmed branch support in the existing implementation so all changes not referenced by the current commit and change lineage are ignored.

lix/packages/sdk/src/combine/combine-changes.ts
janfjohannes: not clear how this is supposed to work, the commits need to be integrated or the changes need to be linked somehow.

lix/packages/sdk/src/plugin.ts
```diff
@@ -1,11 +1,40 @@
...
+	// 	bundle: Bundle,
+	// 	message: Message,
+	// 	variant: Variant,
+	// },
+	/**
+	 * Reports if a change from the source lix conflicts with a change in the target lix.
+	 *
+	 * - returns `undefined` if no conflict exists
+	 */
+	reportConflicts?: (args: {
```
janfjohannes, line 24: better detectConflicts? "report" is something user readable for me or meant to be displayed

lix/packages/sdk/src/plugin.ts
```diff
@@ -1,11 +1,40 @@
...
+		sourceLix: LixReadonly;
```
janfjohannes, line 25: i would at lest expect an argument with the "only in source" leaf changes which i think is always needed

janfjohannes (changes requested):

lix/packages/sdk/src/combine/combine-changes.ts
```diff
@@ -0,0 +1,106 @@
...
+		if (!plugin) {
+			throw new Error("Plugin not found");
+		}
+
+		if (!plugin.reportConflicts) {
+			throw new Error("Plugin does not support conflict reporting");
+		}
+
+		const conflict = await plugin.reportConflicts({
+			sourceLix: args.source,
```
janfjohannes, line 52: isnt n argument missing here? this looks like it regenerates all conflicts on every change but would always lead to the same list of conflicts?

lix/packages/sdk/src/newLix.ts
samuelstroschein: Agree. Changed it https://github.com/opral/monorepo/pull/3067/commits/537d804abbdb1324a9b3f1ad3e4ca1345102679e

lix/packages/sdk/src/plugin.ts
```diff
@@ -1,11 +1,40 @@
...
+		targetLix: LixReadonly;
+	}) => Promise<Conflict[]>;
+	applyChanges?: (args: {
+		lix: LixReadonly;
+		file: LixFile;
+		changes: Array<Change>;
+	}) => Promise<{
+		fileData: LixFile["data"];
+	}>;
+	tryResolveConflict?: () => Promise<
```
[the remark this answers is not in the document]
samuelstroschein: That's just a placeholder

inlang/source-code/sdk2/src/lix-plugin/reportConflicts.test.ts
```diff
@@ -0,0 +1,183 @@
...
+	};
+	await sourceLix.db.insertInto("change").values([change]).execute();
+	const conflicts = await inlangLixPluginV1.reportConflicts!({
+		sourceLix,
+		targetLix,
+	});
+	expect(conflicts).toHaveLength(0);
+});
+
+test("it should report deletions as a conflict if the parent of the target and source are not identical", async () => {
```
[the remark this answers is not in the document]
samuelstroschein: I am confused too :D Will add a todo

inlang/source-code/sdk2/src/lix-plugin/reportConflicts.test.ts
```diff
@@ -0,0 +1,183 @@
...
+		.insertInto("change")
+		.values([...commonChanges, ...changesOnlyInSource])
+		.execute();
+
+	await targetLix.db
+		.insertInto("change")
+		.values([...commonChanges, ...changesOnlyInTarget])
+		.execute();
+
+	// const conflict = await inlangLixPluginV1.reportConflicts!({
```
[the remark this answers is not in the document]
samuelstroschein: added as todo.

inlang/source-code/sdk2/src/lix-plugin/utilities/getChangesNotInTarget.test.ts
```diff
@@ -0,0 +1,42 @@
...
+import { getChangesNotInTarget } from "./getChangesNotInTarget.js";
+
+test("it should find the changes that are not in target", async () => {
+	const sourceLix = await openLixInMemory({
+		blob: await newLixFile(),
+	});
+	const targetLix = await openLixInMemory({
+		blob: await newLixFile(),
+	});
+	const commonChanges: Change[] = [
```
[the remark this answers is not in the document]
samuelstroschein: Added

janfjohannes:
@samuelstroschein  i am not sure a csv usecase without ids is a good second usecase. the realiance on many heuristics will distract us and led to explodin complexities

janfjohannes:
@samuelstroschein  anything with an id. eg excalidraw json files? or csv files but with the assumption of having id column.

its not possible to realiably merge anything that does not have ids, therefore these have to contain complex heuristics which will eat dev resources .

samuelstroschein:
@janfjohannes hmm excalidraw isn't widely used. ideally, we can build a plugin and app that leads to a "wow" moment and actual usage e.g. `csv.opral.com`. 

do you foresee no risk in not developing a plugin an unstrcutured file format like csv? for example, poking around with the csv already led to https://github.com/opral/lix-sdk/issues/42 (snapshots seem to first class)

lix/packages/sdk/src/plugin.ts
samuelstroschein: Yep, will change! Good wording.

inlang/source-code/sdk2/src/lix-plugin/utilities/getFirstCommonParent.ts
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	const sourceChangeExistsInTarget = await args.targetLix.db
+		.selectFrom("change")
+		.selectAll()
+		.where("id", "=", args.sourceChange.id)
+		.executeTakeFirst();
+
+	if (sourceChangeExistsInTarget) {
+		throw Error(
```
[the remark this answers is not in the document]
samuelstroschein: implemented. Renamed the function to `getLowestCommonAncestor()` to signal that a parent is not necessarily returned and to align with https://en.wikipedia.org/wiki/Lowest_common_ancestor

lix/packages/sdk/src/combine/combine-changes.ts
samuelstroschein: implemented

lix/packages/sdk/src/combine/combine-changes.ts
samuelstroschein: The conflict is ignored as it is tracked already. I made this explicit and added a test https://github.com/opral/monorepo/pull/3067/commits/6aeb7408f7fa8e8c783c0da867765b53b4f6c8f5

lix/packages/sdk/src/plugin.ts
samuelstroschein: done

lix/packages/sdk/src/combine/combine-changes.ts
samuelstroschein: copying the commits over has been implemented

lix/packages/sdk/src/combine/combine-changes.ts
samuelstroschein: solved

samuelstroschein:
🥳 🎉 merging works as expected https://www.loom.com/share/28c55a5be891412d9d21304048cd4239

**Context**

- i implemented every (?) feedback you (@janfjohannes) had
- no sequence concept was required to enable merging in lix by filtering changes with conflicts
- lots of todos and unoptimized, potentially unreliable code that can be improved in follow-up PRs. it's good enough for now, as demonstrated in the loom 

@janfjohannes I rec to merge this PR (on my behalf given that I'm going to bed now) to unblock fink2 (cc @NiklasBuchfink @NilsJacobsen) from building merge related workflows and UIs.

**Next steps (in follow-up PRs)**

- https://github.com/opral/lix-sdk/issues/49
- https://github.com/opral/inlang-sdk/issues/133

@janfjohannes I propose you merge this PR and pick either one of the issues while I am asleep. Those are the two last blockers for fink afaik. The rest, like sequence, can/should wait on UX testing.

lix/packages/sdk/src/query-utilities/get-leaf-change.ts
```diff
@@ -0,0 +1,29 @@
...
+	const _true = true;
+
+	let nextChange = args.change;
+
+	while (_true) {
+		const childChange = await args.lix.db
+			.selectFrom("change")
+			.selectAll()
+			.where("parent_id", "=", nextChange.id)
+			.executeTakeFirst();
```
janfjohannes, line 19: there might be more than one child

lix/packages/sdk/src/query-utilities/get-lowest-common-ancestor.ts
```diff
@@ -0,0 +1,54 @@
+import type { Change, LixReadonly } from "@lix-js/sdk";
+
+/**
+ * Finds the lowest common change of the source
+ * change in the source lix and the target lix.
+ *
+ * https://en.wikipedia.org/wiki/Lowest_common_ancestor
```
janfjohannes, line 7: if you need to add a wikipedia link to some math article, the naming is bad. i dont see any reason to add "lowest"

lix/packages/sdk/src/combine/combine-changes.ts
janfjohannes: @samuelstroschein  you cannot just ignore conflicts , this will lead to dataloss. the conflict is tracked in the source, not in the target!

lix/packages/sdk/src/query-utilities/get-lowest-common-ancestor.ts
samuelstroschein: Because you can have multiple common ancestors. Which one do you want? The lowest one in the tree. Hence, the name

janfjohannes: @samuelstroschein  no one would expect to get anything except the first common ancestor. this is how git and any other tool work.

lix/packages/sdk/src/merge/merge.ts
```diff
@@ -0,0 +1,133 @@
...
+
+	const { fileData } = await plugin.applyChanges({
+		changes: nonConflictingLeafChangesInSource,
+		file,
+		lix: args.target,
+	});
+
+	await args.target.db.transaction().execute(async (trx) => {
+		if (sourceChanges.length > 0) {
+			// 1. copy the changes from source
```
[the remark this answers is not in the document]
janfjohannes: @samuelstroschein  how would you get the last change for message id X after merge?

lix/packages/sdk/src/query-utilities/get-lowest-common-ancestor.ts
samuelstroschein: That doesn't seem true. Git does not use the first ancestor but the "best common ancestor" https://git-scm.com/docs/git-merge-base. And here we already have the problem. What common ancestor are you looking for? 

I'll keep the name lowest common ancestor because it describes the query. Thereby, we can add `findBestCommonAncestor` or other queries in the future. 

PS we should name it `find` instead of `get` to align with tree terminology.

PR #3069, opened 2024-08-22
Sherlock on `@inlang/sdk2`

felixhaeberle:
Starting work on integrating `@inlang/sdk2` into Sherlock.

TODO:
- [x] transform all vscode actions + tests to SDK v2
- [x] transform all vscode commands + tests  to SDK v2
  - [ ] copyError.test.ts
  - [x] copyError.ts
  - [ ] createMessage.test.ts
  - [x] createMessage.ts
  - [ ] editMessage.test.ts
  - [x] editMessage.ts
  - [ ] extractMessage.test.ts
  - [x] extractMessage.ts
  - [ ] jumpToPosition.test.ts
  - [x] jumpToPosition.ts
  - [ ] machineTranslate.test.ts
  - [x] machineTranslate.ts
  - [ ] openInFink.test.ts
  - [x] openInFink.ts
  - [ ] openProject.test.ts
  - [ ] openProject.ts
  - [ ] openSettingsFile.test.ts
  - [ ] openSettingsFile.ts
  - [ ] openSettingsView.ts
  - [ ] previewLocaleCommand.test.ts
  - [ ] previewLocaleCommand.ts
  - [ ] toggleInlineAnnotation.test.ts
  - [ ] toggleInlineAnnotation.ts
- [x] transform all vscode decorations + tests  to SDK v2
- [x] transform all vscode diagnostics + tests  to SDK v2
- [ ] transform all services + tests  to SDK v2
- [ ] transform all utility functions + tests  to SDK v2


felixhaeberle:
@samuelstroschein 

Following questions:

1. Is `NodeishFileSystem` legacy and now only `typeof import("node:fs/promises")` is used?

2. Does that mean everything has to be rewritten / remapped (all Sherlock functions, all cross-selling packages etc.)?

3. I need `listProject`, can I put it in SDK v2 or should I move it to Sherlock?

4. Can you translate this query into sqlite format so I can see how it works?

```ts
const _message =
state().project.query.messages.get({
where: { id: message.messageId },
}) ?? state().project.query.messages.getByDefaultAlias(message.messageId)
```

5. What is the solution for `project.customApi()["app.inlang.ideExtension"]` ?

`customApi()` doesn't seem to exist anymore, only in a project.plugins.get() which points to the `InlangPlugin` interface (unresolved).

6. When do I have to use the `project.db` vs the `project.lix` API?

7. When does the work on validation rules start & where does it get tracked? Sherlock can't be release on sdk v2 without those as they mark a major functionality/feature.

samuelstroschein:
> Is NodeishFileSystem legacy and now only typeof import("node:fs/promises") is used?

Yes. You can use the node fs directly. 

> Does that mean everything has to be rewritten / remapped (all Sherlock functions, all cross-selling packages etc.)?

Jein. NodishFs was a subset of node:fs/promises. Which means that you shouldn't have a problem. Problematic could be the overloads. Created an issue to remove the overloads https://github.com/opral/inlang-sdk/issues/136

> I need listProject, can I put it in SDK v2 or should I move it to Sherlock?

You need to do it in Sherlock with a globbing function. See https://www.npmjs.com/package/fast-glob#onlydirectories as an example. 

```ts
const projects = fg.sync('*.inlang', { onlyDirectories: true }); 
```

For context, lix/fs had no globbing support, leading to `listProjects`. `listProjects` emulated a glob. cc @LorisSigrist this applies to you as well. 

> Can you translate this query into sqlite format so I can see how it works?

Yes, reference the docs of kyseley https://kysely.dev/docs/examples/select/a-single-column. 

Here is the translated query (hope it works). Keep in mind that `alias.default` has been removed. What you want to query is "give me a message where the id is this or one of the aliases is the id (not only querying the default alias)

```ts
const bundles = await project.db
  .selectFrom('bundle')
  .selectAll()
  // eb = expression builder
  .where((eb) => eb.or([
    // either the ids is equal to the message id
    eb("id", "=", message.messageId)
    // or the default alias (alias is a json that's why the --> syntax) is equal to the messageId
    eb(eb.ref("alias", "-->").key("default"), "=", message.messageId)
  ]))
  .execute()
```

Or with the `selectBundleNested` helper query:

```ts
const bundles = await selectBundleNested(project.db)
  // eb = expression builder
  .where((eb) => eb.or([
    // either the ids is equal to the message id
    eb("id", "=", message.id)
    // or the default alias (alias is a json that's why the --> syntax) is equal to the messageId
    eb(eb.ref("alias", "-->").key("default"), "=", message.id)
  ]))
  .execute()
```

> What is the solution for project.customApi()["app.inlang.ideExtension"] ?

1. `addCustomApi()` will be deprecated due to https://github.com/opral/inlang-sdk/issues/125
2. Only sherlock uses the custom API feature, hence here is a proposed workaround: 

```ts
const plugin = project.plugins.get().find((p) => p.addCustomApi)
const customApi = plugin.addCustomApi()
customApi['app.inlang.ide-extension']
```

> When do I have to use the project.db vs the project.lix API?

Everything change control related (history, commit, etc.) is queried via `project.lix`. The latest state of an inlang project (bundles, messages , variants) is queried via `project.db`. In the future, we might be able to offer one API under `project.db` which can join over the lix database. 

> When does the work on validation rules start & where does it get tracked? Sherlock can't be release on sdk v2 without those as they mark a major functionality/feature.

Here is the lix validation rule project https://linear.app/opral/project/lix-validation-rules-bdf46c532aa1/overview. 

We agreed on releasing the inlang SDK2 without validating rules. And validation rules are likely 2 months away. The value of inlang SDK 2 with variants is higher than the interim loss of validation rules. 

I propose building Sherlock SDK2 and, once it's ready, asking users if they prefer a Sherlock release with variant support but no interim validation rule support.  

> What happens to _import on the loadProject interface (which is likely needed by Sherlock for importing modules)?

```ts
await loadProjectFromDirectoryInMemory({
  preprocessPlugin: yourPreprocessFunction(),
})

```

samuelstroschein (commented):
Looks as expected. Use a glob libary for list projects

samuelstroschein:
@felixhaeberle i merged https://github.com/opral/monorepo/pull/3070. you can now use `preprocessPluginsBeforeImport` in `loadProjectFromDirectory`

PR #3074, opened 2024-08-26
add: `change.author`

samuelstroschein:
## Context

Closes https://github.com/opral/inlang-sdk/issues/133. 

### Changes

- adds `change.author` 
- refactors `commit.user_id` to `commit.author` for alignment with change.author
- marks `commit.author` as nullable (if a user did not "log in" or specified who they are, no author exists)

### Decisions

- only handles singular author case to avoid longer decision-making on https://github.com/opral/lix-sdk/issues/56 to unblock which will block fink
- does not introduce an `Account` table that author references for foreign key consistency because of https://github.com/opral/lix-sdk/issues/56

### DIFF
--- a/lix/packages/sdk/src/change-queue.test.ts
+++ b/lix/packages/sdk/src/change-queue.test.ts
@@ -1,7 +1,7 @@
-import { expect, test } from "vitest";
+import { expect, test, vi } from "vitest";
 import { openLixInMemory } from "./open/openLixInMemory.js";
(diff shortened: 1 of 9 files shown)


samuelstroschein:
@janfjohannes if you approve this pr, merge it to unblock @NiklasBuchfink and @NilsJacobsen while i am asleep

janfjohannes (approved):

PR #3087, opened 2024-09-03
fixes unawaited getters, and wrong types for matchers in paraglide - …

martin-lysk:
…still todo left

@samuelstroschein ParaglideJs doesn't handle named matchers - it still expects the matchers to be an array and use the index to map selectors. I fixed various types and mocks but want to focus on the loadProjectFromDirectory and since i am not sure whats the state of paraglide is i just made it compile with a todo - ping me if i can support


samuelstroschein:
@martin-lysk merge conflicts everywhere. i am fixing the issues on `lix-intergration` directly. this pr can be closed

samuelstroschein:
Closing because stale. Re-open if you think the close was wrong

PR #3093, opened 2024-09-05
telemetry

samuelstroschein:
## Context 

Closes telemetry and https://github.com/opral/inlang-sdk/issues/166. 

## Changes & Decisions

- adds `settings.telemetry : off` as settings similar to vscode https://code.visualstudio.com/docs/getstarted/telemetry
- adds env variables that are required for telemetry 
- adds SDK_VERSION to the `SDK loaded project` event to know what sdk versions users are on
- adds numbers of bundles, messages, and variants to know how large projects are

### DIFF
--- a/inlang/source-code/sdk2/src/project/loadProject.ts
+++ b/inlang/source-code/sdk2/src/project/loadProject.ts
@@ -10,6 +10,7 @@ import { createProjectState } from "./state/state.js";
 import { withLanguageTagToLocaleMigration } from "../migrations/v2/withLanguageTagToLocaleMigration.js";
 import { exportFiles, importFiles } from "../import-export/index.js";
 import { v4 } from "uuid";
+import { maybeCaptureLoadedProject } from "./maybeCaptureTelemetry.js";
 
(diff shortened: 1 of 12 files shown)


inlang/source-code/sdk2/src/project/loadProject.ts
```diff
@@ -38,6 +39,24 @@ export async function loadProject(args: {
...
+	applicationId?: string;
+	/**
+	 * The version of the application that is using the SDK.
+	 *
+	 * This is used for telemetry purposes. To derive insights like
+	 * which version of the SDK is used, how many projects are loaded, etc.
+	 *
+	 * The application version can be removed at any time in the future.
+	 */
+	applicationVersion?: string;
```
samuelstroschein, line 59: To workaround the posthog `appVersion` problems, i used `applicationVersion`

jannesblobel: The problem, when we change appID, all our insights will crash.

PR #3099, opened 2024-09-07
Add: Import messages from local files

samuelstroschein:
Closes https://github.com/opral/inlang-sdk/issues/174. Messages can now be loaded in Sherlock (cc @felixhaeberle)

https://github.com/user-attachments/assets/2d46814e-bb79-4610-82a8-9b7c94bb37fc

**Changes**

- `plugin.loadMessages` receives a mapped fs that turns relative paths into absolute paths to maintain backward compatibility
- sherlock subscribes to bundles to display messages 

**Follow up todos**

- [ ] pass mapped fs to `plugin.saveMessages` and `plugin.toBeImportedFiles()` after https://github.com/opral/inlang-sdk/issues/159 is merged

### DIFF
--- a/inlang/development-projects/inlang-nextjs/other-folder/backend.inlang/settings.json
+++ b/inlang/development-projects/inlang-nextjs/other-folder/backend.inlang/settings.json
@@ -11,10 +11,10 @@
 	],
 	"plugin.inlang.i18next": {
 		"pathPattern": {
-			"client-page": "./../../app/i18n/locales/{languageTag}/client-page.json",
-			"footer": "./../../app/i18n/locales/{languageTag}/footer.json",
-			"second-page": "./../../app/i18n/locales/{languageTag}/second-page.json",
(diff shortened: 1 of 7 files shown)


felixhaeberle:
@samuelstroschein thank you for making this available in the sdk:) 

Probably only the video but it feels very laggy now, have to test myself tomorrow. But it is that laggy in real-time as in the video, we can't release it to our user without performance improvements.

---

Kinda related to this issue: I'm wondering why we haven't thought about the absolute path resolving before — and therefore wondering if there are any other concepts we might have missed from the previous implementation of sdk v1. 

Is sdk v2 passing all the test for sdk v1?

This would be highly beneficial to have more confidence on backwards compatibility. @martin-lysk what's your take on this?

samuelstroschein:
> and therefore wondering if there are any other concepts we might have missed from the previous implementation of sdk v1.

Just test sherlock with existing projects. bugfixes will quickly be accepted. web apps don't need anything from SDK v1. hence, we haven't copied the tests over.

PR #3119, opened 2024-09-12
use uuid 7 also when creating a new variant

martin-lysk:
While uuid created on the db was changed to uuid v7 in #3118 
The uuid in create Variant was still using the uuid v3 - this pr fixes this

### DIFF
--- a/inlang/source-code/sdk2/src/helper.ts
+++ b/inlang/source-code/sdk2/src/helper.ts
@@ -1,11 +1,10 @@
-import { v4 as uuid } from "uuid";
+import { v7 as uuid } from "uuid";
 import type { ProjectSettings } from "./json-schema/settings.js";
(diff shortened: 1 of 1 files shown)


felixhaeberle:
@martin-lysk just saw this. should we try to leverage v7 in all of the sdk code then?

Bildschirmfoto 2024-09-12 um 12 51 10

samuelstroschein (approved):

samuelstroschein:
> @felixhaeberle  just saw this. should we try to leverage v7 in all of the sdk code then?

Yes. Or, remove uuid in runtime code where possible. SQlite will automatically create a UUID v7 for entities without an id

martin-lysk:
will add them

samuelstroschein:
Closing because outdated.

PR #3121, opened 2024-09-12
adds discussions to lix

martin-lysk:
This adds discussions and comments to lix

Notice this is WIP to see what we should add

### DIFF
--- a/lix/packages/sdk/src/database/createSchema.ts
+++ b/lix/packages/sdk/src/database/createSchema.ts
@@ -77,5 +77,39 @@ export async function createSchema(args: { db: Kysely<any> }) {
   BEGIN
     insert or replace into file_internal(id, path, data, metadata) values(OLD.file_id, OLD.path, OLD.data, OLD.metadata);
   END;
+
+
+  -- change discussions 
+
+  CREATE TABLE discussion (
+    -- TODO in inlang i saw we replace uuid_v3 with uuid_v7 any reason we use v4 in lix 
+    id TEXT PRIMARY KEY DEFAULT (uuid_v4())
+  ) strict;
+
+  CREATE TABLE discussion_change_map (
+    discussion_id TEXT NOT NULL,
+    change_id TEXT NOT NULL,
+
+    FOREIGN KEY(discussion_id) REFERENCES discussion(id),
+    
+    -- NOTE this will prevent us from dropping changes 
+    FOREIGN KEY(change_id) REFERENCES change(id)
+  ) strict;
+
+
+
+  CREATE TABLE comment (
+    --- TODO in inlang i saw we replace uuid_v3 with uuid_v7 any reason we use v4 in lix?
+    id TEXT PRIMARY KEY DEFAULT (uuid_v4()),
+    parent_id TEXT,
+    discussion_id TEXT NOT NULL,
+    author_id TEXT NOT NULL, 
+    created_at TEXT DEFAULT CURRENT_TIMESTAMP NOT NULL,
+    body TEXT NOT NULL,
+
+    FOREIGN KEY(discussion_id) REFERENCES discussion(id)
+  ) strict;
+
 `.execute(args.db);
 }
+
--- a/lix/packages/sdk/src/database/schema.ts
+++ b/lix/packages/sdk/src/database/schema.ts
@@ -9,6 +9,11 @@ export type LixDatabaseSchema = {
 	file_internal: LixFileTable;
 	change_queue: ChangeQueueTable;
 	conflict: ConflictTable;
+
+	// discussion
+	discussion: DiscussionTable;
+	comment: CommentTable;
+	discussion_change_map: DiscussionChangeMapTable;
 };
 
 export type Ref = Selectable<RefTable>;
@@ -132,4 +137,33 @@ type ConflictTable = {
 	 * that resulted from a merge.
 	 */
 	resolved_with_change_id?: ChangeTable["id"];
+};
+
+// ------ discussions ------
+
+export type Discussion = Selectable<DiscussionTable>;
+export type NewDiscussion = Insertable<CommitTable>;
+export type DiscussionUpdate = Updateable<DiscussionTable>;
+type DiscussionTable = {
+	id: Generated<string>;
+};
+
+export type DiscussionChangeMap = Selectable<DiscussionChangeMapTable>;
+export type NewDiscussionChangeMap = Insertable<DiscussionChangeMapTable>;
+export type DiscussionChangeMapUpdate = Updateable<DiscussionChangeMapTable>;
+type DiscussionChangeMapTable = {
+	change_id: string;
+	discussion_id: string;
+};
+
+export type Comment = Selectable<CommentTable>;
+export type NewComment = Insertable<CommentTable>;
+export type CommentUpdate = Updateable<CommentTable>;
+type CommentTable = {
+	id: string;
+	parent_id?: string;
+	discussion_id: string;
+	author_id: string;
+	created_at: Generated<string>;
+	body: string;
 };
--- a/lix/packages/sdk/src/discussion/comment.ts
+++ b/lix/packages/sdk/src/discussion/comment.ts
@@ -1,0 +1,36 @@
+import { Kysely } from "kysely";
+import type { LixDatabaseSchema } from "../database/schema.js";
+import { v4 } from "uuid";
+
+export async function comment(args: {
+	db: Kysely<LixDatabaseSchema>;
+	currentAuthor: string;
+	parentComment: { id: string };
+	body: string;
+}) {
+	const newCommentId = v4();
+
+	return args.db.transaction().execute(async (trx) => {
+		// verify that the parent comment exists and fetch the discussion_id
+		const { discussion_id } = await trx
+			.selectFrom("comment")
+			.select("discussion_id")
+			.where("id", "=", args.parentComment.id)
+			.executeTakeFirstOrThrow();
+
+		const comment = await trx
+			.insertInto("comment")
+			.values({
+				id: newCommentId,
+				parent_id: args.parentComment.id,
+				discussion_id,
+				author_id: args.currentAuthor,
+				// todo - use zoned datetime
+				body: args.body,
+			})
+			.returning("id")
+			.executeTakeFirstOrThrow();
+
+		return comment;
+	});
+}
--- a/lix/packages/sdk/src/discussion/start-discussion.ts
+++ b/lix/packages/sdk/src/discussion/start-discussion.ts
@@ -1,0 +1,68 @@
+import { Kysely } from "kysely";
+import type { LixDatabaseSchema } from "../database/schema.js";
+import { v4 } from "uuid";
+
+export async function startDiscussion(args: {
+	db: Kysely<LixDatabaseSchema>;
+	currentAuthor: string;
+	changes: { id: string }[];
+	body: string;
+}) {
+	const newDiscussionId = v4();
+	const newCommentId = v4();
+
+	return args.db.transaction().execute(async (trx) => {
+		// verify changes exists
+		if (args.changes.length > 0) {
+			console.log("checking changes");
+			const { count } = (await trx
+				.selectFrom("change")
+				.select(trx.fn.countAll().as("count"))
+				// NOTE: for high amount of changes this will be extremly costy since its a huge or clause
+				.where(
+					"change.id",
+					"in",
+					args.changes.map((change) => change.id),
+				) // NOTE: this will need to be batched if the amount of changes passed is bigger than sqlite max parameter count
+				.executeTakeFirstOrThrow()) as { count: number }; // TODO check if this cast is ok - it coult be a string/bigint or number
+
+			if (count !== args.changes.length) {
+				throw new Error("unknown changes passed");
+			}
+		}
+		
+		const discussion = await trx
+			.insertInto("discussion")
+			.values({
+				id: newDiscussionId,
+			})
+			.returning("id")
+			.executeTakeFirstOrThrow();
+
+		// create mappinf from discussion to changes
+		for (const change of args.changes) {
+			await trx
+				.insertInto("discussion_change_map")
+				.values({
+					change_id: change.id,
+					discussion_id: newDiscussionId,
+				})
+				.executeTakeFirstOrThrow();
+		}
+
+		const comment = await trx
+			.insertInto("comment")
+			.values({
+				id: newCommentId,
+				parent_id: undefined,
+				discussion_id: newDiscussionId,
+				author_id: args.currentAuthor,
+				// todo - use zoned datetime
+				body: args.body,
+			})
+			.returning("id")
+			.executeTakeFirstOrThrow();
+
+		return discussion;
+	});
+}
--- a/lix/packages/sdk/src/open/openLix.ts
+++ b/lix/packages/sdk/src/open/openLix.ts
@@ -1,5 +1,7 @@
 import type { LixPlugin } from "../plugin.js";
 import { commit } from "../commit.js";
+import { startDiscussion } from "../discussion/start-discussion.js";
+import { comment } from "../discussion/comment.js";
 import { handleFileChange, handleFileInsert } from "../file-handlers.js";
 import { loadPlugins } from "../load-plugin.js";
 import { contentFromDatabase, type SqliteDatabase } from "sqlite-wasm-kysely";
@@ -170,6 +172,22 @@ export async function openLix(args: {
 		},
 		commit: (args: { description: string }) => {
 			return commit({ ...args, db, currentAuthor });
+		},
+		startDiscussion: (args: { changes: { id: string }[]; body: string }) => {
+			if (currentAuthor === undefined) {
+				throw new Error("current author not set");
+			}
+			return startDiscussion({
+				...args,
+				db,
+				currentAuthor,
+			});
+		},
+		comment: (args: { parentComment: { id: string }; body: string }) => {
+			if (!currentAuthor) {
+				throw new Error("current author not set");
+			}
+			return comment({ ...args, db, currentAuthor });
 		},
 	};
 }
--- a/lix/packages/sdk/src/discussion/start-discussion.test.ts
+++ b/lix/packages/sdk/src/discussion/start-discussion.test.ts
@@ -1,0 +1,173 @@
+/* eslint-disable @typescript-eslint/no-non-null-asserted-optional-chain */
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
+import { expect, expectTypeOf, test } from "vitest";
+import { openLixInMemory } from "../open/openLixInMemory.js";
+import { newLixFile } from "../newLix.js";
+import type { LixPlugin } from "../plugin.js";
+
+const mockPlugin: LixPlugin = {
+	key: "mock-plugin",
+	glob: "*",
+	diff: {
+		file: async ({ old, neu }) => {
+			return [
+				!old
+					? {
+							type: "text",
+							operation: "create",
+							old: undefined,
+							neu: {
+								id: "test",
+								text: "inserted text",
+							},
+						}
+					: {
+							type: "text",
+							operation: "update",
+							old: {
+								id: "test",
+								text: "inserted text",
+							},
+							neu: {
+								id: "test",
+								text: "updated text",
+							},
+						},
+			];
+		},
+	},
+};
+
+test("should be able to start a discussion on changes", async () => {
+	const lix = await openLixInMemory({
+		blob: await newLixFile(),
+		providePlugins: [mockPlugin],
+	});
+
+	lix.currentAuthor.set("Test User");
+
+	const enc = new TextEncoder();
+
+	await lix.db
+		.insertInto("file")
+		.values({ id: "test", path: "test.txt", data: enc.encode("test") })
+		.execute();
+
+	await lix.settled();
+
+	const changes = await lix.db.selectFrom("change").selectAll().execute();
+
+	// console.log(await lix.db.selectFrom("queue").selectAll().execute());
+
+	expect(changes).toEqual([
+		{
+			id: changes[0]?.id,
+			author: "Test User",
+			created_at: changes[0]?.created_at,
+			parent_id: null,
+			type: "text",
+			file_id: "test",
+			plugin_key: "mock-plugin",
+			value: {
+				id: "test",
+				text: "inserted text",
+			},
+			meta: null,
+			commit_id: null,
+			operation: "create",
+		},
+	]);
+
+	const discussion = await lix.startDiscussion({
+		changes: [changes[0]!],
+		body: "comment on a change",
+	});
+
+	expect(discussion.id).toBeTypeOf("string");
+
+	const discussions = await lix.db
+		.selectFrom("discussion")
+		.selectAll()
+		.execute();
+
+	expect(Array.isArray(discussions)).toBe(true);
+
+	expect(discussions).toHaveLength(1);
+
+	expect(discussions[0]).toHaveProperty("id");
+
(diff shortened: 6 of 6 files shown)


lix/packages/sdk/src/discussion/comment.ts
```diff
@@ -0,0 +1,36 @@
+import { Kysely } from "kysely";
+import type { LixDatabaseSchema } from "../database/schema.js";
+import { v4 } from "uuid";
+
+export async function comment(args: {
```
samuelstroschein, line 5: Would rename to `addChangeDiscussionComment` 

- `add` to imply that something is added to the discussion
- `changeDiscussionComment` to avoid ambiguity problems down the road if we introduce other types of comments

lix/packages/sdk/src/discussion/start-discussion.ts
```diff
@@ -0,0 +1,68 @@
+import { Kysely } from "kysely";
+import type { LixDatabaseSchema } from "../database/schema.js";
+import { v4 } from "uuid";
+
+export async function startDiscussion(args: {
```
samuelstroschein, line 5: Minor: `create` sounds more natural than `start`. besides, here too to avoid ambiguity problems down the road, i would call the function `createChangeDiscussion`.

lix/packages/sdk/src/discussion/start-discussion.ts
```diff
@@ -0,0 +1,68 @@
...
+	db: Kysely<LixDatabaseSchema>;
+	currentAuthor: string;
+	changes: { id: string }[];
```
samuelstroschein, line 8: I would use `changeIds` as properties instead of an array of ids. Will make the API more ergonomical because the change object doesn't need to be pruned. 

```ts 
await createChangeDiscussion({ 
  // ...
  changeIds: changes.map((c) => c.id)
})
```

Else, a new object needs to be created for every change.

```ts
await createChangeDiscussion({ 
  // ...
  changes: changes.map((c) => ({ c.id })) 
})

```

lix/packages/sdk/src/discussion/start-discussion.ts
```diff
@@ -0,0 +1,68 @@
+import { Kysely } from "kysely";
+import type { LixDatabaseSchema } from "../database/schema.js";
+import { v4 } from "uuid";
+
+export async function startDiscussion(args: {
+	db: Kysely<LixDatabaseSchema>;
```
samuelstroschein, line 6: If you pass lix as an argument, you can eliminate `currentAuthor` (and other state information that you might need in the future)

```ts
await createChangeDiscussion({
  changeIds,
  lix,
  body
})
```

```ts
export async function createChangeDiscussion(args: { lix, ... }){
  const currentAuthor = await lix.currentAuthor.get()
  // ...
}

```


---

Nice side-effect passing lix is what I sketched in https://github.com/opral/lix-sdk/issues/58#issuecomment-2342285024.

lix/packages/sdk/src/discussion/start-discussion.ts
```diff
@@ -0,0 +1,68 @@
...
+			.insertInto("comment")
+			.values({
+				id: newCommentId,
+				parent_id: undefined,
+				discussion_id: newDiscussionId,
+				author_id: args.currentAuthor,
+				// todo - use zoned datetime
+				body: args.body,
+			})
+			.returning("id")
```
samuelstroschein, line 63: Mistakingly returning id? The comment.id is not used.

lix/packages/sdk/src/discussion/start-discussion.ts
samuelstroschein: EDIT: aha this function is used internally and then exposed via `lix.*` where you fill in the current author. If you agree on modelling lix API tree-shakable, it would prob be good to expose the two functions directly with a lix object. If not, just ship with `lix.*` API and we can revisit the tree-shakable discussion after the feature complete milestone.

samuelstroschein:
From Discord https://discord.com/channels/897438559458430986/1283682858489221161/1283785662373429410

> I didn't touch merging of comments yet since the merging model will change with branching - right?

Branches don't seem to influence comments. Remember the definition of a branch "A branch is a filter for changes that are currently applied". The definition is not a completely isolated state. From a user perspective, any comment on a change seems to be good if it's available in another branch. That avoids conflicts + unnecessary switching of branches. If I am in branch `x` and you are in branch `y`, we can still discuss a change that is shown in both of our branches. 

Merging two lix'es would be copying the comments from one to another (like we do with commits atm). 

**Proposal**

1. Add copying the change discussions in merge from source to target with an onConflict do nothing. Similar to copying the commits over https://github.com/opral/monorepo/blob/bd822e03368873a0e7d5f5af5d057c92944f8361/lix/packages/sdk/src/merge/merge.ts#L106-L114. 

2. Ignore conflicts (two changes have the same parent_id) for now. We can work on discussion conflicts for the release candidate https://linear.app/opral/issue/LIXDK-160/model-conflicts-in-change-discussions

PS it seems like "merging two branches" is different from "merging to lixes". Merging a branch does not merge data because the data of changes, comments, etc. is available in all branches. cc @janfjohannes cautious when implementing branches. it might be a good idea not to touch the `merge` function and create a new `mergeBranch` function instead. Afterward, we can evaluate if finding a better word for merging branches is appropriate (given that no data is merged)

> I added only wrting functions "startDiscussion" and "comment" to lix. -hope this matches your idea of a minimal API 

Yes 👍 

> What would you like to see tested to close this pr?

Add a testing for merging two lixes. The discussions should be copied over. And I would appreciate rename `comment` etc to `changeDiscussionComment`. Then ready to go! 🚀

lix/packages/sdk/src/database/createSchema.ts
```diff
@@ -77,5 +77,39 @@ export async function createSchema(args: { db: Kysely<any> }) {
...
+  ) strict;
+
+  CREATE TABLE discussion_change_map (
+    discussion_id TEXT NOT NULL,
+    change_id TEXT NOT NULL,
+
+    FOREIGN KEY(discussion_id) REFERENCES discussion(id),
+    
+    -- NOTE this will prevent us from dropping changes 
+    FOREIGN KEY(change_id) REFERENCES change(id)
```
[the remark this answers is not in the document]
martin-lysk: Didn't at that for two reasons:
1. we skipped deletion in this ticket 
2. deletion would most likely need to be implemented using a deletion flag to have a tombstone for others to understand the action

lix/packages/sdk/src/discussion/comment.ts
janfjohannes: @samuelstroschein  addChangeDiscussionComment   >>  no developer who is not a java developer would be ok with reading and writing THAT verbose for a part of the core api. i see no reason for going that route instead of addComment or similar simple names

lix/packages/sdk/src/database/createSchema.ts
```diff
@@ -77,5 +77,39 @@ export async function createSchema(args: { db: Kysely<any> }) {
   BEGIN
     insert or replace into file_internal(id, path, data, metadata) values(OLD.file_id, OLD.path, OLD.data, OLD.metadata);
   END;
+
+
+  -- change discussions 
+
+  CREATE TABLE discussion (
```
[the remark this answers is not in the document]
martin-lysk: Discussions don't need to have a change - at least from the current model so i thought we gonna use the discussion for general purpose. But i am fine with change_discussion as well - your call based on this input

lix/packages/sdk/src/database/createSchema.ts
```diff
@@ -77,5 +77,39 @@ export async function createSchema(args: { db: Kysely<any> }) {
...
+
+    FOREIGN KEY(discussion_id) REFERENCES discussion(id),
+    
+    -- NOTE this will prevent us from dropping changes 
+    FOREIGN KEY(change_id) REFERENCES change(id)
+  ) strict;
+
+
+
+  CREATE TABLE comment (
```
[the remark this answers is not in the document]
martin-lysk: depneding on https://github.com/opral/monorepo/pull/3121#discussion_r1757249621 we can call it `change_discussion_comment` or `discussion_comment`. 

I renamed it to just comment before i saw that discussion id can't be null in the schema you proposed. 

I originally had the mental model of a general comment comment where it could reference a comments and/or something else like a discussion.

lix/packages/sdk/src/database/schema.ts
```diff
@@ -9,6 +9,11 @@ export type LixDatabaseSchema = {
 	file_internal: LixFileTable;
 	change_queue: ChangeQueueTable;
 	conflict: ConflictTable;
+
+	// discussion
+	discussion: DiscussionTable;
+	comment: CommentTable;
```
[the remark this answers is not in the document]
martin-lysk: see https://github.com/opral/monorepo/pull/3121#discussion_r1757254954

lix/packages/sdk/src/database/schema.ts
```diff
@@ -133,3 +138,32 @@ type ConflictTable = {
...
+type DiscussionChangeMapTable = {
+	change_id: string;
+	discussion_id: string;
+};
+
+export type Comment = Selectable<CommentTable>;
+export type NewComment = Insertable<CommentTable>;
+export type CommentUpdate = Updateable<CommentTable>;
+type CommentTable = {
+	id: string;
```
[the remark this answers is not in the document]
martin-lysk: Will add that and change the tests correspondingly

lix/packages/sdk/src/discussion/comment.ts
```diff
@@ -0,0 +1,36 @@
...
+	db: Kysely<LixDatabaseSchema>;
+	currentAuthor: string;
+	parentComment: { id: string };
+	body: string;
+}) {
+	const newCommentId = v4();
```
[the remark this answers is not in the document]
martin-lysk: will change test ->  https://github.com/opral/monorepo/pull/3121#discussion_r1757256478

lix/packages/sdk/src/discussion/comment.ts
```diff
@@ -0,0 +1,36 @@
...
+			.executeTakeFirstOrThrow();
+
+		const comment = await trx
+			.insertInto("comment")
+			.values({
+				id: newCommentId,
+				parent_id: args.parentComment.id,
+				discussion_id,
+				author_id: args.currentAuthor,
+				// todo - use zoned datetime
```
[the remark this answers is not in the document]
martin-lysk: left over from coping insert statement - will be gone ;-)

lix/packages/sdk/src/database/createSchema.ts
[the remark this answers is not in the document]
samuelstroschein: > Discussions don't need to have a change 

I see. Too much ambiguity to nail the design for generalized discussions (I would assume a generalized discussion is a file that can be distributed). I feel more comfortable naming it `change_discussion`, eventually knowing that the concept will be deprecated.

lix/packages/sdk/src/discussion/comment.ts
samuelstroschein: @janfjohannes, I expect change discussion comments to evolve, maybe even be succeeded by a generalizable solution. Being verbose now reduces the risk of ambiguity down the road. We can always introduce an alias `addComment` later once we are sure that the concept is nailed.

samuelstroschein: EDIT: I had a mental error, see https://github.com/opral/monorepo/pull/3121#discussion_r1757318548 

AKA yes, just name it discussion and comment. We have a generalizable solution to discuss or comment anything that happens in lix :)

lix/packages/sdk/src/discussion/start-discussion.ts
samuelstroschein: EDIT: due to https://github.com/opral/monorepo/pull/3121#discussion_r1757318548, i would have the following signature 

```ts
await createDiscussion(args: { 
   // ...
  changeIds?: Array<string>
})
```

Making change ids optional means that we can add other entity maps in the future. 

```diff
await createDiscussion(args: { 
   // ...
  changeIds?: Array<string>
+ branchIds?: Array<string>
})
```

martin-lysk: The idea was the other way around - we most likely have an array of changes (as in your example) in the application code already - no need to call map just pass the changes. I could change the type to `Pick<Change, 'id'>` to clarify this.

The function internally maps the ids anyway we could also make the type a mixed type:

Like Pick<Change, 'id'>[] | Change["id"][] - I am ambiguous as this type :D

lix/packages/sdk/src/database/createSchema.ts
```diff
@@ -77,5 +77,39 @@ export async function createSchema(args: { db: Kysely<any> }) {
   BEGIN
     insert or replace into file_internal(id, path, data, metadata) values(OLD.file_id, OLD.path, OLD.data, OLD.metadata);
   END;
+
+
+  -- change discussions 
+
+  CREATE TABLE discussion (
+    -- TODO in inlang i saw we replace uuid_v3 with uuid_v7 any reason we use v4 in lix 
+    id TEXT PRIMARY KEY DEFAULT (uuid_v4())
```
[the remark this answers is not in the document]
martin-lysk: I linked the corresponding ticket to avoid mixing up the concepts in the current state.

martin-lysk:
@samuelstroschein  I Agree, that we can merge the changes without branches and use the same approach of copying over the changes.

> Branches don't seem to influence comments. Remember the definition of a branch "A branch is a filter for changes that are currently applied". The definition is not a completely isolated state. From a user perspective, any comment on a change seems to be good if it's available in another branch. That avoids conflicts + unnecessary switching of branches. If I am in branch `x` and you are in branch `y`, we can still discuss a change that is shown in both of our branches.

Lets say you have two branches `main` and `start-day-updated` and you model company spend on salary:

state in `main`
| Employee        | Yearly Salary | Start Day | Cost 2024 |
|-----------------|---------------|-----------|-----------|
| Max Mustermann | 70k €     | 1. April | 58.3k € |

state in `start-day-updated`
| Employee        | Yearly Salary | Start Day | Cost 2024 |
|-----------------|---------------|-----------|-----------|
| Max Mustermann  | 70k €         | 1. March  | 58.3k €   |

A view on the change table with a join on the branch table would contain something like this:

| change id | parent change | value.id | value.value | branches |
|-----------|---------------|-----------|-----------|-----|
| `1` | null | a0 | Max Musterman | `main`, `start-day-updated` |
| `2` | null | b0 | 70k € | `main`, `start-day-updated` |
| `3` | null | c0 | 1. April | `main`, `start-day-updated` |
| `4` | null | d0 | 58.3k€ | `main`, `start-day-updated` |
| `5` | 3 | c0 | 1. March |  `start-day-updated` |

Now someone points to the fact that the column "Cost 2024" was not updated in`start-day-updated`.
One could start a discussion on different entities:

1. Only on the column "Cost 2024" which contains the error

A discussion with the `discussion_change_map` and `comment` table joined would look like this:
| discussion | changes | comments |
|---|---|---|
| `1` | `4` | "Cost have not been updated after changing the start date!" |

2. The Whole line (pointing to all leaf changes in the line of Max Mustermann"

A discussion with the `discussion_change_map` and `comment` table joined would look like this:

| discussion | changes | comments |
|---|---|---|
| `1` | `1`,`2`,`4`,`5` | "Cost have not been updated after changing the start date!" |

3. Only on the column that is wrong "Cost 2024" and the one that produced the error "Start Day"

A discussion with the `discussion_change_map` and `comment` table joined would look like this:
| discussion | changes | comments |
|---|---|---|
| `1` | `4` `5` | "Cost have not been updated after changing the start date!" |

We would need to have a reference to branches because this comment makes no sense for `main`. While one could argue that we could filter comments for a branch that only reference changes available in the current branch - case 1. would be visible anyway.

cc @janfjohannes

lix/packages/sdk/src/discussion/start-discussion.ts
samuelstroschein: Function overloads are hard to discover. I see where you are coming from, but the assumption is that changes have been queried. 

```ts
const changes = await lix.db.selectFrom("change")

await createDiscussion({ changes })
```

The ugly mapping comes into play when a table that only exposes the change_id is queried. 

```ts
const other = await lix.db.selectFrom("otherTable")

await createDiscussion({ changes: other.map((o) => ({ id: o.change_id}) })
```

I would go with `change_id` :)

samuelstroschein:
sync call with @martin-lysk and @janfjohannes : 

- yes, discussions might be tied to a branch (sequence id) 
- we should be able to merge this pr to unblock apps and later implement tying to branches
- conflicts don't need to be modeled because we use the `parent_id` and `created_at` information for sorting and in the UI we would show a "this comment is a reply to"

lix/packages/sdk/src/discussion/start-discussion.ts
martin-lysk: > Function overloads are hard to discover. 

so we won't use it here?!

> I would go with change_id :)

Not 100 % which one you preferred now:

```
await createDiscussion(args: { 
   // ...
  changeIds?: Array<string>
})
```

or 

```
await createDiscussion(args: { 
   // ...
  changes?: {change_id: string}[]
})
```

samuelstroschein: await createDiscussion(args: { 
   // ...
  changeIds?: Array<string>
})

this one

samuelstroschein:
@martin-lysk github notified me that this pr is ready for review but i don't see an explicit prompt from you. lmk if i should review this pr

martin-lysk:
yes @samuelstroschein , i worked in the signature part. Plz review

lix/packages/sdk/src/discussion/comment.ts
```diff
@@ -0,0 +1,32 @@
...
+import type { LixDatabaseSchema } from "../database/schema.js";
+import { v4 } from "uuid";
+
+export async function comment(args: {
+	db: Kysely<LixDatabaseSchema>;
+	currentAuthor: string;
+	parentComment: { id: string };
+	body: string;
+}) {
+	return args.db.transaction().execute(async (trx) => {
```
samuelstroschein, line 11: Found myself wondering what comment is/does. Using `addComment` would clarify that a comment is added to a discussion.

```suggestion
export async function addComment(args: {
	db: Kysely<LixDatabaseSchema>;
	currentAuthor: string;
	parentComment: { id: string };
	body: string;
}) {
	return args.db.transaction().execute(async (trx) => {
```

lix/packages/sdk/src/discussion/create-discussion.ts
```diff
@@ -0,0 +1,67 @@
...
+				.selectFrom("change")
+				.select(trx.fn.countAll().as("count"))
+				// NOTE: for high amount of changes this will be extremly costy since its a huge or clause
+				.where("change.id", "in", changeIds) // NOTE: this will need to be batched if the amount of changes passed is bigger than sqlite max parameter count
+				.executeTakeFirstOrThrow()) as { count: number }; // TODO check if this cast is ok - it coult be a string/bigint or number
+
+			if (count !== changeIds.length) {
+				throw new Error("unknown changes passed");
+			}
+		}
```
samuelstroschein, line 30: If we activate foreign keys, SQLite takes care of handling if the change exists. Do we need to verify if the changes exists?

lix/packages/sdk/src/discussion/create-discussion.ts
```diff
@@ -0,0 +1,67 @@
...
+import { v4 } from "uuid";
+
+export async function createDiscussion(args: {
+	db: Kysely<LixDatabaseSchema>;
+	currentAuthor: string;
+	changeIds?: string[];
+	body: string;
+}) {
+	const newDiscussionId = v4();
+	const newCommentId = v4();
```
samuelstroschein, line 12: Not needed. Discussions and comments have default ids.

lix/packages/sdk/src/discussion/create-discussion.ts
```diff
@@ -0,0 +1,67 @@
...
+			if (count !== changeIds.length) {
+				throw new Error("unknown changes passed");
+			}
+		}
+
+		const discussion = await trx
+			.insertInto("discussion")
+			.values({
+				id: newDiscussionId,
+			})
```
samuelstroschein, line 36: You can insert with `defaultValues()` and then use the reutrn type of discussion to get the inserted id

lix/packages/sdk/src/discussion/create-discussion.ts
```diff
@@ -0,0 +1,67 @@
...
+		// create mapping from discussion to changes
+		if (changeIds) {
+			for (const changeId of changeIds ?? []) {
+				await trx
+					.insertInto("discussion_change_map")
+					.values({
+						change_id: changeId,
+						discussion_id: newDiscussionId,
+					})
+					.executeTakeFirstOrThrow();
```
samuelstroschein, line 49: You can remove the return of values.

```suggestion
					.execute();
```

lix/packages/sdk/src/discussion/create-discussion.ts
```diff
@@ -0,0 +1,67 @@
...
+		}
+
+		await trx
+			.insertInto("comment")
+			.values({
+				id: newCommentId,
+				parent_id: undefined,
+				discussion_id: newDiscussionId,
+				author_id: args.currentAuthor,
+				// todo - use zoned datetime
```
samuelstroschein, line 60: remove outdated todo 

```suggestion
```

lix/packages/sdk/src/discussion/create-discussion.ts
```diff
@@ -0,0 +1,67 @@
...
+				body: args.body,
+			})
+			.executeTakeFirstOrThrow();
```
samuelstroschein, line 63: The return value is not used so `.execute()` is sufficent.

```suggestion
			.execute();
```

lix/packages/sdk/src/discussion/discussion.test.ts
```diff
@@ -0,0 +1,174 @@
...
+				text: "inserted text",
+			},
+			meta: null,
+			commit_id: null,
+			operation: "create",
+		},
+	]);
+
+	const discussion = await lix.createDiscussion({
+		changes: [changes[0]!],
```
samuelstroschein, line 82:
```suggestion
		changeIds: changes.map(c => c.id),
```

samuelstroschein: outdated test i assume

lix/packages/sdk/src/open/openLix.ts
```diff
@@ -171,6 +173,22 @@ export async function openLix(args: {
...
+			if (currentAuthor === undefined) {
+				throw new Error("current author not set");
+			}
+			return createDiscussion({
+				...args,
+				db,
+				currentAuthor,
+			});
+		},
+		comment: (args: { parentComment: { id: string }; body: string }) => {
```
samuelstroschein, line 186: I would use id directly here too to avoid passing around objects that might or might not be convenient to create.

```suggestion
		comment: (args: { parentCommentId: string; body: string }) => {
```

samuelstroschein (changes requested):
Yep, that's what i had in mind. requesting two changes before approval (rest of the comments is minor and can be ignored or implemented): 

- [ ] change `comment` to `addComment`
- [ ] change `lix.comment(args: { parentChange: { id: string }})` to `lix.comment(args: { parentChangeId: string})`

lix/packages/sdk/src/discussion/create-discussion.ts
samuelstroschein: Probobaly not. So, I vote to remove this costly check.

martin-lysk:
Worked in you feedback - feel free to merge if it looks good to you @samuelstroschein

PR #3122, opened 2024-09-12
dont override draft changes

janfjohannes:


lix/packages/sdk/src/change-queue.test.ts
```diff
@@ -122,14 +162,14 @@ test("should use queue and settled correctly", async () => {
 	const queue2 = await lix.db.selectFrom("change_queue").selectAll().execute();
 	expect(queue2).toEqual([
 		{
```
samuelstroschein, line 164: No action needed. I found `expect.objectContaining` to be helpful to avoid updating tests if the data model changes. In my PR, I opted to test the change ids and values while ignoring unrelated properties like type etc. Might be something for future tests.

janfjohannes: @samuelstroschein  depends if the data model is exposed. as this can be accessed from apps using the sdk i figured it is not a downside to reference the model in the expects, but will keep objectContaining in mind.

PR #3124, opened 2024-09-16
branch support

janfjohannes:


lix/packages/sdk/src/branch/get-leaf-changes-diff.ts
```diff
@@ -0,0 +1,113 @@
...
+	const targetChangesByValueId = leafChangesInTarget.reduce((agg, change) => {
+		// @ts-ignore
+		agg[change.value.id] = change;
+		return agg;
+	}, {});
+
+	const toDelete = [];
+	for (const change of leafChangesInTarget) {
+		// @ts-ignore
+		if (!sourceChangesByValueId[change.value.id]) {
```
martin-lysk, line 82: Wouldn't this delete an entry that was created in the target and does not exist in the source - while not really been deleted in source?

lix/packages/sdk/src/branch/get-leaf-changes-diff.ts
```diff
@@ -0,0 +1,113 @@
...
+		// @ts-ignore
+		const targetChange = targetChangesByValueId[change.value.id];
+		if (change.id === targetChange.id) {
+			continue;
+		}
+
+		// @ts-ignore
+		const old = targetChangesByValueId[change.value!.id] as Change | undefined;
+
+		const diff = await plugin.diff[change.type]!({
```
martin-lysk, line 102: I would check for the differ of the given type to exist and throw with an informative error like
```
 `diff function for change type ${change.type} not found in lix plugin`
```

Failing is way better than the silent skip - a problem i faced here so:
https://github.com/opral/monorepo/blob/lix-integration/lix/packages/sdk/src/file-handlers.ts#L143

lix/packages/sdk/src/branch/get-leaf-changes-diff.ts
```diff
@@ -0,0 +1,113 @@
...
+			toDelete.push(change);
+		}
+	}
+
+	const plugin = args.plugins[0]!;
+	const toChange = [];
+
+	for (const change of leafChangesInSource) {
+		// @ts-ignore
+		const targetChange = targetChangesByValueId[change.value.id];
```
martin-lysk, line 94: How does this behave if the value for the given leaf coming from a change in the source does not exist in the target? (Entity was created in source)

lix/packages/sdk/src/branch/switch.ts
```diff
@@ -0,0 +1,88 @@
...
+			leafChangesDiff.reduce((agg, change) => {
+				// @ts-ignore
+				if (!agg[change.file_id]) {
+					// @ts-ignore
+					agg[change.file_id] = [];
+				}
+				// @ts-ignore
+				agg[change.file_id].push(change);
+				return agg;
+			}, {});
```
martin-lysk, line 63:
```suggestion
			leafChangesDiff.reduce<Record<string, Change[]>>((agg, change) => {
				if (!agg[change.file_id]) {
					agg[change.file_id] = [];
				}
				agg[change.file_id]!.push(change);
				return agg;
			}, {});
```

lix/packages/sdk/src/branch/switch.ts
```diff
@@ -0,0 +1,88 @@
...
+			const file = await trx
+				.selectFrom("file")
+				.selectAll()
+				.where("id", "=", fileId)
+				.executeTakeFirstOrThrow();
+
+			const { fileData } = await plugin!.applyChanges!({
+				lix: { db: args.db, plugins: args.plugins },
+				file,
+				// @ts-ignore
```
martin-lysk, line 75: hard to understand where _delete comes from - I suggeste we change the signature of getLeafChangesDiff to return `{ toDelete: Change[], toChange: Change[] }`

This would make the ts-ignore obsolete as well

martin-lysk (commented):
I focused on the new branch switch function first - will continue reviewing

lix/packages/sdk/src/database/schema.ts
```diff
@@ -4,8 +4,9 @@ import type { LixPlugin } from "../plugin.js";
 export type LixDatabaseSchema = {
 	file: LixFileTable;
 	change: ChangeTable;
-	commit: CommitTable;
-	ref: RefTable;
+	change_view: ChangeView;
```
martin-lysk, line 7: if we remove the commit function - fink and so on will fail with this branch - shall we make the required changes in fink, csv app etc part of this brach?

lix/packages/sdk/src/file-handlers.ts
```diff
@@ -141,30 +160,48 @@ export async function handleFileChange(args: {
...
 				if (previousChange) {
 					previousCommittedDiff = await pluginDiffFunction?.[diff.type]?.({
-						old: previousChange?.value,
+						old: previousChange.value,
 						neu: diff.neu,
 					});
 
-					if (previousCommittedDiff?.length === 0) {
-						// drop the change because it's identical
+					if (previousCommittedDiff.length === 0) {
```
martin-lysk, line 167: Intersting that tsc doesn't complain about this...

we asign it to
`await pluginDiffFunction?.[diff.type]?.({
						old: previousChange.value,
						neu: diff.neu,
					});`
which can be undefined if pluginDiffFunction is undefined or it ha no proeperty named like the changes type. This code has changed for that reason in lix-integration

martin-lysk (commented):
I focused on the new branch switch function first - will continue reviewing

lix/packages/sdk/src/branch/errors.ts
```diff
@@ -0,0 +1,18 @@
...
+	constructor(message: string) {
+		super(message);
+		this.name = "BranchError";
+	}
+}
+
+export class BranchExistsError extends BranchError {
+	constructor(args: { name: string }) {
+		super(dedent`
+The branch with id "${args.name}" already exists.
```
samuelstroschein, line 13:
```suggestion
	constructor(args: { id: string }) {
		super(dedent`
The branch with id "${args.id}" already exists.
```

lix/packages/sdk/src/merge/merge.test.ts
```diff
@@ -3,10 +3,11 @@ import { test, expect, vi } from "vitest";
 import { openLixInMemory } from "../open/openLixInMemory.js";
 import { newLixFile } from "../newLix.js";
 import { merge } from "./merge.js";
-import type { NewChange, NewCommit, NewConflict } from "../database/schema.js";
+import type { NewChange, NewConflict } from "../database/schema.js";
 import type { LixPlugin } from "../plugin.js";
 
 test("it should copy changes from the sourceLix into the targetLix that do not exist in targetLix yet", async () => {
+	debugger;
```
samuelstroschein, line 10:
```suggestion
```

lix/packages/sdk/src/open/openLix.ts
```diff
@@ -189,7 +187,25 @@ export async function openLix(args: {
 			}
 			return addComment({ ...args, db, currentAuthor });
 		},
+		createBranch: (
+			name: string,
+			{ branchId }: { branchId: string | undefined } = { branchId: undefined },
+		) =>
```
samuelstroschein, line 193: I recommend to always use objects as arguments to decrease the risk of breaking changes. 

- ordering of arguments is irrelevant (imagine that `name` becomes optional, then the API is screwed with `createBranch(undefined, { branchId })`)
- adding and removing arguments is possible

```suggestion
		createBranch: (args: { id?: string, name: string}) =>
```

lix/packages/sdk/src/open/openLix.ts
```diff
@@ -189,7 +187,25 @@ export async function openLix(args: {
...
+			createBranch({
+				db,
+				name: name,
+				branchId,
+			}),
+		switchBranch: ({ branchId, name }: { branchId?: string; name?: string }) =>
```
samuelstroschein, line 199: It's reasonable to expect that an app knows the id of a branch. 

- simplifies the API
- fixes a bug with the type (brnachId and name can't be both undefined)

```suggestion
		switchBranch: (args: { id: string }) =>
```

samuelstroschein (commented):
Only commented with easy to fix API signature changes. 

- ignoring logic as we agreed to just ship branches and then see how apps need it/what needs to change

lix/packages/sdk/src/branch/errors.ts
samuelstroschein: The `name` has been used as `id` here. That's a bug. If this is a user facing error, adding both id and name to the error makes sense. 

- apps can process the branch id 
- users see an error message for the branch name

```
constructor(args: { id: string, name: string }){
  this.id = args.id
  this.name = args.name
  this.message = `The branch "${args.name}" already exists.`
}
```

lix/packages/sdk/src/branch/get-leaf-changes-diff.ts
lucidNTR: @martin-lysk  yes that is the goal here. remember this does not delete the change entry just the effect it has on the file table which is exactly what switching from source to target branch means. ping me if this does not make sense and we need to discuss...

lix/packages/sdk/src/branch/get-leaf-changes-diff.ts
lucidNTR: in that case old would be undefined in the apply changes function and it would look like a new change which is exactly what this is.

lix/packages/sdk/src/database/schema.ts
lucidNTR: as discussed, yes we do!

lix/packages/sdk/src/query-utilities/get-lowest-common-ancestor.ts
```diff
@@ -16,38 +16,95 @@ export async function getLowestCommonAncestor(args: {
...
+			.where("branch_id", "=", currentSourceBranch.id)
+			.orderBy("seq", "desc")
+			.where(
+				"change.id",
+				"=",
+				nextChange?.parent_id ?? args.sourceChange.parent_id,
+			)
+			.executeTakeFirst()) as Change;
 
-		if (!nextChange || !nextChange.parent_id) {
```
janfjohannes, line 38: @martin-lysk  not sure why you added !nextChange.parent_id  what happens if the root change is the common ancestor?

PR #3125, opened 2024-09-16
fix datastructure and simplify mental model

samuelstroschein:
Part of https://linear.app/opral/issue/MESDK-238/model-import-crud-sdk-export. Watch [this loom](https://www.loom.com/share/e4b34c07e8064838b2e170cc246f8479?sid=c41f0608-7a4c-4768-84e0-66c8ce3c778d) and this [thread](https://linear.app/opral/issue/MESDK-238/model-import-crud-sdk-export#comment-50b0292c) for context. 

The mental model seems simple now "A message declares input and local variables. (in the future, a message can declare even more things)". 

### Fixes

- inputs can't have expressions because they are static 
- Local variables have a different type than variable references.
- (simplifies compiling) local variables are not defined via expressions 

### Changes

1. Extracts the `input` type (instead of being a declaration)
2. Removes `input.value` because not needed atm.
3. (Renames function annotation to function reference because a) it's a reference and b) aligns with variable reference)

### Additional context

- The tests in the sdk2 are passing.


inlang/source-code/sdk2/src/mock/bundle.ts
```diff
@@ -17,35 +17,14 @@ export const pluralBundle: BundleNested = {
 				{
 					type: "input",
 					name: "numProducts",
-					value: {
-						type: "expression",
-						arg: {
-							type: "variable",
-							name: "numProducts",
-						},
-					},
```
samuelstroschein, line 26: The removed code is an example of the confusion that this PR addresses. An input with name `numProducts` is declaring a variable. The author thought that in addition of an input `numProducts`, the expression of declaring a new `variable` `numProduct` is needed. 

- an input declares a variable, the expression doesn't do anyhting
- the expression references a variable, `type: variable` is not a declaration!

inlang/source-code/sdk2/src/json-schema/pattern.ts
```diff
@@ -38,12 +38,17 @@ export const Text = Type.Object({
 	value: Type.String(),
 });
 
-export type Declaration = Static<typeof Declaration>;
-export const Declaration = Type.Object({
+export type Input = Static<typeof Input>;
+export const Input = Type.Object({
 	type: Type.Literal("input"),
 	name: Type.String(),
-	value: Expression,
```
samuelstroschein, line 45: Removing the possibility for an input to have an expression is the only change at this point.

PR #3126, opened 2024-09-16
follow up rename variable to reference

samuelstroschein:
follow up of #3125 which is a nice to have and could be merged but it's not crucial: 

- change `type: variable` to `type:variable-reference` to align with naming and avoid confusions like https://github.com/opral/monorepo/pull/3125/files#r1761844043 
- change `type: function` to `type: function-refernece` to align with naming of the TS type and avoid confusions


samuelstroschein:
Implemented in #3138

PR #3127, opened 2024-09-16
follow up: explicit variable types

samuelstroschein:
Follow up of #3125 that can be accepted or rejected independently. This is a nice feature that will lead to minimal breaking code in dependents of the sdk2. 

**Changes**

- explicit `input variable` and `local variable` to align the concept of "a message can reference (local|input|global) variables"


samuelstroschein:
We could go with a non-breaking PR where the TS type is called `InputVariable` but the type: input stays `type: input`

samuelstroschein:
implemented in #3126

PR #3137, opened 2024-09-19
refactor: remove `bundle.alias`

samuelstroschein:
Closes https://github.com/opral/inlang-sdk/issues/200. 

#### `@inlang/sdk2` 

- removed bundle.alias everywhere
- removed stableHumanId because not needed anymore
- deprecated `bundleIdOrAliasIs`. queries go against the id of a bundle cc @felixhaeberle 
- removed moormash dependency because stableHumanId is not needed anymore

#### `@inlang/fink2`

- removed bundle aliases from demo project creation 
- manual testing worked as expected

#### `@inlang/bundle-component`

- removed aliases from being displayed

#### `vs-code-extension`

- removed aliases from being displayed
- `getBundleByIdOrAlias` only searches by id but still exists to decrease refactoring
- fixed machine translate mocking 
- tests pass but vitest is flaky (see video) cc @martin-lysk wtf is this. i cleared the cache.

https://github.com/user-attachments/assets/7a23c085-f258-4c0d-8fc0-be0c4c9ca376


martin-lysk:
So we see the same behaviour as in the extension in cli now as well :-/ doesn't sound like a coincident... We got to takle this!

PR #3144, opened 2024-09-24
flip the rows in the History view and the arrow

jannesblobel:
**New** 
CleanShot 2024-09-24 at 15 14 09@2x

**Old** 
CleanShot 2024-09-24 at 15 16 38@2x


jannesblobel:
The changed/new row is now above the old row. This helps the reading flow and is a better UX 
Discussed with Nils at a meeting last Wednesday

PR #3152, opened 2024-09-27
Merge CSV2 Branch into lix-integration 

martin-lysk:
The branch contains various fixes needed in lix-integration:

lix/packages/sdk/src/merge/merge.ts:
- merge function accepts merges of lix managing multiple files
- merge function copies the files from source into target - if it finds changes in the source and the file does not exist in target (indicates the file is on change track in lix)
lix/packages/sdk/src/open/openLix.ts
- close function ensures the change queue is settled and stoped - to avoid post lix close db access exceptions and prevent set timeout to leak memory
lix/packages/sdk/src/query-utilities/get-lowest-common-ancestor.ts
- fixes a bug that lead to inifity loops: changes parent check in getLowestCommonAncestor to a falsy check. parent_id is returned as null and not as undefiend -


martin-lysk:
I merge this as preparation of the branching merge

PR #3161, opened 2024-10-11
chore: make typescript a dev dependency

benmccann:
also sorted them while I was at it

ref https://github.com/opral/inlang-paraglide-js/issues/241


benmccann:
closing as this will be deleted

PR #3164, opened 2024-10-14
Lix benchmarking

martin-lysk:
DO NOT MERGE! 
This branch is not meant to be merged, it contains some benchmarks on the tables that gave us insides on the table structure of lix like https://linear.app/opral/issue/LIXDK-167/add-snapshot-table.

The benchmarks will most likely become outdated with the changes in the table structures. 

This branch contains valuable blueprints for future benchmarking initiatives and query optimization like recursive query - so i suggest to keep it open until lix api is stable.

### DIFF
--- a/lix/packages/sdk/package.json
+++ b/lix/packages/sdk/package.json
@@ -8,6 +8,7 @@
 	},
 	"scripts": {
 		"build": "tsc --build",
+		"bench": "vitest bench",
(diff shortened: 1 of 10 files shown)


samuelstroschein:
close because stale?

samuelstroschein:
- [ ]

PR #3166, opened 2024-10-14
Lixdk 111 old neu alternative

martin-lysk:
compare https://linear.app/opral/issue/LIXDK-111/old-neu-alternative


samuelstroschein (approved):
Let's go

PR #3170, opened 2024-10-15
refactor: add entity id

samuelstroschein:
Closes https://github.com/opral/lix-sdk/issues/94

**Changes**

- adds `entity_id` to a change
- removes the `& { id: string}` requirement from snapshots
- cleans tests

### DIFF
--- a/lix/packages/sdk/src/change-queue.test.ts
+++ b/lix/packages/sdk/src/change-queue.test.ts
@@ -4,64 +4,28 @@ import { newLixFile } from "./newLix.js";
 import type { DiffReport, LixPlugin } from "./plugin.js";
 
 test("should use queue and settled correctly", async () => {
-	const mockPlugin: LixPlugin<{
-		text: { id: string; text: string };
-	}> = {
+	const mockPlugin: LixPlugin = {
 		key: "mock-plugin",
 		glob: "*",
 		diff: {
 			file: async ({ before, after }) => {
-				const dec = new TextDecoder();
-				// console.log("diff", after, before?.data, after?.data);
-				const textBefore = dec.decode(after?.data);
-				const textAfter = dec.decode(before?.data);
+				const textBefore = before
+					? new TextDecoder().decode(before?.data)
+					: undefined;
+				const textAfter = after
+					? new TextDecoder().decode(after.data)
+					: undefined;
 
 				if (textBefore === textAfter) {
 					return [];
 				}
-
-				return await mockPlugin.diff.text({
-					before: before
-						? {
-								id: "test",
-								text: textAfter,
-							}
-						: undefined,
-					after: after
-						? {
-								id: "test",
-								text: textBefore,
-							}
-						: undefined,
-				});
-			},
-			text: async ({ before, after }) => {
-				// console.log("text", before, after);
-				if (before?.text === after?.text) {
-					return [];
-				}
-
 				return [
-					!before
-						? {
-								type: "text",
-								before: undefined,
-								after: {
-									id: "test",
-									text: after?.text,
-								},
-							}
-						: {
-								type: "text",
-								before: {
-									id: "test",
-									text: before?.text,
-								},
-								after: {
-									id: "test",
-									text: after?.text,
-								},
-							},
+					{
+						type: "text",
+						entity_id: "test",
+						before: textBefore ? { text: textBefore } : undefined,
+						after: textAfter ? { text: textAfter } : undefined,
+					},
 				];
 			},
 		},
@@ -118,7 +82,20 @@ test("should use queue and settled correctly", async () => {
 	const changes = await lix.db
 		.selectFrom("change")
 		.innerJoin("snapshot", "snapshot.id", "change.snapshot_id")
-		.selectAll()
+		.select([
+			"change.id",
+			"change.author",
+			"change.created_at",
+			"change.snapshot_id",
+			"change.parent_id",
+			"change.entity_id",
+			"change.type",
+			"change.file_id",
+			"change.plugin_key",
+			"snapshot.value as value",
+			"change.meta",
+			"change.commit_id",
+		])
 		.execute();
 
 	expect(changes).toEqual([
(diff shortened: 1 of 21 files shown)


lix/packages/sdk/src/change-queue.test.ts
```diff
@@ -118,7 +82,20 @@ test("should use queue and settled correctly", async () => {
...
+			"change.snapshot_id",
+			"change.parent_id",
+			"change.entity_id",
+			"change.type",
+			"change.file_id",
+			"change.plugin_key",
+			"snapshot.value as value",
+			"change.meta",
+			"change.commit_id",
+		])
```
samuelstroschein, line 98: This is required because an 		.innerJoin("snapshot", "snapshot.id", "change.snapshot_id") select all overshadows the change.id with the snapshot.id

lix/packages/sdk/src/discussion/discussion.test.ts
```diff
@@ -136,33 +114,6 @@ test("should fail to create a disussion on non existing changes", async () => {
...
-			plugin_key: "mock-plugin",
-			value: {
-				id: "test",
-				text: "inserted text",
-			},
-			meta: null,
-			commit_id: null,
-		},
-	]);
-
```
samuelstroschein, line 165: The discussion test tested the change queue :D redundant

lix/packages/sdk/src/file-handlers.ts
```diff
@@ -142,47 +139,42 @@ export async function handleFileChange(args: {
...
-
 				// working change exists but is identical to previously committed change
 				if (previousChange) {
-					previousCommittedDiff = await pluginDiffFunction?.[diff.type]?.({
-						before: previousChange?.value,
-						after: diff.after,
-					});
-
-					if (previousCommittedDiff?.length === 0) {
+					const previousSnapshot = await trx
```
samuelstroschein, line 144: Very nice: The snapshot can now be queried on demand.

lix/packages/sdk/src/change-queue.test.ts
martin-lysk: Oh god what a beast - this combined with flaky run envs in test (importing from dist...) Sounds like fun.

martin-lysk: I looked into the code - seems there are a couple more places where this is still a problem

martin-lysk: I wonder if we can use some bits in the uuid v7 to encode the entity type (like id of a change vs id of a snapshot) - In a former project we used some of the machine id bits in [snowflake](https://en.wikipedia.org/wiki/Snowflake_ID) to signal the type of the id. This allowed us to runtime assert correct entities early based on passed id's - saved us hours, if not days of debugging and post mortems. 

**Not the solution** - and I need into uuid v7 but to outline what I have in mind:

If every id of a change must end with a C
40e97286-9b78-4a35-8fb1-0036bdfbbf7C
Every id of a snapshot ends with an S
40e97286-9b78-4a35-8fb1-0036bdfbbf7S

We could assert in a function like getLeafChange that the id is an expected one:
```
getLeafChange(change) {
  if (!change.id.endsWith('C')) {
    throw new WrongObjectTypePassed("The Id passed is not the id of a change")
  } 
 ....
}

```

martin-lysk: I double checked kysely we can make use selectAll with 'change' as parameter and and an explicit select
```suggestion
+		.selectAll('change')
+		.select('snapshot.value')
```
which is translated to:

```
SELECT
  "change".*,
  "snapshot"."value"
```

I will commit on this pr and fix the other places where we make use of innerjoin as well

lix/packages/sdk/src/file-handlers.ts
```diff
@@ -90,28 +87,30 @@ export async function handleFileChange(args: {
...
+	}[] = [];
 
 	for (const plugin of args.plugins) {
 		// glob expressions are expressed relative without leading / but path has leading /
 		if (!minimatch(normalizePath(args.after.path), "/" + plugin.glob)) {
 			break;
 		}
 
-		const diffs = await plugin.diff!.file!({
+		const diffs = await plugin.diff?.file?.({
```
martin-lysk, line 101: see previous comment about failling instead of optional chaining

lix/packages/sdk/src/file-handlers.ts
```diff
@@ -142,47 +139,42 @@ export async function handleFileChange(args: {
...
 					.values({
-						id: v4(),
 						type: diff.type,
 						file_id: fileId,
 						plugin_key: pluginKey,
+						entity_id: diff.entity_id,
 						author: args.currentAuthor,
 						parent_id: previousChange?.id,
-						snapshot_id: snapshotId,
+						snapshot_id: snapshot.id,
```
martin-lysk, line 177: way better/cleaner than the (await ...).id

lix/packages/sdk/src/change-queue.test.ts
samuelstroschein: Oh wow I was looking for `select('change.*')`. I didn't think about using `selectAll('change')`. Nice.

lix/packages/sdk/src/change-queue.test.ts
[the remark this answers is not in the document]
samuelstroschein: I propose to remove it in another PR. One step at a time

lix/packages/sdk/src/file-handlers.ts
```diff
@@ -20,50 +19,48 @@ export async function handleFileInsert(args: {
...
+	}[] = [];
 	// console.log({ args });
 	for (const plugin of args.plugins) {
 		// glob expressions are expressed relative without leading / but path has leading /
 		if (!minimatch(normalizePath(args.after.path), "/" + plugin.glob)) {
 			break;
 		}
 
-		const diffs = await plugin.diff!.file!({
+		const diffs = await plugin.diff?.file?.({
```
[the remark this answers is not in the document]
samuelstroschein: Will impement that in https://linear.app/opral/issue/LIXDK-107/plugindifffileortype-plugindetectchanges

lix/packages/sdk/src/file-handlers.ts
samuelstroschein: Will impement this in https://linear.app/opral/issue/LIXDK-107/plugindifffileortype-plugindetectchanges

lix/packages/sdk/src/change-queue.test.ts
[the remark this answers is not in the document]
martin-lysk: was just about that one line  in the test (as you where cleaning up the discussion tests as well - but not important)

PR #3173, opened 2024-10-16
add eslint

samuelstroschein:
Adds eslint to the lix sdk. 

- recommended rules for JS and TS
- restricted import from dist (fixes https://github.com/opral/lix-sdk/issues/96)
- fixes various lint reports


samuelstroschein:
cc @martin-lysk merging without review because doesn't touch source code. don't be surprised if you see lint warnings now though :D

PR #3184, opened 2024-10-22
paraglide-sveltekit: Fix constants

Venmit:


samuelstroschein:
Big thank you @Venmit !

samuelstroschein:
Published as `0.11.5` 

CleanShot 2024-10-22 at 18 48 05@2x

PR #3199, opened 2024-11-03
Allow for next.config.ts in paraglide cli init command

gerardmarquinarubio:
Really silly PR but I think its important taking into account that NextJS 15 init will now by default create `next.config.ts` files.

`inlang/source-code/paraglide/paraglide-next/src/cli/flows/addPluginToNextConfig.ts` doesn't need to be modified as it relies purely on regex to modify the NextJs config.

Couldn't test as I can't seem to get a build going on my Ubuntu machine.

Apologies in advance for possibly missing any procedures in my contribution but I couldn't find any guides or references to do so for this repo.

### DIFF
--- a/.changeset/curly-ladybugs-happen.md
+++ b/.changeset/curly-ladybugs-happen.md
@@ -1,0 +1,5 @@
+---
+"@inlang/paraglide-next": patch
+---
+
+allow for next.config.ts in paraglide cli init command https://github.com/opral/monorepo/pull/3199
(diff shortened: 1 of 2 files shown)


gerardmarquinarubio:

gerardmarquinarubio:
I could run the build commands succesfully after downgrading to node 22 https://github.com/evanw/esbuild/issues/3951 .

samuelstroschein:
mercy for the PR. 

i suspect that nothing will be affected, given that the only change is adding a matching string the the array and CI runs through. will be published as `0.7.3`

PR #3206, opened 2024-11-11
Lix csv plugin row header entities

martin-lysk:
Resolves https://linear.app/opral/issue/LIXDK-223/add-row-and-header-row-entities-to-csv-plugin


samuelstroschein:
Merged in #3221

PR #3224, opened 2024-11-19
introduce support for rollup virtual modules used by other vite-plugins

TazorDE:
The paraglide vite plugin resolves all files during the build process. If any filename contains a NULL character (like some vite/rollup virtual modules do) the paraglide-process doesn't check if the paths contains NULL characters.

> If your plugin uses 'virtual modules' (e.g. for helper functions), prefix the module ID with \0. This prevents other plugins from trying to process it. (https://rollupjs.org/plugin-development/#conventions)

Paraglide doesn't use virtual modules, but if other plugins use them paraglide inevitably will touch them. This PR will instruct the paraglide vite plugin to ignore any file, if the filename contains a NULL character.

resolves https://github.com/opral/inlang-paraglide-js/issues/192

### DIFF
--- a/.changeset/swift-eels-crash.md
+++ b/.changeset/swift-eels-crash.md
@@ -1,0 +1,5 @@
+---
+"@inlang/paraglide-unplugin": patch
+---
+
+Support for NULL character prefixed virtual modules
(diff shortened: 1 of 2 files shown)


TazorDE:

TazorDE:
recheck

PR #3247, opened 2024-12-03
Make paraglide files available via Virtual module

LorisSigrist:
Fixes opral/inlang-paraglide-js#264

This PR adds a `useVirtualModule` option to `paraglide-unplugin` (and therefore all it's descendants). This makes a `$paraglide` virtual module available instead of writing to disk. You can import it like so:

```js
import { languageTag } from "$paraglide/runtime.js"
import * as m from "$paraglide/messages.js"
```

Types are provided via a `.d.ts` file with `declare module` declarations.

**Where should the `.d.ts` file be placed?**
In the `outdir`.

## Todos
- [x] Changesets
- [x] Always use virtual module during build, even if a version is written to disk (the virtual module has better code-splitting). 
- [x] Pass CI


samuelstroschein:
@manuel3108 @benmccann, where should the d.ts files go? 

Using the `outdir` seems reasonable, or hardcoded into `/src/paraglide.d.ts` and throw when `outdir` is specified with `useVirtualModule` (because there no outdir 'really').

benmccann:
My suggestion would be to make it so that `paraglide.d.ts` does not have dynamically generated content and then the whole problem goes away. Instead of having `availableLanguageTags` as a constant defining which languages are available, you could have something such as a method `getAvailableLanguageTags(): Set<string>`. Then that could be shipped in the paraglide type definitions

inlang/source-code/paraglide/paraglide-js/example/tsconfig.json
```diff
@@ -8,6 +8,7 @@
 		"allowJs": true,
 		"checkJs": true,
 		"outDir": "dist",
+		"skipLibCheck": true,
```
manuel3108, line 11: Is this a leftover or would such a configuration be necessary for all descendants like `sv` to also setup new projects like this?

manuel3108:
> My suggestion would be to make it so that `paraglide.d.ts` does not have dynamically generated content and then the whole problem goes away.

This would be the ideal solution.

I'm wondering why this would even be necessary, as the current integration seems to able to just work without such file. At least I was unable to find any `.d.ts` file inside the `outDir`. Or is this now necessary because of the virtual modules?

samuelstroschein:
> @benmccann  My suggestion would be to make it so that paraglide.d.ts does not have dynamically generated content and then the whole problem goes away.

Not sure if I understand. The d.ts file is dynamically created based on the messages? If the d.ts file is static, there would be no typesafety for messages. Could you clarify how typesafety can be achieved without a dynamic .d.ts file? 

>  At least I was unable to find any .d.ts file inside the outDir. Or is this now necessary because of the virtual modules?

Because Paraglide compiles JS with JSDoc which TypeScript can handle. Virtual modules now take away the JSDoc definitions so TypeScript needs dedicated d.ts files. 

In other words, using virtual modules opens up complexity that Paraglide JS did not have before. It just compiled JS files with JSDoc. Works everywhere if the TS setting is set to `allowJs: true`. But, for some projects allow JS is not possible (https://github.com/opral/inlang-paraglide-js/issues/238) and users have an issue with compiled code in src (https://github.com/opral/inlang-paraglide-js/issues/189, https://github.com/opral/inlang-paraglide-js/issues/264).

inlang/source-code/paraglide/paraglide-js/example/tsconfig.json
LorisSigrist: Just done to accelerate compile times on my machines. Does this change affect the output of `sv`? If so, I'll reverse it.

manuel3108: Since this is only the example, it should not. Should be good to go!
I was worried that this might be a requirement, but this does not seem to be the case.

LorisSigrist:
One of the main reasons to use paraglide is the typesafe messages. We can't provide those unless there are dynamically generated type-definitions _somewhere_. A `.d.ts` seems like the most sensible place. 

I suggest that we place a `paraglide.d.ts` file in the `outdir` if `useVirtualModule` is specified. Seems like the most sensible thing

benmccann:
Hmm, yeah, I guess there's not much option except to generate the `.d.ts` file. Though at that point, and I hate to say this with all the work you've put into it already, but I wonder if generating virtual files is even worth it?

Another solution I just thought of would be to instead generate a package in `node_modules`. That would keep it out of `src` and let you generate a `.d.ts` file

samuelstroschein:
> Another solution I just thought of would be to instead generate a package in node_modules. That would keep it out of src and let you generate a .d.ts file

Doesn't work. HMR and everything breaks. Paraglide messages are source code. 

> I suggest that we place a paraglide.d.ts file in the outdir if useVirtualModule is specified. Seems like the most sensible thing

@LorisSigrist Yes, let's emit the d.ts file in the outdir and see how users react. Could you make the config experimental e.g. `experimentalUseVirtualModules` ?

benmccann:
> HMR and everything breaks.

Looks like you might be able to add a workaround in the Vite plugin to make that work: https://github.com/vitejs/vite/issues/8619#issuecomment-2254154530

samuelstroschein:
@LorisSigrist is a blocker remaining? Lmk when it's ready to review

samuelstroschein (changes requested):
Looks good, but one bug needs to be fixed. Then it can be merged. 

The `outdir` is not considered when creating the paths for the virtual modules. The virtual modules seem to emit by default into `$paraglide` instead of using the outdir path. 

CleanShot 2024-12-09 at 15 46 31@2x

CleanShot 2024-12-09 at 15 46 16@2x


https://github.com/user-attachments/assets/a825812a-844e-4078-97ed-17cebaebe471

LorisSigrist:
Ah, so you want the virtual files to still be importable via the regular file-path, not just a virtual namespace @samuelstroschein?

Should be doable, but we would probably need one `.d.ts` file per virtual file-path.

LorisSigrist:
Path imports can't be done via `declare module` in a single `.d.ts` file. The TS compiler won't resolve the path so relative imports and path-aliases like `$lib` won't work. 

SvelteKit generates an entirely separate filetree in the `.svelte-kit` folder just for types AFAIK. We won't have to do that, we'll just have `message.d.ts`, `runtime.d.ts` etc.

samuelstroschein:
> We won't have to do that, we'll just have message.d.ts, runtime.d.ts 

Sounds good. And I think is intuitive. 

- Virtual modules = the js files are omitted, only the declaration files emitted. 
- skipping per language declaration files seems fine


```.
└── src/
    └── paraglide/
        ├── messages.d.ts
        └── runtime.d.ts
```

samuelstroschein (approved):
Expanded the changelog and only emit the runtime.d.ts and messages.d.ts file.

Thank you loris! 

/tip 350

samuelstroschein:
@LorisSigrist how was the process for you with the tipping and PR in general?

PR #3253, opened 2024-12-09
feat: working proxy server for csv and fm

NiklasBuchfink:
Add a proxy server to host the website, file-manager and csv app on the same domain.

Follow Up: 
- website will be added by coping the current site to lix-integration
- update csv app links to fm

### DIFF
--- a/packages/lix-website-server/src/main.ts
+++ b/packages/lix-website-server/src/main.ts
@@ -50,7 +50,7 @@ const lixApps = [
   },
 ];
 
-lixApps.forEach((lixApp) => {
+for (const lixApp of lixApps) {
   app.use(
     `/app/${lixApp.route}`,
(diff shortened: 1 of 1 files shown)


packages/lix-website-server/src/main.ts
```diff
@@ -0,0 +1,75 @@
...
+  app.get(`/app/${lixApp.route}/*`, (req, res) => {
+    const indexPath = join(node_modules, `${lixApp.module}/dist/index.html`);
+    if (fs.existsSync(indexPath)) {
+      res.sendFile(indexPath);
+    } else {
+      res.status(404).send("File not found");
+    }
+  });
+});
+
```
samuelstroschein, line 67: Nice on. I would use a `for ... of` loop for better readability but this works too!

samuelstroschein (approved):

PR #3269, opened 2024-12-14
simplify state management

samuelstroschein:
- fixes database throw by not inserting into the lix_id column
- uses derived state instead of setting jotai state
- handles open lix logic with id in the navbar
- simplifies account login by only storing the active account

follow up tasks: 

- [x] query from the server
- [ ] beautify the id in the navbar (lix SDK job, NOT LIX FM) https://github.com/opral/lix-sdk/issues/189
- [x] csv app needs lix id logic in url too
- [ ] testing

feedback: 

- @felixhaeberle @NiklasBuchfink only derive state. don't mutate jotai. jotai is the derived state of lix and the url. 

https://www.loom.com/share/465d374f28ff49f09ec94fd9c36cd784?sid=ba32115c-936f-49a5-b0ba-bf91e284b4c6


felixhaeberle:
@samuelstroschein I've 

- updated the file manager state to fetch from server and also 
- started the implementation in the csv app 

but hadn't have time for testing, have to resume tomorrow or you take over as I have to run now.

PR #3276, opened 2024-12-16
feat: human ids read in a grammatically sound way

revosw:
By making sure all adjectives are in positive form and all verbs are in simple present tense, we get grammatically sound sentences


samuelstroschein (changes requested):
Thanks for the PR, I will close this one though. 

The list has been vetted against potentially inappropriate and long words. I am sure the words can be improved but I found several inappropriate words and the length increased.

PR #3303, opened 2024-12-25
Parjs-213-remove-vite-in-favor-of-tsc

samuelstroschein:
Closes https://github.com/opral/inlang-paraglide-js/issues/208


samuelstroschein:
Merging because failing CI is a ts-expect-error that is unused and i can't replicate it locally.

PR #3328, opened 2025-01-08
refactor: LIXDK-321 change "confirmed" to "checkpoint" and "intermediate"

NiklasBuchfink:
"Confirmed" was not understand or used in similar fashion like "save".

The concept of a checkpoint seem to fit the use and visual representation. It is also a known concept from video games for non-technical users.

Therefore:
- "confirmed" is renamed to "checkpoint"
- "confirm changes" is renamed to "create checkpoint"
- "unconfirmed changes" is renamed to "intermediate changes"

### DIFF
--- a/.changeset/lemon-cherries-unite.md
+++ b/.changeset/lemon-cherries-unite.md
@@ -1,0 +1,7 @@
+---
+"lix-file-manager": minor
+"csv-app": minor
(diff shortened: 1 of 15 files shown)


samuelstroschein (approved):
Let's go. Don't forget to write a changeset

PR #3340, opened 2025-01-12
Changed incorrect variable used

Secreto31126:
closes https://github.com/opral/inlang-paraglide-js/issues/310

This bug prevented the SDK from working on Windows due to a POSIX path conversion being performed but not used later.

Some notes while working on the repo with Windows:

 - When running `pnpm run ci` I encountered a few failed tests, all of them related to POSIX paths. However, none of them seem to be a bug, but rather a matter of running the checks on Windows. I wrote a test for preventing this specific bug that should fail both in Windows and Linux.
 - I almost accidentally commit a 320 files change due to CRLF 🤣. Maybe adding to .gitattributes `* text=auto eol=lf` to enforce LF endings would be cool.

### DIFF
--- a/.changeset/shiny-snakes-carry.md
+++ b/.changeset/shiny-snakes-carry.md
@@ -1,0 +1,20 @@
+---
+"@inlang/sdk": patch
+---
+
+fix: replaced wrong variable
+
+closes https://github.com/opral/inlang-paraglide-js/issues/310
+
+This bug prevented the SDK from working on Windows due to a POSIX path conversion being performed but not used later.
+
+```diff
+// inlang/packages/sdk/src/project/loadProjectFromDirectory.ts:550
+await args.lix.db
+    .insertInto("file") // change queue
+    .values({
+-       path: path,
++       path: posixPath,
+        data: new Uint8Array(data),
+    })
+```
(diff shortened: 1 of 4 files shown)


Secreto31126:

samuelstroschein:
I repeat, what a great interaction. Thank you so @Secreto31126! 

- wrote changesets
- added unit tests 

One test is failing, I will fix the test and merge your PR. Thanks for the contribution!

> When running pnpm run ci I encountered a few failed tests, all of them related to POSIX paths. However, none of them seem to be a bug, but rather a matter of running the checks on Windows. I wrote a test for preventing this specific bug that should fail both in Windows and Linux.

If you run the codebase with WSL, do the tests run through? If yes, I'd rather add that windows users should use WSL. 

> I almost accidentally commit a 320 files change due to CRLF 🤣. Maybe adding to .gitattributes * text=auto eol=lf to enforce LF endings would be cool.

https://github.com/opral/monorepo/commit/7fc6c3789f03b5ffd845f0699017a9f4e4009583 🫡

samuelstroschein:
Merging because you confirmed it works on your machine but without running the windows related test. the test is failing on mac and linux without the ability to mock the node.process.platform and subsequent path.sep variable. i tried for an hour.

PR #3353, opened 2025-01-15
Parjs 315 update astro adapter to v2

samuelstroschein:


samuelstroschein:
Closes in favor of #3354 because this PR has pnpm lock issues

PR #3363, opened 2025-01-21
Update strategy.md

dmsynge:
I have tried to explain in a bit more detail (but hopefully clearer/easier to understand) how the implementation fits into workings the paraglide runtime, fixed some typos/errors in the example code, and provided some specific implementation examples for Sveltekit. 

I didn't touch the #examples section, because I think perhaps it isn't needed -- if there's clear example code in each of the sections about implementation, then we can rely on users to know how to get/set cookies if that's what they want to do.


samuelstroschein (approved):
Awesome update to the docs! 

Thank you so much for contributing. I read through the diffs. Each paragraph you added increases comprehension. Hence, big big thank you!

dmsynge:

PR #3390, opened 2025-02-10
fix `saveProjectToDirectory`

felixhaeberle:
- fix currently not work when loadMessages & exportFiles is simultaneously defined
- fix not respect location of existing language files (e.g. pathPattern)


samuelstroschein:
Don't forget to write changesets to trigger deployments

samuelstroschein (approved):

PR #3399, opened 2025-02-14
fix: plugin imports on Bun

miikakokkonen:
Bun cannot dynamically import same format as node. 
Adds bun specific dynamic import and fixes sdk usage on Bun.

https://github.com/opral/inlang-paraglide-js/issues/335

### DIFF
--- a/inlang/packages/sdk/src/plugin/importPlugins.ts
+++ b/inlang/packages/sdk/src/plugin/importPlugins.ts
@@ -32,14 +32,14 @@ export async function importPlugins(args: {
 			let moduleAsURL;
 			if (process.versions.bun) {
 				// In bun we need to do dynamic imports differently
-				moduleAsURL = URL.createObjectURL(new Blob([moduleAsText], { type: 'text/javascript' }));
+				moduleAsURL = URL.createObjectURL(
+					new Blob([moduleAsText], { type: "text/javascript" })
+				);
 			} else {
 				// node and others
-				moduleAsURL = 'data:text/javascript;base64,' + btoa(moduleAsText);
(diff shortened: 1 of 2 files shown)


inlang/packages/sdk/src/plugin/importPlugins.ts
```diff
@@ -29,11 +29,17 @@ export async function importPlugins(args: {
...
-				"data:application/javascript," + encodeURIComponent(moduleAsText);
-			const { default: module } = await import(
-				/* @vite-ignore */ moduleWithMimeType
-			);
+			let moduleAsURL;
+			if (process.versions.bun) {
+				// In bun we need to do dynamic imports differently
+				moduleAsURL = URL.createObjectURL(
+					new Blob([moduleAsText], { type: "text/javascript" })
+				);
```
samuelstroschein, line 37: I think this works in node and the browser too. Are you sure that the conditional logic for bun is needed?

miikakokkonen: I thought so too but I could not find a way that works for both. 

a related issue:
https://github.com/nodejs/node/issues/47573

samuelstroschein: somehow the createObjectUrl API also makes problems in chrome. Leaving the flag here as is. Good to merge but with no guarantees that we test bun in CI/CD

CleanShot 2025-02-14 at 08 31 24@2x

samuelstroschein:
@miikakokkonen can you comment with `I have read the CLA Document and I hereby sign the CLA`

miikakokkonen:

PR #3401, opened 2025-02-14
Add better error handling to paragglide-js cli

miikakokkonen:
If an error is thrown while compiling, it will get added to the errors from loading the project.
The errors will be logged when the compile command is over.
if there are any errors, the success message is switched to warn.


miikakokkonen:
There is no changeset as @inlang/paraglide-js is in the list of ignored packages in .changeset/config.json

miikakokkonen:
@samuelstroschein

What are your opinions on the error messages and logging them?

Also, should i remove paraglide-js from the excluded list in .changesets so i can mark it as having changes?

samuelstroschein:
Looks good will merge this. 

> Also, should i remove paraglide-js from the excluded list in .changesets so i can mark it as having changes?

Not yet. Need the full GA release of paraglide js 2.0 Changesets and NPM play badly together with beta releases.

PR #3441, opened 2025-02-26
fix `serverMiddleware()` throws in next js when `url` strategy is not used 

samuelstroschein:
closes https://github.com/opral/inlang-paraglide-js/issues/411#issuecomment-2683530533


samuelstroschein:
tests run through locally. i assume the recent sherlock commits on main cause this (cc @felixhaeberle )

PR #3472, opened 2025-03-06
Change inlang landingpage to markdwon

jannesblobel:
Decided to render a Markdown doc on the landing page instead of having a designed page


samuelstroschein (approved):
Looks good. I didn't see the landing page online because we have no preview deployment but the beauty of markdown as landing page is that we can quickly iterate without writing code

PR #3487, opened 2025-03-10
make `setLocale()` set all strategies

samuelstroschein:
Setting all strategies aligns with user expectations and ensures that server APIs can receive the cookie of the client, for example.

closes https://github.com/opral/inlang-paraglide-js/issues/439


samuelstroschein:
closes in favor of https://github.com/opral/monorepo/pull/3488

PR #3507, opened 2025-03-16
Prevent infinite redirects due to trailing slash config

axekan:
Closes [opral/inlang-paraglide-js#408](https://github.com/opral/inlang-paraglide-js/issues/408)


axekan:

axekan:
recheck

axekan:
@samuelstroschein

I looked into this and it seems that this line:

https://github.com/opral/monorepo/blob/d658dcb4e49419bfdd1f5bbc8621330891932bd8/inlang/packages/paraglide/paraglide-js/src/compiler/server/middleware.js#L80

can add a trailing slash to an url in Cloudflare Workers. This means that this comparison will always fail, which causes the infinite redirects:

https://github.com/opral/monorepo/blob/d658dcb4e49419bfdd1f5bbc8621330891932bd8/inlang/packages/paraglide/paraglide-js/src/compiler/server/middleware.js#L81-L83

I updated the code to do this comparison by ignoring trailing slashes on the URL:s. I figured that should be fine since paraglide should have nothing to do with trailing slashes anyway, considering that its a framework configuration. With these changes, everything works for me locally and [deployed to Cloudflare pages](https://inlang-redirect-loop.pages.dev).

I wasn't able to get the tests to run locally [see nx.app logs](https://cloud.nx.app/runs/G7VSmXTpfI), but maybe you are able to test this?

Let me know how you want to proceed.

samuelstroschein:
@axekan willing to merge this! can you add a unit test to assure that the behavior won't chang in the future?

samuelstroschein:
the tests should be locally runnable with `pnpm --filter @inlang/paraglide-js test`

axekan:
@samuelstroschein Added some tests that should cover it. Is a changeset needed for this?

samuelstroschein:
@axekan no changeset needed

inlang/packages/paraglide/paraglide-js/src/compiler/server/middleware.test.ts
```diff
@@ -373,6 +373,170 @@ test.skip("doesn't redirect if disableUrlRedirect is true", async () => {
...
+			settings: {
+				baseLocale: "en",
+				locales: ["en", "fr"],
+			},
+		}),
+		compilerOptions: {
+			strategy: ["url", "baseLocale"],
+			urlPatterns: [
+				{
+					pattern: "https://example.com/:path(.*)?",
```
samuelstroschein, line 388: @axekan why would the middleware not execute if a path has a trailing slash or not?

samuelstroschein: Or why is the middleware execution dependent on trailing slashes and non trailing slashes. The test seems reundant

samuelstroschein:
@axekan the tests pass even if `normalizeUrl()` is commented out

samuelstroschein:
Here is why. The URLPattern already prunes trailing slashes. Hence, it must be the default pattern solution which is shipped

CleanShot 2025-03-16 at 17 33 43@2x

samuelstroschein:
I couldn't get my tests to fail either. Given that you stated this works when deployed, I am going to merge this as is

PR #3532, opened 2025-03-24
add cookieMaxAge option to compiler and runtime

seriousm4x:
closes https://github.com/opral/inlang-paraglide-js/issues/483

- Added cookieMaxAge option to setLocale, allowing control over the cookie expiration time.
- Default cookieMaxAge is set to 400 days (Chrome’s max-age limit).
- Updated getLocale tests to verify max-age is set in cookies.
- Fixed minor typos in documentation.


seriousm4x:

samuelstroschein:
@seriousm4x thanks! Why did you choose to set the cookie expiration in `setLocale()` instead of the compiler options? 

I would think that the compiler options are easier. All central options at once place, like the cookie name.

seriousm4x:
I thought it would be more flexible if someone wanted to change the time dynamically rather than having it fixed for the entire project. I don’t have a specific use case for this, but others might.

samuelstroschein:
@seriousm4x let's go for the compiler option. otherwise, someone might call `setLocale()` somewhere in the code with the cookie option and at another place without it and wonders why the cookie duration is changing. 

can you update the PR?

seriousm4x:
Removed from `setLocale()` and added to the compiler options.

Let me know if I'm missing something as I'm not familiar with the codebase.

samuelstroschein (approved):
dope PR thank you!

PR #3534, opened 2025-03-25
Update registry.json(plugin-message-format-hjson-toml)

ooopus:
Added support for TOML and HJSON. Now, in many cases, we no longer need to use quotation marks.
https://github.com/opral/inlang-paraglide-js/issues/475#event-16948656729


ooopus:

PR #3537, opened 2025-03-26
rename `lix-server-api` to `lix-server-protocol`

felixhaeberle:
cc https://github.com/opral/lix-sdk/issues/250


felixhaeberle:
cc @samuelstroschein I think there is something wrong with how the schema is exported. 
What is the desired behavior here – can you have a look?

link: https://cloud.nx.app/runs/RIQmbu8IEJ

samuelstroschein:
@felixhaeberle the problem was a non type import 

```diff
-import * as ServerProtocol
+import type * as ServerProtocol
```

i also moved the package into lix/packages. THANKS FOR JUMPING RIGHT INTO ACTION <3

PR #3538, opened 2025-03-26
Fix dependency path from async_hooks to node:async_hooks

adrian-budau:
Without this fix the generated middleware would break deno.


adrian-budau:

samuelstroschein:
@adrian-budau with prefix breaks webpack and nextjs :/ does deno have an issue open for the async hooks?

samuelstroschein:
Closing this as it's unclear how webpack will work with this change

PR #3539, opened 2025-03-26
update json handling to always use jsonb

felixhaeberle:
cc https://github.com/opral/lix-sdk/issues/283


packages/lix-sdk/src/database/json.example.ts
```diff
@@ -0,0 +1,162 @@
...
+			}),
+		})
+		.execute();
+
+	// Insert metadata as JSON object for file
+	await db
+		.insertInto("file")
+		.values({
+			path: "/example/path.txt",
+			data: Buffer.from("Hello World"), // data as binary
```
samuelstroschein, line 29: use `new TextEncoder().encode("Hello World")`. `Buffer` is a node api

packages/lix-sdk/src/database/index.ts
```diff
@@ -1,5 +1,7 @@
 export * from "./schema.js";
 export * from "../key-value/database-schema.js";
-export { jsonObjectFrom, jsonArrayFrom } from "kysely/helpers/sqlite";
+// Replace the kysely helpers with our own implementations for JSONB
+export { jsonObjectFrom, jsonArrayFrom, jsonb } from "./json.js";
```
samuelstroschein, line 4: The `jsonObjectFrom` and `jsonArrayFrom` are likely not needed. The SQLite functions used in the `json_array()` Kysely helper should work on jsonb as well. 

If not, all up for having our own jsonObjectFrom and jsonArrayFrom functions.

samuelstroschein (commented):
Looks good so far. Needs the following: 

- add the jsonb columns to https://github.com/opral/monorepo/blob/7ac4e611611b07ee64f52bf5ea0ed274d4b15520/packages/lix-sdk/src/database/init-db.ts#L25-L33 

- update the apps to insert jsonb

packages/lix-sdk/src/database/json.example.ts
felixhaeberle: done 👍

packages/lix-sdk/src/database/index.ts
felixhaeberle: true 👍

felixhaeberle:
> add the jsonb columns to

correct me if wrong, but I did not find any other BLOB (JSON) fields – what where you referring to?

> update the apps to insert jsonb

this is also not needed, as the sdk does it automatically with provided json by the apps? -> BUT: had to adjust `create-change` and `create-snapshot` to use the `jsonb` function

> target the lix-v0.5 branch, not main

done ✅

packages/lix-sdk/src/snapshot/create-snapshot.ts
```diff
@@ -20,7 +21,7 @@ export async function createSnapshot(args: {
 		const snapshot = await trx
 			.insertInto("snapshot")
 			.values({
-				content: args.content ?? null,
+				content: args.content ? jsonb(args.content) : null,
```
samuelstroschein, line 24: Using the jsonb() functoin shouldn't be needed. the `SerializeJsonbPlugin` in `init-db` should take care of it

samuelstroschein:
>> add the jsonb columns to
> correct me if wrong, but I did not find any other BLOB (JSON) fields – what where you referring to?

@felixhaeberle you need to add the jsonb columns to the parsing plugin https://github.com/opral/monorepo/pull/3539#pullrequestreview-2718890085

samuelstroschein:
Oh and please add a minor change set to get a good changelog.md for v0.5 .

samuelstroschein:
The PR is not adjusting the schema which was the desired change

PR #3575, opened 2025-04-29
Memoize extractLocaleFromUrl

mquandalle:
As reported on [Discord](https://discord.com/channels/897438559458430986/1366763869464301672), I have a performance issue when navigating to a page with about 200 `localizeHref` calls. This comes from the repeated calls to `extractLocaleFromUrl` even if the URL and the extract locale is constant on a given page.

This PR adds a cache to memoize this function.

### DIFF
--- a/inlang/packages/paraglide/paraglide-js/src/compiler/runtime/extract-locale-from-url.js
+++ b/inlang/packages/paraglide/paraglide-js/src/compiler/runtime/extract-locale-from-url.js
@@ -25,11 +25,11 @@ let cachedLocale;
  */
 export function extractLocaleFromUrl(url) {
 	const urlString = typeof url === "string" ? url : url.href;
-	
+
 	if (cachedUrl === urlString) {
 		return cachedLocale;
 	}
-	
+
 	let result;
 	if (TREE_SHAKE_DEFAULT_URL_PATTERN_USED) {
 		result = defaultUrlPatternExtractLocale(url);
@@ -56,7 +56,7 @@ export function extractLocaleFromUrl(url) {
 			if (result) break;
 		}
 	}
-	
+
 	cachedUrl = urlString;
 	cachedLocale = result;
 	return result;
--- a/inlang/packages/paraglide/paraglide-js/src/compiler/runtime/extract-locale-from-url.js
+++ b/inlang/packages/paraglide/paraglide-js/src/compiler/runtime/extract-locale-from-url.js
@@ -9,8 +9,13 @@ import {
 /**
  * If extractLocaleFromUrl is called many times on the same page and the URL
  * hasn't changed, we don't need to recompute it every time which can get expensive.
+ * We might use a LRU cache if needed, but for now storing only the last result is enough.
+ * https://github.com/opral/monorepo/pull/3575#discussion_r2066731243
  */
-const urlToLocaleCache = new Map();
+/** @type {string|undefined} */
+let cachedUrl;
+/** @type {Locale|undefined} */
+let cachedLocale;
 
 /**
  * Extracts the locale from a given URL using native URLPattern.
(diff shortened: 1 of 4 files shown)


inlang/packages/paraglide/paraglide-js/src/compiler/runtime/extract-locale-from-url.js
```diff
@@ -6,38 +6,54 @@ import {
...
 			}
+			if (result) break;
 		}
 	}
-
-	return undefined;
+	
+	urlToLocaleCache.set(urlString, result);
+	return result;
 }
```
mquandalle, line 58: I get the following CI error

```
/home/runner/work/monorepo/monorepo/inlang/packages/paraglide/paraglide-js/src/compiler/runtime/extract-locale-from-url.js
  65:10  error  'defaultUrlPatternExtractLocale' is defined but never used  @typescript-eslint/no-unused-vars
```

which isn't related to my change

inlang/packages/paraglide/paraglide-js/src/compiler/runtime/extract-locale-from-url.js
```diff
@@ -6,38 +6,54 @@ import {
...
+					break;
+				}
 			}
+			if (result) break;
 		}
 	}
-
-	return undefined;
+	
+	urlToLocaleCache.set(urlString, result);
```
samuelstroschein, line 55: The cache will infinitely grow in size. 

If we do this, we should have some sort of LRU cache. This might be enough: 

```ts
class LRUCache {
  constructor(max = 100) {
    this.max = max;
    this.map = new Map();
  }
  get(key) {
    if (!this.map.has(key)) return undefined;
    const val = this.map.get(key);
    // refresh insertion order
    this.map.delete(key);
    this.map.set(key, val);
    return val;
  }
  set(key, val) {
    if (this.map.has(key)) this.map.delete(key);
    this.map.set(key, val);
    // if we’re over capacity, drop the oldest
    if (this.map.size > this.max) {
      const oldestKey = this.map.keys().next().value;
      this.map.delete(oldestKey);
    }
  }
}

const cache = new LRUCache(200);
```

samuelstroschein (changes requested):

inlang/packages/paraglide/paraglide-js/src/compiler/runtime/extract-locale-from-url.js
mquandalle: Hmm, maybe to make it even simpler, storing a single pair `url`, `locale` is enough? Or are there scenarios where `extractLocaleFromUrl` is called on a set of repeated distinct URLs?

samuelstroschein: On the server extractLocaleFromUrl can be called for N routes. But given that your problem was client side, starting with a  single pair should fix the client side performance issues

mquandalle: On my home page https://esim.info, `extractLocalFromUrl` is called 1500 times, always with the same argument. So caching the result is sensible. I don't think we need to cache more than one URL but I'm open to do it.

samuelstroschein: Let's do the simple implementation. Avoids caching pitfalls

mquandalle: I used `console.count` on the server and `extractLocalFromUrl` isn't called as many time (1500 times on the client to load my home page, only 4 times on the server to SSR it)

PR #3583, opened 2025-05-07
Docs: clarify localizeHref() requirement for URL localization in v2

philippviereck:
In response to: https://github.com/opral/inlang-paraglide-js/issues/485#issuecomment-2859636254

### DIFF
--- a/inlang/packages/paraglide/paraglide-js/docs/basics.md
+++ b/inlang/packages/paraglide/paraglide-js/docs/basics.md
@@ -111,8 +111,7 @@ console.log(getLocale()); // "de"
 ```
 
 ## Routing
-<doc-callout type="info">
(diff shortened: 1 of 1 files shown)


philippviereck:

samuelstroschein:
@philippviereck no preview unavailable, unfortunately. If you change the <a> to a markdown link, things should work!

philippviereck:
Are there other open issues besides https://github.com/opral/inlang-paraglide-js/issues/485, which is specifically for sveltekit, for other frameworks which used to have this feature as part of their adapter?

I'd like to link to those because I believe it makes this requirement less "scary" for people if they know there's open issues about finding a solution to the current, stable but quite manual approach and hopefully this will lead to some constructive ideas or alternatives. - instead of "+1"s

PR #3584, opened 2025-05-10
next-intl plugin support for multiple namespaces in usetranslations hooks

philipaarseth:
## Enhanced Namespace Handling in Message Reference Matching

This update significantly improves how message references are matched when using multiple `useTranslations` or `getTranslations` calls with different namespaces within the same file or component. The parser can now accurately associate a translation function call (e.g., `t("messageKey")`) with its correct namespace, even when multiple such translation functions are defined with varying namespaces.

### New Functionality

The primary enhancement is the ability to correctly resolve message IDs in scenarios where multiple translation contexts (namespaces) are active. For example:

```typescript
// previously might have led to incorrect namespace association
const tOne = useTranslations("namespaceOne");
const tTwo = useTranslations("namespaceTwo");

// ...

tOne("greeting"); // Correctly resolves to "namespaceOne:greeting"
tTwo("farewell"); // Correctly resolves to "namespaceTwo:farewell"
```

The system now considers the declaration position of each `useTranslations` (or `getTranslations`) call and associates subsequent translation function calls with the most recent, relevant namespace declaration in scope.

### Code Changes

1.  **`inlang/packages/plugins/next-intl/src/ideExtension/messageReferenceMatchers.ts`**:
    *   **`parseNameSpaces` function**:
        *   Modified to capture and store the `declarationPosition` (character offset in the source code) for each `useTranslations` and `getTranslations` declaration. This allows for determining which namespace was active at a specific point in the code.
    *   **`FunctionNameToNamespaces` type**:
        *   Updated its structure. Instead of a direct mapping from a function name to a single namespace, it now maps a function name (e.g., `t`, `tOne`) to an *array* of namespace objects. Each object in this array contains the namespace string (`ns`) and its `declarationPosition`. This handles cases where the same variable name (like `t`) might be re-declared or shadowed with different namespaces.
    *   **`getFinalMessageId` function**:
        *   Logic was enhanced to use the new `declarationPosition` information. When a translation function (e.g., `tOne("messageKey")`) is encountered, this function identifies the invoked function/variable name (e.g., `tOne`). It then searches through all namespace declarations associated with this specific name.
        *   It selects the namespace declaration whose `declarationPosition` (character offset) is the largest (most recent) value that is still *before* the starting position of the current translation function call. This ensures that the message key is prefixed with the namespace from the correct, most recent, and preceding in-scope declaration for that specific function/variable name.
        *   **Disclaimer: Message references are matched to namespaces based on the declaration order of translation hooks (like \`useTranslations\`) and the specific variable name used for the \`t\` function; this is a heuristic.** It does not fully understand JavaScript's lexical scoping rules (e.g., block scope, closures) or component boundaries in frameworks like React. For instance, in nested React components, the "most recent preceding namespace declaration for that function name" found in the file might not be the one truly in scope for a child component if namespaces are defined in parent components. For perfect scope-aware namespace resolution across the entire application (including component hierarchy), a more complex Abstract Syntax Tree (AST) parser and potentially inter-file analysis would be required. However, this method provides a significant improvement for common use cases within a single file.

2.  **`inlang/packages/plugins/i18next/src/ideExtension/messageReferenceMatchers.test.ts`**:
    *   A new test case, `should correctly reference messages with multiple useTranslations hooks`, was added.
    *   This test specifically validates the new multi-namespace handling by defining several `useTranslations` calls with distinct namespaces and then invoking their respective translation functions. It asserts that each message ID is correctly prepended with its corresponding namespace.

These changes provide more robust and accurate message reference matching in complex components that utilize multiple internationalization namespaces.

### DIFF
--- a/inlang/packages/plugins/next-intl/src/ideExtension/messageReferenceMatchers.test.ts
+++ b/inlang/packages/plugins/next-intl/src/ideExtension/messageReferenceMatchers.test.ts
@@ -313,3 +313,35 @@ it("should add the defined namespace by useTranslations hook", async () => {
 	expect(matches).toHaveLength(1);
 	expect(matches[0]?.messageId).toBe("DirectoryPage.title");
 });
+
+
+it("should correctly reference messages with multiple useTranslations hooks", async () => {
+	const sourceCode = `
+		const tOne = useTranslations("namespaceOne");
+		const tTwo = useTranslations("namespaceTwo");
+		const tThree = useTranslations("namespaceThree");
+		const tNested = useTranslations("nestedNamespace.section");
+
+		// Simulate usage of these translation functions
+		const message1 = tOne("messageKey1");
+		const message2 = tTwo("messageKey2");
+		const message3 = tThree("messageKey3");
+		const message4 = tNested("messageKey4");
+
+		// Interleaved usage
+		const message5 = tOne("anotherKey1");
+		const message6 = tTwo("anotherKey2");
+	`;
+	const settings: PluginSettings = {
+		pathPattern: "./{language}.json",
+	};
+	const matches = parse(sourceCode, settings);
+
+	expect(matches).toHaveLength(6);
+	expect(matches[0]?.messageId).toBe("namespaceOne.messageKey1");
+	expect(matches[1]?.messageId).toBe("namespaceTwo.messageKey2");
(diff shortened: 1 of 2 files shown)


samuelstroschein:
Awesome PR @philipaarseth. Thank you very much! The patch will be released with https://github.com/opral/monorepo/pull/3581 which is expected to be merged around Wednesday

PR #3590, opened 2025-05-27
add solidstart example

HokkaidoInu:


inlang/packages/paraglide/paraglide-js/examples/solidstart/src/app.tsx
```diff
@@ -0,0 +1,31 @@
...
+          <Suspense>{props.children}</Suspense>
+        </MainLayout>
+      )}
+    >
+      <Switch fallback={<FileRoutes />}>
+        <Match when={getLocale() === "de"}>
+          <Route path="/de/" component={Home} />
+          <Route path="/de/ueber-uns" component={About} />
+          <Route path="*404" component={NotFound} />
+        </Match>
```
samuelstroschein, line 27: Do you need to hardcode the localized routes for matching?

HokkaidoInu: Yes I believe so. I tried without hardcoding and it gives 404 error.

HokkaidoInu: I just found out I could use a catch-all route. But I'm not sure if this would be good or not.

**src/app.tsx** contents would change to this:

```
import { Router } from '@solidjs/router';
import { FileRoutes } from '@solidjs/start/router';
import { Suspense } from 'solid-js';

import "./app.css";

import MainLayout from './ui/layout/main-layout';

export default function App() {
  return (
    <Router
      root={(props) => (
        <MainLayout>
          <Suspense>{props.children}</Suspense>
        </MainLayout>
      )}
    >
      <FileRoutes />
    </Router>
  );
}
```

**src/routes/[...404].tsx** would be replaced by **src/routes/[...lang].tsx** which basically does the following:

- Matches any route not already covered by a static file in src/routes
- Looks up the canonical route (e.g. /about) using urlPatterns from src/paraglide/runtime.js
- Then dynamically loads the correct component (like about.tsx) based on that canonical path.

The contents would be this:

```
import { useLocation } from "@solidjs/router";
import { lazy } from "solid-js";
import { urlPatterns } from "~/paraglide/runtime";

// Load all route files recursively (including dynamic and nested)
const pages = import.meta.glob("/src/routes/**/*.tsx");

const routeMap = new Map(
  urlPatterns.flatMap(({ pattern, localized }) =>
    localized.map(([lang, path]) => [path, pattern])
  )
);

function matchCanonicalToFile(canonical: string): string | null {
  // Normalize pattern like /projects/123/details to segments
  const parts = canonical.split("/").filter(Boolean);

  for (const pagePath of Object.keys(pages)) {
    const routePath = pagePath
      .replace("/src/routes", "")
      .replace(/\.tsx$/, "")
      .split("/")
      .filter(Boolean);

    if (routePath.length !== parts.length) continue;

    const matches = routePath.every((segment, i) =>
      segment.startsWith("[") ? true : segment === parts[i]
    );

    if (matches) return pagePath;
  }

  return null;
}

export default function CatchAllRoute() {
  const location = useLocation();
  const currentPath = location.pathname;
  const canonicalPath = routeMap.get(currentPath);

  const filePath = canonicalPath ? matchCanonicalToFile(canonicalPath) : null;

  const Component = lazy(() => {
    const importer = filePath && pages[filePath];
    return importer
      ? importer()
      : Promise.resolve({ default: () => <div>404 - Not Found</div> });
  });

  return <Component />;
}
```

samuelstroschein:
I  am closing this as "unsure how to maintain the example for solidstart [and solidstart has virtually no adoption to make it worth it]". 

@HokkaidoInu thanks for the PR!

PR #3594, opened 2025-05-30
perf(paraglide): improve bundle size for fully translated messages

Le0Developer:
Closes opral/inlang-paraglide-js#521

This could be further optimized (eg for cases where multiple languages return the same message, that branch could be moved to the end to remove the comparisons) but that's for another time.
(afaik SWC is already smart enough to optimize `cond1 ? a() : cond2 ? b() : a()` into `cond2 ? b() : a()` if cond1 has no side-effects, which applies here)

### DIFF
--- a/inlang/packages/paraglide/paraglide-js/src/compiler/compile-bundle.test.ts
+++ b/inlang/packages/paraglide/paraglide-js/src/compiler/compile-bundle.test.ts
@@ -1,6 +1,6 @@
 import { test, expect } from "vitest";
 import { compileBundle } from "./compile-bundle.js";
-import type { BundleNested } from "@inlang/sdk";
+import type { BundleNested, ProjectSettings } from "@inlang/sdk";
 import { toSafeModuleId } from "./safe-module-id.js";
 
 test("compiles to jsdoc", async () => {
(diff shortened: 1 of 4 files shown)


inlang/packages/paraglide/paraglide-js/src/compiler/compile-bundle.test.ts
```diff
@@ -62,8 +65,7 @@ export const blue_moon_bottle = (inputs, options = {}) => {
 	const locale = options.locale ?? getLocale()
 	trackMessageCall("blue_moon_bottle", locale)
 	if (locale === "en") return en.blue_moon_bottle(inputs)
-	if (locale === "en-US") return en_us2.blue_moon_bottle(inputs)
-	return "blue_moon_bottle"
+	return en_us2.blue_moon_bottle(inputs)
 };"`
 	);
 });
```
samuelstroschein, line 71: Can you add a unit test for a bundle that has a missing message and therefore falls back to the id?

Le0Developer: Added

PR #3597, opened 2025-06-04
Implement custom strategy concept

akkie:
Implements a way to define custom strategies based on the discussions in https://github.com/opral/inlang-paraglide-js/issues/518

Fixes https://github.com/opral/inlang-paraglide-js/issues/518

### DIFF
--- a/inlang/packages/paraglide/paraglide-js/src/compiler/runtime/strategy.js
+++ b/inlang/packages/paraglide/paraglide-js/src/compiler/runtime/strategy.js
@@ -1,0 +1,77 @@
+/**
+ * @typedef {"cookie" | "baseLocale" | "globalVariable" | "url" | "preferredLanguage" | "localStorage"} BuiltInStrategy
+ */
+
+/**
+ * @typedef {`custom_${string}`} CustomStrategy
+ */
+
+/**
+ * @typedef {BuiltInStrategy | CustomStrategy} Strategy
+ */
+
+/**
+ * @typedef {Array<Strategy>} Strategies
+ */
+
+/**
+ * @typedef {{ getLocale: (request?: Request) => string | undefined }} CustomServerStrategyHandler
+ */
+
+/**
+ * @typedef {{ getLocale: () => string | undefined, setLocale: (locale: string) => void }} CustomClientStrategyHandler
+ */
+
+export const customServerStrategies = new Map();
+export const customClientStrategies = new Map();
+
+/**
+ * Checks if the given strategy is a custom strategy.
+ *
+ * @param {any} strategy The name of the custom strategy to validate.
+ * Must be a string that starts with "custom-" followed by alphanumeric characters.
+ * @returns {boolean} Returns true if it is a custom strategy, false otherwise.
+ */
+export function isCustomStrategy(strategy) {
+	return typeof strategy === "string" && /^custom-[A-Za-z0-9]+$/.test(strategy);
+}
+
+/**
+ * Defines a custom strategy that is executed on the server.
+ *
+ * @param {any} strategy The name of the custom strategy to define. Must follow the pattern `custom-<name>` where
+ * `<name>` contains only alphanumeric characters.
+ * @param {CustomServerStrategyHandler} handler The handler for the custom strategy, which should implement
+ * the method `getLocale`.
+ * @returns {void}
+ */
+export function defineCustomServerStrategy(strategy, handler) {
+	if (!isCustomStrategy(strategy)) {
+		throw new Error(
+			`Invalid custom strategy: "${strategy}". Must be a custom strategy following the pattern custom-<name>` +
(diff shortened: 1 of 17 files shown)


inlang/packages/paraglide/paraglide-js/src/compiler/runtime/strategy.js
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Checks if the given strategy is a custom strategy.
+ *
+ * @param {any} strategy The name of the custom strategy to validate.
+ * Must be a string that starts with "custom-" followed by alphanumeric characters.
+ * @returns {boolean} Returns true if it is a custom strategy, false otherwise.
+ */
+export function isCustomStrategy(strategy) {
+	return typeof strategy === "string" && /^custom-[A-Za-z0-9]+$/.test(strategy);
```
samuelstroschein, line 36: very good assertion that custom strategies have to start with `custom*` to avoid namespace conflicts

inlang/packages/paraglide/paraglide-js/src/compiler/runtime/strategy.js
```diff
@@ -0,0 +1,59 @@
...
+}
+
+/**
+ * Defines a custom strategy that is executed on the server or on the client.
+ *
+ * @param {any} strategy The name of the custom strategy to define. Must follow the pattern `custom-<name>` where
+ * `<name>` contains only alphanumeric characters.
+ * @param {ServerCustomStrategyHandler | ClientCustomStrategyHandler} handler The handler for the custom strategy, which
```
samuelstroschein: The overload seems tricky. How would the calling code know if it's running client-side or server-side? 

Defining a strategy as an object seems more explicit. Why did you opt for the overload? 

```ts
defineCustomStrategy({
   server: { getLocale }, 
   client: { getLocale, setLocale }
})

```

samuelstroschein: I guess it's this

```ts
defineCustomStrategy({
  getLocale: ({ request }) => {
     if (import.meta.env.SSR){
        // server side
        
        // how would ts / the caller know that request is defined? 
        request.X
     } else {
        // client side
     }
  }
})

```

My hesitation is the passing of the request on the server for `getLocale()` and `setLocale`.  

We could expose an `isServer` variable on `getLocale()` which can be used to check the context. 

```ts

defineCustomStrategy({
  getLocale: ({ isServer, request }) => {
     if (isServer){
        // request is not undefined
        request.X
     } else {
        // client side
       // request is undefined
     }
  }
})

```

akkie: I assumed that the code for the server and client would be kept separate. This is how I would implement it with SvelteKit. For the server part, I would implement it in the server hooks. I use a dedicated database connection that can only be used on the server. The database code includes server secrets and including these in the client code would result in an error.

I also have a separate database connection for the client that uses a JWT token to connect to the database. This code should only be loaded on the client.

I think the problem with your approach is that code from the server may be loaded on the client side. So my idea was that ServerCustomStrategyHandler should only be used on the server and ClientCustomStrategyHandler on the client, to ensure that no server code is loaded on the client and vice versa.

For use cases where you need to define code in a file shared between the server and client, you could use something like this I guess:

```js
if (import.meta.env.SSR) {
  defineCustomStrategy({
    getLocale: (request) => {
       
    }
  })
} else {
  defineCustomStrategy({
    getLocale: () => {
       
    },
    setLocale: (locale) => {

    }
  })
}
```

or

```js

defineCustomStrategy(import.meta.env.SSR ? {
  getLocale: (request) => {
       
  }
} : {
  getLocale: () => {
       
  },
  setLocale: (locale) => {

  }
})
```

Does this answer your questions? However, it's possible that I haven't understood your point. If so, could you please provide more details about your concerns?

akkie: But maybe it would make more sense if we made it more explicit, as you suggested. Something like:

```js
defineCustomServerStrategy({
  getLocale: (request) => {
       
  }
});
defineCustomClientStrategy({
  getLocale: () => {
       
  },
  setLocale: (locale) => {

  }
});
```

samuelstroschein: The more explicit approach is better IMO. Otherwise, the arguments for a functions are unclear/not typesafe. 

I like the explicit `defineCustomServerStrategy` and `defineCustomClientStrategy` API. What do you think?

akkie: Should be ready now

samuelstroschein (approved):
Very much appreciated. Deploy will happen with https://github.com/opral/monorepo/pull/3581. ETA end of next week

PR #3600, opened 2025-06-12
Fix the type of the define custom strategy factories

shivan-s:
Closes https://github.com/opral/inlang-paraglide-js/issues/527

First official PR into this repo editing code. Hope this is done right and that it helps.


shivan-s:
Closing, this PR does not fix the issue.

See conversation in the original issue: https://github.com/opral/inlang-paraglide-js/issues/527

PR #3601, opened 2025-06-15
Make `extractLocaleFromRequest` asynchronous

shivan-s:
This will make `extractLocaleFromRequest` asynchronous. This means that it will return a Promise and if this is being used in code, all instances will need to be awaited.

The reason this change was made was because if a user supplied a custom server strategy via the `defineCustomServerStrategy` and defined a `getLocale` that made an asynchronous database call for example, this would mean that the `extractRequestFromLocale` would have to be made asynchronous.

Users should await `extractLocaleFromRequest` if it is used in code.

Closes https://github.com/opral/inlang-paraglide-js/issues/527

### DIFF
--- a/inlang/packages/paraglide/paraglide-js/src/compiler/runtime/extract-locale-from-request.test.ts
+++ b/inlang/packages/paraglide/paraglide-js/src/compiler/runtime/extract-locale-from-request.test.ts
@@ -18,7 +18,7 @@ test("returns the locale from the cookie", async () => {
 			cookie: `PARAGLIDE_LOCALE=fr`,
 		},
 	});
-	const locale = runtime.extractLocaleFromRequest(request);
+	const locale = await runtime.extractLocaleFromRequest(request);
 	expect(locale).toBe("fr");
 });
 
@@ -43,7 +43,7 @@ test("returns the locale from the pathname for document requests", async () => {
 			"Sec-Fetch-Dest": "document",
 		},
 	});
-	const locale = runtime.extractLocaleFromRequest(request);
+	const locale = await runtime.extractLocaleFromRequest(request);
 	expect(locale).toBe("en");
 });
 
@@ -58,7 +58,7 @@ test("returns the baseLocale if no other strategy matches", async () => {
 		strategy: ["baseLocale"],
 	});
 	const request = new Request("http://example.com");
-	const locale = runtime.extractLocaleFromRequest(request);
+	const locale = await runtime.extractLocaleFromRequest(request);
 	expect(locale).toBe("en");
 });
 
@@ -73,7 +73,7 @@ test("throws an error if no locale is found", async () => {
 		strategy: ["cookie"],
 	});
 	const request = new Request("http://example.com");
-	expect(() => runtime.extractLocaleFromRequest(request)).toThrow(
+	expect(runtime.extractLocaleFromRequest(request)).rejects.toThrowError(
 		"No locale found. There is an error in your strategy. Try adding 'baseLocale' as the very last strategy."
 	);
 });
@@ -95,7 +95,7 @@ test("returns the preferred locale from Accept-Language header", async () => {
 			"Accept-Language": "en-US;q=0.8,fr;q=0.9,de;q=0.6",
 		},
 	});
-	const locale = runtime.extractLocaleFromRequest(request);
+	const locale = await runtime.extractLocaleFromRequest(request);
 	expect(locale).toBe("fr");
 
 	// language tag match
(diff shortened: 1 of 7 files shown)


inlang/packages/paraglide/paraglide-js/src/compiler/runtime/strategy.js
```diff
@@ -15,11 +15,11 @@
  */
 
 /**
- * @typedef {{ getLocale: (request?: Request) => string | undefined }} CustomServerStrategyHandler
+ * @typedef {{ getLocale: (request?: Request) => Promise<string | undefined> | (string | undefined) }} CustomServerStrategyHandler
```
shivan-s, line 18: I wonder, is this how we define the type?

inlang/packages/paraglide/paraglide-js/src/compiler/runtime/strategy.js
```diff
@@ -15,11 +15,11 @@
...
  */
 
 /**
- * @typedef {{ getLocale: () => string | undefined, setLocale: (locale: string) => void }} CustomClientStrategyHandler
+ * @typedef {{ getLocale: () => Promise<string|undefined> | (string | undefined), setLocale: (locale: string) => Promise<void> | void }} CustomClientStrategyHandler
```
shivan-s, line 22: Revert this change.

inlang/packages/paraglide/paraglide-js/src/compiler/runtime/extract-locale-from-request.test.ts
```diff
@@ -286,7 +288,55 @@ test("returns locale from custom strategy", async () => {
...
+	});
+
+	class FakeDB {
+		db = new Map<string, string>();
+		constructor() {
+			this.db.set("1", "fr");
+		}
+
+		async getUserLocaleById(id: string) {
+			setTimeout(() => null, 4);
```
shivan-s:
```suggestion
			setTimeout(() => null, 4);
```

This could be removed. If you declare a method to be `async`, it's enough for it to be `async`.

inlang/packages/paraglide/paraglide-js/src/compiler/runtime/strategy.js
samuelstroschein: Seems correct. Provide either an async funcition or a sync one

shivan-s:
Hey @samuelstroschein, tried my best with this PR but I am stuck on two things:
1. these tests are failing: 
```
  FAIL  src/compiler/compile-project.test.ts > options > ./messages.js types
       FAIL  src/compiler/compile-project.test.ts > options > ./messages.js types
      AssertionError: expected 1 to deeply equal +0

      - Expected
      + Received

      - 0
      + 1

       ❯ src/compiler/compile-project.test.ts:936:31
          934|     console.error(diagnostic.messageText, diagnostic.file?.fileName);
          935|    }
          936|    expect(diagnostics.length).toEqual(0);
             |                               ^
          937|   });
          938|  }

      ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/2]⎯

       Test Files  1 failed | 35 passed (36)
            Tests  2 failed | 236 passed | 5 skipped (243)
         Start at  23:07:50
         Duration  9.91s (transform 1.20s, setup 0ms, collect 19.37s, tests 28.54s, environment 7ms, prepare 3.39s)

       ELIFECYCLE  Test failed. See above for more details.
```

Is it some warning that I can't use async?

2. Am in the right direction? I know you mentioned to remove the functionality for custom strat in the original sync version of `extractLocaleFromRequest`, but I will have to remove some tests, plus, if you are using the sync API would you expect your custom strat to also be considered (maybe we have to add some kind of warning to stay `extractLocaleFromRequest` won't work properly if you have an async `getLocale()` in which case use the async version. Plus I left there just in case people are using it and have a sync `getLocale()`.

Thanks for your help in advance, I understand if you don't have the capacity. I just don't have the skill! So thanks for the oppourtunity.

samuelstroschein:
@shivan-s log the diagnostics. some type error popped up. 

Regarding 2, i would add in the jsdoc that `extractLocaleFromRequest` does not support a custom server side strategy. `extractLocaleFromRequestAsync` has to be used.

samuelstroschein:
You might have run into https://github.com/opral/inlang-paraglide-js/issues/529 regarding the bug. Lmk if I have to take over the PR or if you want to complete it

shivan-s:
@samuelstroschein I would like to continue but if I'm blocking you feel free to take over.

samuelstroschein:
@shivan-s you can complete it no rush :)

felixhaeberle:
@samuelstroschein Can you give guidance on what is missing here to accomplish this task? 
Seems a bit vague from your description.

samuelstroschein:
@felixhaeberle discussed in https://github.com/opral/inlang-paraglide-js/issues/527#issuecomment-2978875450

PR #3605, opened 2025-06-18
introduce `state_history` api

samuelstroschein:
Closes https://github.com/opral/lix-sdk/issues/319

Lovely. This shows the power of our view/change-sourced architecture. We can just add APIs as needed while keeping one source of truth.  

**Design decisions**

1. `created_at` and `updated_at` are not exposed to avoid full graph traversal 

2. the history state is materialized from the global change graph state (the graph is already materialzied in state. why build a second materialization logic from scratch? :D) 

**Desired Feedback**

1. Is the API intutive by looking at the unit tests? 

2. Should we add a `parent_change_set_ids: []` colum in this PR to let apps build up their own lineage, or wait for apps until the requirement pops up? 

For context:

Some sort of lane column is desired. But, dependent on the diagram that is displayed, what is considered a lane differs. A railway diagram where the main version stays on the left is a different graph from a DAG. 

Exposing `parent_change_set_ids` enables apps to build their own lineage logic.  

Example "lane" modelling:

```
root    V0            (lane_anchor = V0)   lane 0
         │
         V1                               lane 0
         │
         V2                               lane 0
        /  \
       /    \
   V3a(*)  V3b(*)         ← first fork     lane 1 / lane 2
                  \
                   V4                       lane 2
                  /  \
               V5a(*) V5b(*)  ← second fork lane 3 / lane 4
```


samuelstroschein:
Interesting that GH copilot pr review flagged the parent ids. I assume it took that context from my comments. @martin-lysk prob low risk to expose parent ids now. so implement now? Otherwise, apps can build a forked change graph

samuelstroschein:
Left out the parents to be delt with in a follow up issue https://github.com/opral/lix-sdk/issues/320

PR #3614, opened 2025-06-25
Human-AI section & docs improvement

felixhaeberle:


felixhaeberle:
cc @martin-lysk @samuelstroschein – I really like this section. 
I will merge this and we will improve upon this state, its WIP anyhow.

We can probably tune the headline – any suggestions?

PR #3616, opened 2025-06-25
refactor: useQuery useKeyValue instead of jotai state

NiklasBuchfink:


packages/md-app/src/state-queries.ts
```diff
@@ -0,0 +1,178 @@
...
+		return Promise.resolve(searchParams.get("t"));
+	}, 100);
+	return threadId;
+}
+
+// Core state hooks
+export function useLix() {
+	const [lix, loading, error, refetch] = useQuery(() => selectLix(), 250);
+	return { lix, loading, error, refetch };
+}
```
samuelstroschein: This seems doppeltgemoppelt. 

What's the benefit to wrap every query in a use? 


```diff
function MyComponent(){
-  const [lix] = useLix()
+  const [lix] = useQuery(selectLix)
}

```

Not wrapping the queries in use means you can get rid of the state-queries.ts file. Just have queries.ts and invoke them in `useQuery`

NiklasBuchfink: I think this was to not set the intervall every time you use a query. Different intervals are used, but we can also set all on the the default with 250ms

samuelstroschein: theoretically you can use useQuery(selectLix, 250) in a component too. i would get rid of the wrapping.  up to you

NiklasBuchfink: I know, but I have to remember what the value was in the other components that are using the same state or not? I apply the changes and you can answer

NiklasBuchfink: Done ✅

PR #3652, opened 2025-07-17
Optimize lix observe query executions

NiklasBuchfink:
Closes:
- https://github.com/opral/lix-sdk/issues/351
- https://linear.app/opral/issue/LIXDK-445/optimizelixoberserve-to-only-re-execute-a-query-if-the-state-for-the#comment-88408f5d

### DIFF
--- a/packages/lix-sdk/src/observe/determine-schema-keys.test.ts
+++ b/packages/lix-sdk/src/observe/determine-schema-keys.test.ts
@@ -307,3 +307,48 @@ test("should debug log compiled query structure", async () => {
 	
 	await lix.close();
 });
+
+test("should observe key value and handle untracked insertions", async () => {
+	const lix = await openLix({});
+	const values: any[] = [];
+
+	// Listen to a specific key value
+	const subscription = lix
+		.observe(
+			lix.db
+				.selectFrom("key_value")
+				.selectAll()
+				.where("key", "=", "untracked_test_key")
+		)
+		.subscribe({ next: (rows) => values.push(rows) });
+
+	// Wait for initial emission
+	await new Promise((resolve) => setTimeout(resolve, 10));
+	expect(values).toHaveLength(1);
+	expect(values[0]).toEqual([]);
+
+	// Insert a key value with lixcol_untracked: true
+	await lix.db
+		.insertInto("key_value")
+		.values({ 
+			key: "untracked_test_key", 
+			value: "untracked_value",
+			lixcol_untracked: true
+		})
+		.execute();
+
+	// Wait for update
+	await new Promise((resolve) => setTimeout(resolve, 10));
+
+	// Should have received an update with the untracked value
+	expect(values).toHaveLength(2);
+	expect(values[1]).toHaveLength(1);
+	expect(values[1][0]).toMatchObject({
+		key: "untracked_test_key",
+		value: "untracked_value",
+		lixcol_untracked: 1, // SQLite stores boolean as integer
+	});
+
+	subscription.unsubscribe();
+	await lix.close();
+});


packages/lix-sdk/src/observe/create-observe.test.ts
```diff
@@ -499,3 +499,367 @@ test("subscribeTakeFirst with explicit limit(1) should not double-limit", async
...
+
+	// Wait for emission
+	await new Promise((resolve) => setTimeout(resolve, 10));
+
+	// Observer should have received an update
+	expect(joinEmissions).toHaveLength(countBefore + 1);
+
+	joinSub.unsubscribe();
+	await lix.close();
+});
```
samuelstroschein: How is this testing a join? This is testing a normal where clause

packages/lix-sdk/src/observe/create-observe.test.ts
```diff
@@ -499,3 +499,367 @@ test("subscribeTakeFirst with explicit limit(1) should not double-limit", async
...
+	
+	// Insert into key_value should trigger re-execution
+	await lix.db.insertInto("key_value").values({ key: "ast_test_key", value: "value" }).execute();
+	await new Promise((resolve) => setTimeout(resolve, 10));
+	
+	expect(emissions.length).toBe(initialCount + 1);
+	
+	sub.unsubscribe();
+	await lix.close();
+});
```
samuelstroschein: That was prob the first tets that is not needed anymore given that the other tests pass

packages/lix-sdk/src/observe/create-observe.ts
```diff
@@ -28,6 +28,170 @@ function areResultsEqual<T>(a: T[], b: T[]): boolean {
 	return JSON.stringify(a) === JSON.stringify(b);
 }
 
+/**
+ * Maps table names to their corresponding schema keys.
+ * This is the reverse mapping of what's done in schema.ts.
+ */
+export function determineSchemaKeys(compiledQuery: any): string[] {
```
samuelstroschein: I would put this into a separate file `src/observe/determine-schema-keys.ts`

packages/lix-sdk/src/observe/create-observe.ts
```diff
@@ -28,6 +28,170 @@ function areResultsEqual<T>(a: T[], b: T[]): boolean {
...
+		change_set_element: "lix_change_set_element",
+		change_set_edge: "lix_change_set_edge",
+		change_set_label: "lix_change_set_label",
+		change_set_thread: "lix_change_set_thread",
+		// Add special tables that don't have direct schema mappings but are important
+		change: "change", // Special case for change table
+		state: "state", // Virtual state table - could include multiple schema keys
+		active_version: "lix_version", // Maps to version schema
+		active_account: "lix_account", // Maps to account schema
+	};
```
samuelstroschein: Import the `LixSchemaViewMap` from `src/database/schema.ts` for a non hardcoded version. if you get a circular dependency issue, move the schema view map into its own file in `src/database/schema-view-map.ts`

packages/lix-sdk/src/observe/create-observe.ts
```diff
@@ -28,6 +28,170 @@ function areResultsEqual<T>(a: T[], b: T[]): boolean {
...
+		}
+	}
+
+	return schemaKeys;
+}
+
+/**
+ * Extracts table names from Kysely AST nodes recursively.
+ */
+function extractTableNamesFromQueryNode(
```
samuelstroschein: Wil that was simpler then expected. Awesome

packages/lix-sdk/src/observe/determine-schema-keys.test.ts
```diff
@@ -0,0 +1,291 @@
...
+	const schemaKeys = determineSchemaKeys(checkpointsQuery.compile());
+	
+	// Should include schema keys for all joined tables
+	expect(schemaKeys).toContain("lix_change_set");
+	expect(schemaKeys).toContain("lix_change_set_element");
+	expect(schemaKeys).toContain("lix_change_set_label");
+	expect(schemaKeys).toContain("lix_label");
+	
+	await lix.close();
+});
```
samuelstroschein, line 103: AWESOME!

packages/lix-sdk/src/observe/determine-schema-keys.test.ts
```diff
@@ -0,0 +1,291 @@
...
+	const schemaKeys = determineSchemaKeys(activeFileQuery.compile());
+	
+	// Should include schema keys for both main query and subquery tables
+	expect(schemaKeys).toContain("lix_file");
+	expect(schemaKeys).toContain("lix_key_value");
+	
+	await lix.close();
+});
+
+test("should handle queries with aliases", async () => {
```
samuelstroschein, line 130: I would add an aliases join with a schema key like `lix_key_value`. so just add a innerJoin(key_value) and expect key value to exsits too

samuelstroschein (changes requested):
WOW THIS IS AWESOME!

packages/lix-sdk/src/observe/create-observe.test.ts
samuelstroschein: I would test a proper join. lix.db.selectFrom(key_value).innerJoin(something_else)

packages/lix-sdk/src/hooks/create-hooks.ts
```diff
@@ -25,7 +25,7 @@ export type LixHooks = {
 	 * unsubscribe();
 	 * ```
 	 */
-	onStateCommit: (handler: () => void) => () => void;
+	onStateCommit: (handler: (data: any) => void) => () => void;
```
samuelstroschein: Type the data please. I think we expect an array of changes right?

samuelstroschein: `data: { changes: Change[] }`

samuelstroschein: Actually, no its not changes. It's `data: { changes: StateAllRow }` because we might have untracked state which is not captured in changes! 

@NiklasBuchfink important to type it right

We whouls also have a test for untracked state. The subscription should work for it too ofc

packages/lix-sdk/src/observe/create-observe.test.ts
NiklasBuchfink: indeed, makes no sense. pls check the updated test

packages/lix-sdk/src/hooks/create-hooks.ts
NiklasBuchfink: But we pass in changesToRealize which is an Array of Arrays

NiklasBuchfink: @samuelstroschein What is `StateAllRow`? I can't find it in the entire repo

NiklasBuchfink: @samuelstroschein Please check https://github.com/opral/monorepo/pull/3652/commits/4f7dfa706e5e6e301c6466f02d18034c4f1cb0ad as well. How do you expect to add untracked state as well? I think there is no clear type for `Changes`, `ChangesToRealize` and untracked state.

samuelstroschein: Make a test that listens to a key value. 

Then insert a key value with `lixcol_untracked: true`

Regarding the type, you can leave it as Change for now

NiklasBuchfink: Done

samuelstroschein (approved):
Lets go

PR #3681, opened 2025-08-24
fix: make paraglide crash build processes if it can't compile

selimhex:
This is a simple proposal to fix https://github.com/opral/inlang-paraglide-js/issues/553

### What it does
`npm run build` in the  [reproduction repo](https://github.com/selimhex/paraglide-repro-non-failing-fails) currently succeeds with the unexpected and unwanted result:
```
ERROR  [paraglide-js] Failed to compile project: ENOENT: no such file or directory, open 'project.inlang/settings.json'                                        

ℹ [paraglide-js] Please check your translation files for syntax errors.                                                                                        
✓ 106 modules transformed.
dist/index.html                  0.46 kB │ gzip: 0.30 kB
dist/assets/index-yJpzg09Q.css   1.26 kB │ gzip: 0.63 kB
dist/assets/index-D5wfMMrL.js   22.08 kB │ gzip: 9.44 kB
✓ built in 196ms
```

When paraglide with this fix is used in the same repo this results with a better output of actually crashing build process:
```
ERROR  [paraglide-js] Failed to compile project: ENOENT: no such file or directory, open 'project.inlang/settings.json'                                        

ℹ [paraglide-js] Please check your translation files for syntax errors.                                                                                        
✓ 0 modules transformed.
✗ Build failed in 5ms
```

#### Only run if `isProduction`
Enforcing the `throw`only during build time and not when starting dev mode (`npm run dev`) should avoid unwanted hiccups while the user is actively working on their setup. But this could be re-considered.

#### `throw` vs `this.error`
One could apparently also use `this.error`[^1][^2] in the unplugin plugin but the `buildStart` method seems to only have the `UnpluginBuildContext` type and not `UnpluginContext` as well (which implements the `.error` and `.warn` methods).
https://github.com/unjs/unplugin/blob/main/src/types.ts#L85

Upon testing I found out that `this.error(...)` in here actually works (with vite) but one has to trick TypeScript first in such a way:
```js
if (isProduction) {
	if ("error" in this && typeof this.error === "function") {
		this.error(error);
	} else throw error;
}
```

This is just too verbose to achieve the same thing, so I went with a simple throw.

[^1]:  https://unplugin.unjs.io/guide/#supported-context
[^2]: https://rollupjs.org/plugin-development/#this-error

### DIFF
--- a/.changeset/pink-streets-push.md
+++ b/.changeset/pink-streets-push.md
@@ -1,0 +1,5 @@
+---
+"@inlang/paraglide-js": minor
+---
+
+Crash the build process (for example during `vite build`) if paraglide can't compile the project.


selimhex:

PR #3692, opened 2025-09-10
Improve async custom set locale

AlanBreck:
This PR ensures that the reload functionality in `setLocale` will wait for any async custom `setLocale` functions in the array of strategies.

### DIFF
--- a/inlang/packages/paraglide/paraglide-js/docs-api/runtime/type/-internal-.md
+++ b/inlang/packages/paraglide/paraglide-js/docs-api/runtime/type/-internal-.md
@@ -921,7 +921,7 @@ avoid a circular import between `runtime.js` and
 
 > **overwriteSetLocale**(`fn`): `void`
 
-Defined in: [runtime/set-locale.js:141](https://github.com/opral/monorepo/tree/main/inlang/packages/paraglide/paraglide-js/src/compiler/runtime/set-locale.js)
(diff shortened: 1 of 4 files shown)


samuelstroschein:
Can you add a unit test that asserts this behavior?

AlanBreck:
I've added tests.

I've also slightly modified things such that `setLocale` will _only_ be async if one of the custom strategies has an async `setLocale`, otherwise it will proceed as a synchronous call. This should make it logical in behavior, and be minimal to impact on existing users.

Thanks to @fallaciousreasoning for help with both.

samuelstroschein:
@AlanBreck thanks for the effort! 

Your reasoning on why the overload of Promise<T> | T for `setLocale()` is that "no custom async strategy defined = setLocale remains sync and exsiting code keeps functioning"?

If yes, then I agree that this can be published as patch and not as a new major version

AlanBreck:
> Your reasoning on why the overload of Promise | T for setLocale() is that "no custom async strategy defined = setLocale remains sync and exsiting code keeps functioning"?

Correct. That and `setLocale` on the server always being synchronous (unless that should be dependent on whether custom `setLocale` functions are async as well; can't say myself as I've not used it on the server).

> If yes, then I agree that this can be published as patch and not as a new major version

I'll just caveat this by saying I don't have a strong opinion on whether this is a patch or a minor. :-) I'm happy to change the changeset to whatever you recommend.

samuelstroschein:
Merged and released as `2.3.1`

PR #3721, opened 2025-10-28
fix: make locale matching case-insensitive

stonith404:
This PR fixes https://github.com/opral/inlang-paraglide-js/issues/549 which reported that locales containing uppercase characters like `pt-BR` failed to load when using `extractLocaleFromNavigator` or `extractLocaleFromHeader`.

The issue occurred because these functions converted locales to lowercase, while the comparison logic inside `assertIsLocale` and `isLocale` wasn't case-sensitive.

List of changes:
- Ensured locale comparisons in `assertIsLocale()` and `isLocale()` are fully case-insensitive.
- Made `assertIsLocale()` return the canonical-cased locale from `locales` instead of the raw input.
- Added new test coverage for case-insensitive behavior in `assertIsLocale()` and `isLocale()`.

### DIFF
--- a/.changeset/eleven-rats-tap.md
+++ b/.changeset/eleven-rats-tap.md
@@ -1,0 +1,16 @@
+---
+"@inlang/paraglide-js": patch
+---
+
+fix: make locale matching case-insensitive
+
+Closes https://github.com/opral/inlang-paraglide-js/issues/549
+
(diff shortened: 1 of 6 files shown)


stonith404:

stonith404:
recheck

samuelstroschein:
NIce, thank you!

I double checked. BCP-47 language tags are case insensitive https://en.wikipedia.org/wiki/IETF_language_tag#:~:text=Subtags%20are%20not%20case%2Dsensitive%2C. Aka, we're good!

PR #4374, opened 2026-06-16
Fix next-intl translator alias matching

### LINKED ISSUE #124: Cannot use nested structure of next-intl messages with Sherlock
Looking for documentation of [next-intl messages](https://next-intl-docs.vercel.app/docs/usage/messages) we can create nested message strucute and call this slicing with dots like this

```json
{
  "About": {
    "title": "About us"
  }
}
```

And use in components like:

```tsx
import {useTranslations} from 'next-intl';
 
export default function About() {
  const t = useTranslations('About');
  return <h1>{t('title')}</h1>;
}
```

If I have a sherlock plugin with a inlang project configured, sherlock does not recognize the structure of the messages and returns an error like

image

samuelstroschein:
## Summary
- resolve simple and chained aliases for namespaced next-intl translation functions
- avoid raw matches before aliases are declared and block stale aliases after shadow declarations
- support renamed destructured translators and direct `getTranslations({ namespace })` assignments with quoted namespace values
- add a minor changeset for `@inlang/plugin-next-intl`

Closes opral/sherlock#124

## Tests
- pnpm --filter @inlang/plugin-next-intl test
- git diff --check

---

> [!NOTE]
> **Medium Risk**
> Large parser rewrite in IDE-only matching logic; wrong bindings could mis-link message IDs in the editor, but behavior is heavily test-covered and does not affect runtime app code.
> 
> **Overview**
> **Replaces** the Parsimmon-based Sherlock message matcher in `@inlang/plugin-next-intl` with a custom lexer plus scope-aware binding analysis so IDE references stay accurate for more real-world next-intl patterns.
> 
> The matcher now **resolves** simple and chained aliases (e.g. `const translate = t`), renamed destructuring (`{ t: translate }`), and **direct** `const t = await getTranslations({ namespace: "..." })` with single- or double-quoted `namespace` values—not only destructured `{ t }` from translation hooks.
> 
> It also **tightens** when matches are emitted: no matches before an alias exists, no stale aliases after shadowing (blocks, parameters, catch/for bindings, reassignment), while **still** treating non–next-intl `t` bindings (e.g. `const t = other`) as unprefixed keys. Extensive new Vitest coverage documents these rules; a **minor** changeset records the plugin bump.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0936e1521e3c857ab4282acb7f79977226d50d7b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/fuzzy-next-intl-aliases.md
+++ b/.changeset/fuzzy-next-intl-aliases.md
@@ -1,0 +1,7 @@
+---
+"@inlang/plugin-next-intl": minor
+---
+
+Improve Sherlock message reference matching for next-intl translator aliases and quoted object namespace assignments.
+
+The matcher now resolves simple aliases like `const translate = t`, chained aliases, renamed destructured translation functions, and direct `getTranslations({ namespace })` assignments with single- or double-quoted namespace values.
--- a/packages/plugins/next-intl/src/ideExtension/messageReferenceMatchers.test.ts
+++ b/packages/plugins/next-intl/src/ideExtension/messageReferenceMatchers.test.ts
@@ -398,6 +398,103 @@ it("should stop resolving an alias after a later non-translator declaration shad
 	expect(matches).toHaveLength(0);
 });
 
+it("should stop resolving an alias when a block shadow uses property access", async () => {
+	const sourceCode = `
+		const t = useTranslations("DirectoryPage");
+		const translate = t;
+		{
+			const translate = t.rich;
+			<p>{translate("title")}</p>
+		}
+	`;
+	const settings: PluginSettings = {
+		pathPattern: "./{language}.json",
+	};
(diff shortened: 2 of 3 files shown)


avinashkamat48 (commented):
The alias shadowing logic is position-based but not scope-aware, so the new shadowing behavior can suppress valid calls after a block ends. For example, `const translate = t; { const translate = other; } translate("title")` should still resolve through the outer alias after the block, but `getNamespaceForFunctionName` will pick the later inner declaration and return no namespace. The added test only covers the call inside the shadowing block. Please add a regression test for a call after the block and either restore scope handling or avoid treating block-local shadowing declarations as global aliases.

PR #4378, opened 2026-07-02
[codex] Require BYOK for machine translate

samuelstroschein:
## Summary

- Require `INLANG_GOOGLE_TRANSLATE_API_KEY` before `inlang machine translate` runs, removing the hosted RPC fallback.
- Update CLI docs and BYOK docs so users know they must provide their own Google Cloud Translation API key.
- Remove the unused `@inlang/rpc` package, legacy website `/_rpc` proxy, and RPC smoke test script.

## Why

The CLI previously fell back to the hosted inlang RPC translation service when users did not provide their own key, which could incur API costs for inlang. This makes BYOK explicit and removes the unused hosted RPC surface from the repo.

## Validation

- `corepack pnpm --filter @inlang/cli exec tsc --noEmit`
- `corepack pnpm --filter @inlang/cli exec vitest run src/commands/machine/translate.test.ts --test-timeout=10000`
- `corepack pnpm --filter @inlang/website-v2 exec tsc --noEmit`
- `corepack pnpm --filter @inlang/plugin-m-function-matcher exec vitest run src/ideExtension/messageReferenceMatchers.test.ts --test-timeout=10000`
- `git diff --cached --check` before commit


Akashic101:
Are there any plans on removing the existing provided key in lower versions down the line?

samuelstroschein:
@Akashic101 yes! we are paying a couple of hundred bucks per month atm :D will remove this for lower versions too this month. 

I highly recommend that you instruct a coding agent with a skill to translate your project. you will get better translations

Akashic101:
Thank you for the reply, I was always a bit curious on the costs that are beeing carried from your end 😆 LLM is of course one way of providing translations, but is there a wish for using for example deepl as a provider next to google translate? I'd be willing to put in a PR for this to add this feature

samuelstroschein:
A PR is welcome and will be merged for deepL support :)

PR #4391, opened 2026-08-07
Faster importing by keeping map of bundle ids

SebbeBroman:
Importing many / large locales is time consuming, using a map seems to speed it up a bit. I ran the experiment with and without a map, here are the results. What is in the PR is only the map path left.

SCR-20260807-jdln


samuelstroschein:
implemented in another pr!

PR #4394, opened 2026-08-10
perf: batch fresh project imports

samuelstroschein:
## Stack

Builds on #4392. This is the corrected replacement for the accidentally empty stack PR; it is based on the map-only branch.

## What changed

When a project database is empty and a plugin returns unambiguous generated ids, insert bundles, messages, and variants in bounded batches inside the existing transaction. Inputs with ids, duplicate keys, missing relationships, or existing data continue through the established row-by-row path.

## Why

Paraglide compilation imported each message and variant with individual SQLite round trips. The fresh JSON-project path is the common compiler case.

## Validation

- SDK typecheck and tests: 112 passed, 19 skipped
- 10 locales × 1,000 messages load benchmark: ~1,669 ms → ~196 ms (~88% faster)
- batch sizes are capped at 500 to avoid SQLite parameter limits
- no public API changes

### DIFF
--- a/.changeset/batched-fresh-project-import.md
+++ b/.changeset/batched-fresh-project-import.md
@@ -1,0 +1,5 @@
+---
+"@inlang/sdk": patch
+---
+
+Batch fresh-project imports to reduce SQLite round trips during compilation.
--- a/.changeset/fast-bundle-import.md
+++ b/.changeset/fast-bundle-import.md
@@ -1,0 +1,5 @@
+---
+"@inlang/plugin-message-format": patch
+---
+
+Speed up importing large message-format projects by indexing bundles by id.
--- a/packages/plugins/inlang-message-format/src/import-export/importFiles.bench.ts
+++ b/packages/plugins/inlang-message-format/src/import-export/importFiles.bench.ts
@@ -1,0 +1,50 @@
+import { bench, describe } from "vitest";
+import { importFiles } from "./importFiles.js";
+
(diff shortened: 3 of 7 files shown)


samuelstroschein:
Rebased and benchmarked on `v3` (base `6662ad9`, head `8c617c6`).

Local Node/Vitest results:
- 1 locale × 200 messages: `origin/v3` 538–908 ms (median 662 ms, 5 cold runs); batched PR 114–154 ms (median 134 ms), ~4.9× faster / ~80% less wall time.
- 10 locales × 1,000 messages: batched PR 425–533 ms (5 cold runs). The row-by-row v3 baseline did not complete within 60 s in the same harness.
- Parser-only `importFiles`: 10×1,000 = 12.9 ms mean; 30×5,000 = 303.6 ms mean.
- Focused SDK import tests: 9/9 pass.

The old `common-operations` benchmark is not v3-compatible because v3 removed the `bundle` table; it fails with `unknown SQL table bundle`. The existing review concern about mixed optional-column shapes in batched rows remains to be addressed before merge.

samuelstroschein:
Fixed in `ee3c2f75a3`. Fresh-project batch inserts now group rows by optional-column shape before issuing each multi-row insert: bundles by `declarations`, messages by `selectors`, and variants by `matches`/`pattern`. Added a regression test mixing sparse and populated rows; SDK import tests 10/10, typecheck, ESLint, and the load benchmark pass.

